PR #21, opened 2025-11-19
This change introduces using prometheus exporter integration

manoj-freyr:
We have device-metrics-exporter as publicly available package to have prometheus exporter capability. Adding this to nodes to export device metrics which can be collected by central node and also have grafna integration to the same with said prometheus data source. Tested locally for deployemnt using

./utils/deploy_monitoring_stack.sh
which deploys these three utilities to nodes.
After this can run any other cvs tests.

	Device Metrics Exporter runs 24/7 exposing GPU metrics
	Prometheus scrapes metrics every 15 seconds automatically
	Tests run independently - metrics collected in background
	After test, you can query Prometheus to see what happened during test

> ./utils/deploy_monitoring_stack.sh
============================================
CVS Monitoring Stack Deployment
============================================

Working Directory: /etc/cvs
Cluster File: ./input/cluster_file/local_test_cluster.json
Monitoring Config: ./input/config_file/monitoring/monitoring_config.json
Prometheus Version: 2.55.0
Grafana Version: 10.4.1
Exporter Version: v1.4.0

post deployment success we can query either curl or UI based:

curl -s 'http://localhost:9090/api/v1/query?query=gpu_edge_temperature' | jq -r '.data.result[] | "GPU \(.metric.gpu_id) on \(.metric.hostname): \(.value[1])°C"'
{90427263-E74D-4AEC-909E-ACF8BCED4DA7}

or via browsing at http://localhost:9090/

Running gives

> ./utils/deploy_monitoring_stack.sh    input/cluster_file/sample_monitor_cluster.json     input/config_file/monitoring/monitoring_config.json
{99D35318-BFFB-4C18-AEFF-DB965A7BDA5C}

and 

 pytest tests/monitoring/install_device_metrics_exporter.py     --cluster_file=input/cluster_file/sample_monitor_cluster.json     --config_file=input/config_file/monitoring/monitoring_config.json   v

{E1E7AEAE-1102-4990-A50A-15BC461E91DC}

### DIFF
--- a/monitoring/dashboards/gpu-metrics-dashboard.json
+++ b/monitoring/dashboards/gpu-metrics-dashboard.json
@@ -4,851 +4,31 @@
   },
   "editable": true,
   "fiscalYearStartMonth": 0,
-  "graphTooltip": 1,
+  "graphTooltip": 0,
   "id": null,
   "links": [],
   "panels": [
     {
-      "type": "row",
-      "collapsed": false,
-      "gridPos": {
-        "h": 1,
-        "w": 24,
-        "x": 0,
-        "y": 0
-      },
-      "id": 100,
-      "title": "\ud83d\udd34 Critical Health Metrics"
-    },
-    {
-      "datasource": {
-        "type": "prometheus",
-        "uid": "prometheus"
-      },
-      "title": "GPU Edge Temperature",
-      "type": "stat",
-      "gridPos": {
-        "h": 4,
-        "w": 4,
-        "x": 0,
-        "y": 1
-      },
-      "id": 1,
-      "fieldConfig": {
-        "defaults": {
-          "unit": "celsius",
-          "thresholds": {
-            "mode": "absolute",
-            "steps": [
-              {
-                "color": "green",
-                "value": null
-              },
-              {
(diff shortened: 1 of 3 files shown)


input/cluster_file/local_test_cluster.json
solaiys: I think you added this file for testing with the local host. which may not be needed in actual case.

May be we can rename it as "sample_localhost_monitor_clustor.json"

input/config_file/monitoring/monitoring_config.json
```diff
@@ -0,0 +1,38 @@
...
+
+    "prometheus_host": "{prometheus-host}",
+    "prometheus_port": 9090,
+    "prometheus_version": "{prometheus-version}",
+    "prometheus_url": "http://{prometheus-host}:{prometheus-port}",
+
+    "grafana_host": "{grafana-host}",
+    "grafana_port": 3000,
+    "grafana_version": "{grafana-version}",
+    "grafana_url": "http://{grafana-host}:{grafana-port}",
```
solaiys: You already have the url here.
So we can remove from the conftest.py

PR #85, opened 2026-02-16
Wan + Flux disaggregated multinode implementation

speriaswamy-amd:
- Support disaggregated multinode parallel wan + flux bootstrap and tests
- Deploy model docker containers parallelly across multiple nodes and kick of tests
- Schema validation for input and output
- Gate regression from mentioned baseline


cvs/tests/inference/pytorch_xdit/pytorch_xdit_wan22_i2v_a14b.py
```diff
@@ -344,14 +414,25 @@ def s_phdl(cluster_dict):
...
 
-    # Default: use parallel-ssh for remote execution
-    return Pssh(
-        log,
-        node_list,
-        user=cluster_dict.get('username'),
-        password=cluster_dict.get('password'),
-        pkey=cluster_dict.get('priv_key_file'),
-    )
+    # Multi-node mode: prefer system OpenSSH for robustness; fall back to parallel-ssh only if it works.
```
cijohnson: @speriaswamy-amd is there any known issues in parallel-ssh, which leads to implementation of OpenSshMultiPssh?

speriaswamy-amd: @cijohnson, parallel-ssh (libssh2 client path) intermittently fail non-interactive auth on some of our clusters, while system OpenSSH succeeds with the same nodes/users. I added OpenSshMultiPssh so that for multi-node we try Pssh first with a quick auth smoke test, and if auth fails we automatically fall back to OpenSSH to keep runs reliable.

cijohnson (approved):

PR #86, opened 2026-02-18
Add CVS Aorta Benchmark CI workflow (with artifact upload)

araravik-psd:
Adds a new GitHub Actions workflow to run the CVS Aorta benchmark (test_aorta.py) on the MI325 scale runner.

The workflow:

1. Executes the benchmark via run_direct.sh

2. Runs in Non-SLURM/manual cluster mode

3. Uploads the generated pytest HTML report as a workflow artifact

This enables automated benchmark validation and easier debugging directly from CI.

### DIFF
--- a/.github/workflows/aorta-benchmark.yml
+++ b/.github/workflows/aorta-benchmark.yml
@@ -1,0 +1,47 @@
+name: CVS Aorta Benchmark
+
+on:
+  push:
+    branches:
+      - users/arravikum/aorta_benchmark
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


speriaswamy-amd (approved):
LGTM, note a recent PR was merged into CVS - #88 which enables robust aorta benchmarking by default, although our local pinned versions of CVS and the PR version is pretty similar, I have added some more guard rails to avoid more footguns in the PR, please feel free to use the upstream version or the pinned version.

PR #92, opened 2026-03-02
Fixed the issue while parsing the output in Transfer bench test with …

urtiwari:
Fixed the issue while parsing the output in Transfer bench test with latest ROCm.


cijohnson (changes requested):
Can we make this backward compatible

cijohnson (approved):

PR #118, opened 2026-04-09
fixed exec syntax, it needs a command not only strings

ahskabir:
- it fixes the following run-time error
    [err]   bash: line 1: Starting: command not found

## Motivation

CVS exec command syntax is wrong without this fix

## Technical Details

exec was given only string but a command must exists

## Test Plan

This error was revealed while running perftest. The error is non-fatal.

## Test Result

With this fix, bash error command will not be seen in the log

## Submission Checklist

### DIFF
--- a/cvs/tests/ibperf/ib_perf_bw_test.py
+++ b/cvs/tests/ibperf/ib_perf_bw_test.py
@@ -233,7 +233,7 @@ def test_ib_bw_perf(phdl, bw_test, config_dict):
         for qp_count in config_dict['qp_count_list']:
             # Log a message to Dmesg to create a timestamp record
             start_time = phdl.exec('date +"%a %b %e %H:%M"')
-            phdl.exec(f'Starting Test {bw_test} for {msg_size} and QP count {qp_count} | sudo tee /dev/kmsg')
+            phdl.exec(f'echo "Starting Test {bw_test} for {msg_size} and QP count {qp_count}" | sudo -n tee /dev/kmsg')
             ib_bw_dict[bw_test][msg_size][qp_count] = ibperf_lib.run_ib_perf_bw_test(
                 phdl,
                 bw_test,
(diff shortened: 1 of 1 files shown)


cvs/tests/ibperf/ib_perf_bw_test.py
```diff
@@ -233,7 +233,7 @@ def test_ib_bw_perf(phdl, bw_test, config_dict):
         for qp_count in config_dict['qp_count_list']:
             # Log a message to Dmesg to create a timestamp record
             start_time = phdl.exec('date +"%a %b %e %H:%M"')
-            phdl.exec(f'Starting Test {bw_test} for {msg_size} and QP count {qp_count} | sudo tee /dev/kmsg')
+            phdl.exec(f'echo Starting Test {bw_test} for {msg_size} and QP count {qp_count} | sudo tee /dev/kmsg')
```
speriaswamy-amd: @ahskabir , thank you for working on this, 

1. I think we need to quote the str or `printf` to prevent bash space expansion if any of those variables have a space (unlikely) but just being cautions: 

```bash
phdl.exec(
    f'echo "Starting Test {bw_test} for {msg_size} and QP count {qp_count}" | sudo tee /dev/kmsg'
)

```

```bash
phdl.exec(
    f'printf "%s\n" "Starting Test {bw_test} for {msg_size} and QP count {qp_count}" | sudo tee /dev/kmsg'
)
```
2. Can we please replace `sudo tee /dev/kmsg` with `sudo -n tee /dev/kmsg`, I think this would hang in password prompt if passwordless sudo was not setup

```bash
phdl.exec(
    f'printf "%s\n" "Starting Test {bw_test} for {msg_size} and QP count {qp_count}" | sudo -n tee /dev/kmsg'
)
```
A

speriaswamy-amd (changes requested):

cvs/tests/ibperf/ib_perf_bw_test.py
ahskabir: For 1, I will choose 'echo "Starting ..". Yes, I added "sudo -n.."
I have corrected the phdl.exec echo quoting in another place of this file as well

speriaswamy-amd (approved):

PR #125, opened 2026-04-16
Fix AORTA benchmark config resolution and documentation

speriaswamy-amd:
- Align the AORTA benchmark flow with the validated behavior by resolving AORTA config placeholders before schema/path checks, so the shipped sample config works like other CVS suites. 
- Refresh the AORTA README and reference/how-to docs to match the current runner behavior, document auto-clone and SSH key path caveats.


urtiwari (commented):
In aorta_path: /scratch/users/{user-id}/aorta, /scratch/users/ is confusing will it work if you use /home/{user-id}/aorta, all other configs have similar path

speriaswamy-amd:
@urtiwari, I have made the requested changes

urtiwari (approved):
Approved

PR #129, opened 2026-04-21
Unify logging and CLI output across CVS

cijohnson:
- Use cvs.lib.globals.log in library and test code; replace print with appropriate log.info, log.warning, or log.error where messages are diagnostic or operational.
- Correct log levels for failures, timeouts, skips, and validation messages.
- Keep print() for user-facing CLI paths: main.py plugin load, extension.ini parse warnings, monitors/base.py, generators (heatmap, cluster_json), and check_cluster_health monitor output.
- Remove duplicate log.info after log.warn/log.error in verify_lib snapshot diff.
- Fix html_lib heatmap tests to assert on logging with assertLogs instead of mocked stdout.
- make lint: run Ruff as before, then Pylint on cvs/ with only logging logging-too-many-args (E1205) and logging-too-few-args (E1206); install pylint in .ruff_venv alongside ruff.

Made-with: Cursor

## Submission Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,14 @@ TEST_VENV_DIR = .test_venv
 CVS_VENV_DIR = .cvs_venv
 RUFF_VENV_DIR = .ruff_venv
 RUFF_VERSION = 0.14.8
+# Scoped logging checks: argument count vs %-placeholders (runtime TypeError class).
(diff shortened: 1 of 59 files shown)


solaiys (approved):
LGTM

PR #136, opened 2026-04-26
Add scalable multi-process SSH with clean architecture

cijohnson:
Implement auto-sharding SSH operations for large host lists with improved module organization and clean separation of concerns.

- MultiProcessPssh: Auto-shards operations when host_list > hosts_per_shard
- PsshSharder: Manages host chunking and process pool execution
- ParallelConfig: Configures sharding behavior (hosts_per_shard, max_workers)
- Backward compatibility: Existing imports continue to work seamlessly

- cvs.lib.parallel.multiprocess_pssh: Multi-process SSH with auto-sharding
- cvs.lib.parallel.pssh_sharder: SSH operation sharding engine
- cvs.lib.parallel.pssh: Single-process SSH (unchanged)
- cvs.lib.parallel.config: Configuration for parallel operations
- cvs.lib.parallel_ssh_lib: Backward compatibility layer (DEPRECATED)

from cvs.lib.parallel.multiprocess_pssh import MultiProcessPssh from cvs.lib.parallel.config import ParallelConfig

config = ParallelConfig(hosts_per_shard=32, max_workers_per_cpu=4) ssh = MultiProcessPssh(log, host_list, user='root', config=config) result = ssh.exec('uptime')  # Auto-shards if host_list > 32

from cvs.lib.parallel.pssh import Pssh
ssh = Pssh(log, host_list, user='root')

from cvs.lib.parallel_ssh_lib import Pssh  # Maps to MultiProcessPssh

- Scalable: Handles large host lists via process sharding
- Configurable: Tunable sharding and worker parameters
- Fast: Parallel execution across host chunks
- Compatible: All existing code continues to work
- Maintainable: Clear separation between sharding and SSH logic

Made-with: Cursor

Closes-Jira: AIMVT-140

## Submission Checklist


cvs/lib/parallel/pssh_sharder.py
```diff
@@ -0,0 +1,115 @@
...
+                i = futures[fut]
+                results[i] = fut.result()
+
+        return results
+
+    def merge_results(self, shard_returns, original_host_list):
+        """Merge results from all shards maintaining original host order."""
+        cmd_output = {}
+        for host in original_host_list:
+            for r in shard_returns:
```
solaiys: PsshSharder.run_shard sets result = None for scp and reboot, but merge_results does:

if host in r['result']:

If r['result'] is None, that raises TypeError: argument of type 'NoneType' is not iterable in Python 3.

So for len(host_list) > hosts_per_shard (default 32), MultiProcessPssh.scp_file and MultiProcessPssh.reboot_connections are very likely to crash after shards finish, unless this is fixed (e.g. treat None as {} or skip merging for non-dict results). Unit tests mostly mock merge_results, so this may not show up in unittest.

cijohnson: def **scp** and def **reboot_connections** are dead code, as of now no users/callers for it, should we maintain it? can we remove it ?

How about this fix?

`def merge_results(self, shard_returns, original_host_list):
    """Merge results from all shards maintaining original host order."""

    # Step 1: Build a single lookup map
    merged = {}
    for shard in shard_returns:
        result = shard.get('result') or {}  # treat None as {}
        merged.update(result)

    # Step 2: Preserve original order
    cmd_output = {}
    for host in original_host_list:
        if host in merged:
            cmd_output[host] = merged[host]

    return cmd_output
`

solaiys: I think we ( cvs-internal) are using pssh.scp in cvs-internal tests.. just 1-2 places.. let me confirm that...

solaiys: Yes. cvs-internal uses the phdl.scp_file() for file copy... 
i think that will be routed to scp_file() in pssh.py in this PR.
So we are good here with the above fix i guess..

cvs/lib/parallel/pssh.py
```diff
@@ -0,0 +1,262 @@
...
+from pssh.clients import ParallelSSHClient
+from pssh.exceptions import Timeout, ConnectionError, SessionError
+
+from cvs.lib.env_lib import build_env_prefix
+
+
+def _chunk_hosts(host_list, chunk_size):
+    """Yield successive slices of host_list of length up to chunk_size."""
+    for i in range(0, len(host_list), chunk_size):
+        yield host_list[i : i + chunk_size]
```
solaiys: The standalone function _chunk_hosts() in cvs/lib/parallel/pssh.py (line 377-380) is defined but never used — PsshSharder.chunk_hosts() is the one actually called. Dead code.

cvs/lib/parallel_ssh_lib.py
```diff
@@ -3,324 +3,43 @@
...
+
+__all__ = [
+    'Pssh',
+    'scp',
+    'ParallelConfig',
+    'ParallelSSHClient',
+    'Timeout',
+    'ConnectionError',
+    'SessionError',
+]
```
solaiys, line 57: Nice to have:
Can we warn the callers about this DEPRECATION of this old Pssh class through some warning print.?

cvs/lib/parallel/multiprocess_pssh.py
```diff
@@ -0,0 +1,265 @@
...
+        for i in range(0, len(self.reachable_hosts), chunk_size):
+            shard_hosts = self.reachable_hosts[i : i + chunk_size]
+            k = len(shard_hosts)
+            sub_list = expanded[offset : offset + k]
+            offset += k
+            payloads.append(
+                self.sharder.create_payloads(
+                    'cmd_list', [shard_hosts], self._shard_init_kwargs(), cmd_list=sub_list, timeout=timeout
+                )[0]
+            )
```
solaiys, line 261: This may be the old bug carried over here. Judge the severity and fix it based on that.

exec_cmd_list Slicing Bug with Unreachable Hosts (Medium)

  In MultiProcessPssh.exec_cmd_list(), the sharding iterates over self.reachable_hosts but slices from expanded using a running offset. If reachable_hosts is shorter than host_list (some hosts were pruned
  earlier), the offset won't align with the original cmd_list positions — commands could be sent to the wrong hosts.

cvs/lib/parallel/multiprocess_pssh.py
```diff
@@ -0,0 +1,265 @@
...
+        """Create initialization kwargs for shard workers (host_list will be added by create_payloads)."""
+        return {
+            'log': self.log,
+            'user': self.user,
+            'password': self.password,
+            'pkey': self.pkey,
+            'host_key_check': self.host_key_check,
+            'stop_on_errors': self.stop_on_errors,
+            'env_vars': self.env_vars,
+        }
```
solaiys, line 115: See if this is a serious issue or not. in cvs-internal, A|FAIK, we are not using the file handlers for logging.
OR
We have to mention the limitation, DO NOT use the filehandlers for loggers.

**Logger Pickling in Multiprocessing (High)**

  _shard_init_kwargs() passes self.log into spawn'd child processes via ProcessPoolExecutor. Python logging.Logger objects are not reliably picklable across spawn contexts. Depending on the logger
  configuration (handlers with file descriptors, custom formatters, etc.), this will fail with PicklingError at runtime. The globals.log logger used throughout cvs-internal likely has file handlers that won't
  survive pickle.

cvs/lib/parallel/multiprocess_pssh.py
```diff
@@ -0,0 +1,265 @@
...
+        self.pkey = pkey
+        self.password = password
+        self.host_key_check = host_key_check
+        self.stop_on_errors = stop_on_errors
+        self.unreachable_hosts = []
+        self.env_vars = env_vars
+        self.env_prefix = build_env_prefix(env_vars)
+        self.client = None
+        self._use_process_sharding = True
+        self.log.debug(f"Environ vars: {self.env_prefix}")
```
solaiys, line 100: **Backward Compatibility issue with cvs-internal :**

  11 files in cvs-internal use from cvs.lib.parallel_ssh_lib import Pssh — all 10 test files plus build_tools/github_runner_config_setup.py.

  The shim maps Pssh to MultiProcessPssh (not the original Pssh):

 In new parallel_ssh_lib.py (line 2102 of diff):
  from cvs.lib.parallel.multiprocess_pssh import MultiProcessPssh as Pssh

  This means every existing Pssh(...) call in cvs-internal will silently become a MultiProcessPssh. While MultiProcessPssh extends Pssh and falls back to single-process mode for small host lists (<=32), this
  is a silent behavioral change — not a true backward-compatible alias.

  Specific Risk: Pssh(None, nodes, ...) in build_tools/github_runner_config_setup.py:17

  This call passes log=None. The new MultiProcessPssh.__init__ calls self.config = config or ParallelConfig.from_env() before calling super().__init__(), so None logger will work for small lists. But if
  len(nodes) > 32 (or if CVS_HOSTS_PER_SHARD env var is set low), it enters _init_sharded() which calls self.log.debug(...) — this will crash with AttributeError: 'NoneType' object has no attribute 'debug'.

-----------------------------------------------
Either we should fix it here by checking valid log object 
OR 
we should create a JIRA to fix this by passing a logger from cvs-internal build_tools/github_runner_config_setup.py:17

let me know how we can go about it .

solaiys:
Update the readme file for the following:
 - about the standalone scp.py if we are going to keep this file.
 - ENV used for host sharding (CVS_HOSTS_PER_SHARD, CVS_WORKERS_PER_CPU) expose it to the user .

in Deprecation Note, we may want to mention the following  - Silent Behavioral Change via Shim (High)
   The compat shim aliases Pssh → MultiProcessPssh, not Pssh → Pssh. Existing code that expected single-process Pssh will now get multi-process sharding behavior once host lists exceed 32.

solaiys:
**I think this is expected behavior:**
Log output will differ between small and large host lists for the same operation.

_In pssh_sharder.py:run_shard(), all shard executions are called with print_console=False (or payload.get('print_console', False)). But the caller MultiProcessPssh.exec() does its own _print_merged_outputs()
  to compensate. The problem: _print_merged_outputs reconstructs output from cmd_output strings, but the per-host output in the base Pssh._process_output formats differently (with exception strings, "ABORT"
  messages, etc.) than the merged re-print._

cvs/lib/parallel/pssh_sharder.py
cijohnson: fixed

cvs/lib/parallel/pssh.py
cijohnson: removed

cvs/lib/parallel_ssh_lib.py
cijohnson: Done

cvs/lib/parallel/multiprocess_pssh.py
cijohnson: fixed

cvs/lib/parallel/multiprocess_pssh.py
cijohnson: Following other modules and Using global.py log instead of passing log object to Pssh and child process

cvs/lib/parallel/multiprocess_pssh.py
cijohnson: Following other modules and Using global.py log instead of passing log object to Pssh and child process

cijohnson:
Updated README and .rst and addressed all the comments

solaiys (approved):
LGTM.

**One more thing which you may want to double check:** 

```
3. ParallelConfig.max_workers formula can produce very large values (Low)

  return max(self.hosts_per_shard, cpu_count * self.max_workers_per_cpu)
  On a 128-core machine with defaults (4 workers/CPU), this yields max(32, 512) = 512 worker processes. For a 1000-node cluster split into ~32 shards, you'd spawn 512 processes but only have 32 payloads —
  min(len(payloads), config.max_workers) caps this, so it's safe in practice. But the formula itself is confusing — max_workers should logically be capped at the number of shards, not at CPU-derived values for    large machines.
```

I felt it should be OK.
claude review also stated as LOW severity.

Just double check if anything special handling needed if its going to be 512 processes.

PR #137, opened 2026-04-27
[WIP] Implement sshkeyscan subcommand for cluster SSH host key management

cijohnson:
This adds a new CLI subcommand to scan and update SSH host keys for all nodes in a cluster, supporting both local and remote execution modes.

- **sshkeyscan subcommand**: Parallel SSH host key scanning for cluster nodes
- **Local execution**: Run ssh-keyscan from current node
- **Remote execution**: Run ssh-keyscan from cluster head node via pssh
- **Parallel processing**: Configurable concurrent ssh-keyscan processes (default: 20)
- **Key management**: Optional removal of existing host keys before scanning
- **Dry-run mode**: Preview operations without making changes
- **Flexible configuration**: Support for CLUSTER_FILE env var or --cluster_file arg

cvs sshkeyscan --cluster_file cluster.json                    # Local scan
cvs sshkeyscan --cluster_file cluster.json --at head         # Remote scan
cvs sshkeyscan --at head --remove-existing --dry-run         # Preview cleanup
CLUSTER_FILE=cluster.json cvs sshkeyscan --parallel 50       # High concurrency

Made-with: Cursor

## Submission Checklist

### DIFF
--- a/cvs/cli_plugins/sshkeyscan_plugin.py
+++ b/cvs/cli_plugins/sshkeyscan_plugin.py
@@ -1,0 +1,467 @@
+import json
+import sys
+import os
+import subprocess
+import tempfile
+import logging
+
+from .base import SubcommandPlugin
+from cvs.lib.parallel_ssh_lib import Pssh
+
+
+class SSHKeyScanPlugin(SubcommandPlugin):
+    def __init__(self):
+        super().__init__()
+        # Create logger for Pssh instances
+        self.logger = logging.getLogger(self.__class__.__name__)
+
+    def get_name(self):
+        return "sshkeyscan"
+
+    def get_parser(self, subparsers):
+        parser = subparsers.add_parser("sshkeyscan", help="Scan and update SSH host keys for all cluster nodes")
+        parser.add_argument(
+            "--cluster_file", help="Path to cluster configuration JSON file (overrides CLUSTER_FILE env var)"
+        )
+        parser.add_argument(
+            "--known_hosts", default="~/.ssh/known_hosts", help="Path to known_hosts file (default: ~/.ssh/known_hosts)"
+        )
+        parser.add_argument("--remove-existing", action="store_true", help="Remove existing host keys before scanning")
+        parser.add_argument("--dry-run", action="store_true", help="Show what would be done without making changes")
+        parser.add_argument(
+            "--parallel", type=int, default=20, help="Number of parallel processes to use (default: 20)"
+        )
+        parser.add_argument(
+            "--at",
+            choices=['local', 'head'],
+            default='local',
+            help="Where to run the scan: 'local' (current node) or 'head' (head node from cluster)",
+        )
+        parser.set_defaults(_plugin=self)
+        return parser
+
+    def get_epilog(self):
+        return """
+SSH Key Scan Commands:
+  cvs sshkeyscan --cluster_file cluster.json                    Scan locally (default)
+  cvs sshkeyscan --cluster_file cluster.json --at head         Scan from head node (MPI)
+  cvs sshkeyscan --at head --remove-existing                   Remove old keys and rescan from head
+  cvs sshkeyscan --at head --dry-run                           Show what would be done on head node
+  CLUSTER_FILE=cluster.json cvs sshkeyscan --at head           Use env var for cluster file"""
+
+    # Core command building functions
+    def build_remove_keys_command(self, hosts_file, known_hosts_file):
+        """Build command to remove existing SSH host keys"""
+        return f"cat {hosts_file} | xargs -I {{}} ssh-keygen -f {known_hosts_file} -R {{}}"
+
+    def build_scan_command(self, hosts_file, known_hosts_file, parallel, batch_size, timeout, dry_run=False):
+        """Build ssh-keyscan command with xargs and temp files for race-condition safety"""
+        # For dry run, just return a simple command
+        if dry_run:
+            return f"echo 'DRY RUN: Would scan hosts and append to {known_hosts_file}'"
+
+        # Generate unique temp directory for this scan operation
+        import uuid
+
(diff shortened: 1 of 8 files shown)


solaiys:
What Looks Good

  - Clean plugin architecture following existing patterns (SubcommandPlugin base class)
  - Good test coverage (542 lines of tests for 300 lines of code)
  - Comprehensive documentation with troubleshooting guide
  - Dry-run mode is a nice safety feature
  - Error handling with clear user-facing messages

cvs/cli_plugins/sshkeyscan_plugin.py
```diff
@@ -0,0 +1,300 @@
...
+            if not lines or not any(
+                line
+                for line in lines
+                if ('ssh-' in line or line.startswith('|1|') or 'ecdsa-sha2' in line or 'ed25519' in line)
+            ):
+                return f"{host}: FAILED - Invalid output: {output.strip()}"
+
+            # Append to known_hosts file - escape single quotes in the output
+            escaped_output = output.strip().replace("'", "'\"'\"'")
+            append_cmd = f"echo '{escaped_output}' >> {known_hosts_file}"
```
solaiys: When It runs on multiple threads, do you think any chance of race condition or corruption during the appending of the same file "known_hosts_file". ?
I hope it should not.

cvs/cli_plugins/sshkeyscan_plugin.py
```diff
@@ -0,0 +1,300 @@
...
+    ):
+        """Scan SSH host key for a single host (remote execution via thread-local pssh)."""
+        try:
+            # Create thread-local pssh handle to avoid race conditions
+            head_node = cluster_config['head_node']
+            username = cluster_config['username']
+            priv_key_file = cluster_config['priv_key_file']
+            env_vars = cluster_config.get('env_vars')
+
+            thread_pssh = Pssh(None, [head_node], user=username, pkey=priv_key_file, env_vars=env_vars)
```
solaiys: None as the logger:
The current Pssh.__init__ calls self.log.debug(...) — this will crash with AttributeError: 'NoneType'
  object has no attribute 'debug'. If PR #136 is also merged (which aliases Pssh to MultiProcessPssh), the failure mode is the same but in a different code path.

cvs/cli_plugins/sshkeyscan_plugin.py
```diff
@@ -0,0 +1,300 @@
...
+            username = cluster.get('username')
+            priv_key_file = cluster.get('priv_key_file')
+            env_vars = cluster.get('env_vars')
+
+            if not username or not priv_key_file:
+                print("Error: username and priv_key_file required in cluster file for remote execution.")
+                sys.exit(1)
+
+            # Create pssh handle to head node
+            phdl = Pssh(None, [head_node], user=username, pkey=priv_key_file, env_vars=env_vars)
```
solaiys: None as the logger:
The current Pssh.__init__ calls self.log.debug(...) — this will crash with AttributeError: 'NoneType'
  object has no attribute 'debug'. If PR #136 is also merged (which aliases Pssh to MultiProcessPssh), the failure mode is the same but in a different code path.

cvs/cli_plugins/sshkeyscan_plugin.py
```diff
@@ -0,0 +1,300 @@
...
+                    for host in hosts
+                }
+
+            # Collect results as they complete
+            for future in as_completed(future_to_host):
+                host = future_to_host[future]
+                try:
+                    result = future.result()
+                    results.append(result)
+                    if "SUCCESS" in result:
```
solaiys: Dry-Run Counts as "Failed" (Low — Bug)

  The run() method checks if "SUCCESS" in result to count successes. But dry-run returns "Would scan and add SSH host key" which doesn't contain "SUCCESS", so all dry-run results are counted as failures, and
  the command exits with code 1. The test file works around this by fabricating fake "SUCCESS" messages (in comment: "the code counts anything without SUCCESS as failed").

solaiys:
No Cleanup of pssh Connections (Low)

  scan_host_key_remote_threadsafe creates a Pssh per thread but never calls destroy_clients(). Each thread leaks an SSH connection to the head node until garbage collection.

cvs/cli_plugins/sshkeyscan_plugin.py
solaiys: I think we are good here..

**Response from Claude:**

● Good question. Let me clarify — there are two separate concurrency concerns here:

  1. Pssh object sharing across threads — thread_pssh (one Pssh per thread) solves this correctly. Pssh is not thread-safe, so this was the right design choice.
  2. Multiple threads appending to the same known_hosts file on the remote host — thread_pssh doesn't help here. Each thread runs its own echo '...' >> known_hosts via a separate SSH session, but they all
  target the same file on the same remote machine.

  However, you're right to push back on the severity. On Linux, >> opens the file with O_APPEND, which atomically seeks to the end before each write. Since SSH host keys are small (a few hundred bytes — well
  under PIPE_BUF of 4096 bytes), each echo completes in a single write() syscall. So in practice, the lines won't interleave and the file won't corrupt.

  The realistic risk is minimal for this use case. I'd downgrade it from "High" to Low — it's theoretically imperfect but practically safe for SSH key-sized data. I overstated that one.

solaiys:
Do you think this will be perf issue when we scale at 100s of nodes ?

 **One SSH Connection Per Host Scan in Remote Mode (Medium — Performance)**

  scan_host_key_remote_threadsafe creates a brand-new Pssh(None, [head_node], ...) for every single host being scanned. With 100 nodes and --parallel 50, that's 100 separate SSH connections to the head node.
  This defeats pssh's purpose (one connection, many commands) and could overwhelm the head node's SSH daemon. A better approach would be batching hosts into a single pssh command.

PR #142, opened 2026-04-28
fix(rccl): surface clearer RCCL result-file failures

speriaswamy-amd:
## Summary
- add a dedicated RCCL result-file loader that reports the unreadable path and raw head-node output
- use that helper in both regression and perf result parsing paths so missing or empty JSON files fail with a clearer message

## Test plan
- [x] `python -m compileall -q cvs/lib/rccl_lib.py`
- [x] `pytest cvs/lib/unittests/test_rccl_lib.py -q --cluster_file cvs/input/cluster_file/cluster.json --config_file cvs/input/config_file/rccl/rccl_config.json`

### DIFF
--- a/cvs/lib/rccl_lib.py
+++ b/cvs/lib/rccl_lib.py
@@ -49,6 +49,22 @@ def _is_severe_wrong_corruption_error(err: ValidationError) -> bool:
     return 'SEVERE DATA CORRUPTION' in s or "'#wrong'" in s
 
 
+def _load_rccl_json_result(shdl, head_node, result_file):
+    """Read and parse a rccl-tests JSON result file with a clearer failure mode."""
+    result_dict_out = shdl.exec(f'cat {result_file}')
+    raw_output = result_dict_out[head_node].strip()
+    try:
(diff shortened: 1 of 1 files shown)


cvs/lib/rccl_lib.py
cijohnson: which the proposed change in _read_json_from_head_node, dtype_result_out can be a empty list

so we should change this log  line to

log.info(f'{dtype}: {len(validated)} rccl-tests row(s) passed schema validation')


or 

if validated:
    log.info(f'{dtype}: schema validation passed for {len(validated)} result row(s)')
else:
    log.warning(f'{dtype}: no result rows to validate (empty JSON list or parse failure already recorded)')

cijohnson (changes requested):

cvs/lib/rccl_lib.py
speriaswamy-amd: Updated in 96f7dd44 to report the number of rccl-tests rows that passed schema validation, so an empty result no longer logs a misleading validation-success message.

cijohnson (approved):

PR #147, opened 2026-04-29
fixed namings for rccl results files

ahskabir:
## Motivation

In the rccl config json file, user provides path to a file where rccl result is supposed to be saved. CVS also allows to run collectives for various data types, for example float. This file name is appended by "_float.json" when data type is json. Besides, without this patch the CVS wrongfully tries to save the result in the controller's filesystem not on the head node.

## Technical Details

CVS uses with open syntax of python to save result file. It assumes /home/"userid" to be present, where userid is provided in cluster file. But then again it was meant for head node. CVS wronfully tries to open the file on the controller's filesystem and not on the headnode.

## Test Plan

It is tested on a 2N SuperMicro MI350 systems.

## Test Result

Without this patch the rccl_perf test case was failing. 

## Submission Checklist

### DIFF
--- a/cvs/input/config_file/rccl/rccl_config.json
+++ b/cvs/input/config_file/rccl/rccl_config.json
@@ -26,6 +26,7 @@
             "data_types": ["float"]
         },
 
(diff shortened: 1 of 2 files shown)


speriaswamy-amd (approved):
LGTM

PR #152, opened 2026-05-04
Fix RCCL post-test JSON save for payloads >30 KB by replacing libssh2 heredoc with scp

speriaswamy-amd:
The post-test combined raw and aggregated RCCL JSON files are written to the
MPI head node by `rccl_lib.rccl_perf` via a `cat > path << 'EOF' ... EOF`
heredoc over an `ssh2-python` (libssh2) exec channel.

Above roughly 30 KB, libssh2 rejects the exec request with
`InvalidRequestError`, which propagates out of `Pssh.exec` and causes pytest
to mark the entire collective FAILED even though the benchmark itself ran
clean and its results were already schema-validated. At cluster scales of
~30+ nodes (combined files routinely 30-550 KB, aggregated 90-250 KB), every
collective hits this. A 62-node overnight run on Apr 30 reproduced the issue
deterministically across all 6 collectives.

This change replaces the heredoc save with an OpenSSH `scp` subprocess from
the conductor (login) node to the head node, reusing the SSH credentials
already loaded in the `cluster_dict` fixture (`username`, `priv_key_file`):

- New private helper `_push_json_to_remote_head(...)` writes the JSON to a
  temp file on the conductor and `scp -i <key> -o BatchMode=yes -o
  StrictHostKeyChecking=no <tmp> <user>@<head>:<path>`.
- `rccl_perf(...)` gains a required `cluster_dict` positional argument and
  validates it carries `username` and `priv_key_file` (`ValueError` otherwise).
  No env-var fallback, no hardcoded defaults.
- Save failure logs `log.error(...)` and continues. Benchmark and schema
  validation have already completed by this point and raw data is in
  `test.log`; this matches the file's convention of reserving `fail_test()`
  for real benchmark/validation problems.
- `tests/rccl/rccl_perf.py::test_rccl_perf` passes the existing `cluster_dict`
  fixture to `rccl_lib.rccl_perf(...)`.

Validated:
- 4-node smoke: pytest exit 0, 2 `scp push succeeded` INFO lines, 0
  `InvalidRequestError`, JSONs land on head and parse cleanly.
- 62-node overnight: 12/12 saves succeeded across 6 collectives x 2 files,
  0 failures, payloads up to ~550 KB.

### DIFF
--- a/cvs/lib/parallel_ssh_lib.py
+++ b/cvs/lib/parallel_ssh_lib.py
@@ -299,9 +299,7 @@ class Pssh:
         Raises:
           IOError: If transfer fails on any host. Message lists offending hosts.
         """
-        self.log.info(
-            'SFTP upload %s -> %s on %s', local_file, remote_file, self.reachable_hosts
-        )
+        self.log.info('SFTP upload %s -> %s on %s', local_file, remote_file, self.reachable_hosts)
         cmds = self.client.copy_file(local_file, remote_file, recurse=recurse)
         self.client.pool.join()
         errors = []
@@ -343,12 +341,8 @@ class Pssh:
         Raises:
           IOError: If transfer fails on any host. Message lists offending hosts.
         """
-        self.log.info(
-            'SFTP download %s -> %s from %s', remote_file, local_file, self.reachable_hosts
-        )
-        cmds = self.client.copy_remote_file(
-            remote_file, local_file, recurse=recurse, suffix_separator=suffix_separator
-        )
+        self.log.info('SFTP download %s -> %s from %s', remote_file, local_file, self.reachable_hosts)
+        cmds = self.client.copy_remote_file(remote_file, local_file, recurse=recurse, suffix_separator=suffix_separator)
         self.client.pool.join()
         errors = []
         result = {}
--- a/cvs/lib/parallel_ssh_lib.py
+++ b/cvs/lib/parallel_ssh_lib.py
@@ -269,19 +269,19 @@ class Pssh:
         return cmd_output
 
     def scp_file(self, local_file, remote_file, recurse=False):
+        """
+        Backward-compatible alias for upload_file.
+
+        Kept so existing callers (and log-grep tooling looking for the legacy
+        "About to copy local file..." line) keep working. New code should call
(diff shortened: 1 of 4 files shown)


cvs/lib/parallel_ssh_lib.py
cijohnson, line 271: upload_file and download_file name looks good, but we already have scp_file. Can we make scp_file call upoload_file?

cvs/lib/parallel_ssh_lib.py
```diff
@@ -254,6 +254,90 @@ def scp_file(self, local_file, remote_file, recurse=False):
                 raise Exception("Expected IOError exception, got none")
         return
 
+    def upload_file(self, local_file, remote_file, recurse=False):
+        """
+        SFTP-upload a local file from the conductor to `remote_file` on every host
```
cijohnson: by conductor, do you mean the runner node? if yes lets call it runner instead of conductor

cijohnson (commented):
can you add unit tests for upload_file / download_file  in test_parallel_ssh_lib.py mocking (copy_file / copy_remote_file, cmd.get(), multi-host error aggregation). That would lock in the error messages and the {host: local_path} contract.

cvs/lib/parallel_ssh_lib.py
speriaswamy-amd: `scp_file` is now a one-line alias that delegates to `upload_file`. corresponding UT also locked in -`test_scp_file_delegates_to_upload_file`

cvs/lib/parallel_ssh_lib.py
speriaswamy-amd: Renamed across both files

speriaswamy-amd:
> can you add unit tests for upload_file / download_file in test_parallel_ssh_lib.py mocking (copy_file / copy_remote_file, cmd.get(), multi-host error aggregation). That would lock in the error messages and the {host: local_path} contract.

Done - Added 12 tests in TestPsshFileTransfer covering both methods + the {host: local_path} contract + the IOError aggregation message format. Run output: 34/34 pass

cijohnson (approved):

PR #153, opened 2026-05-04
Add configurable cvs_exec_timeout, rccl-tests -T timeout, and -A algoproto output to rccl_perf and rccl_regression

speriaswamy-amd:
At cluster scale (62 nodes x 12 dtypes x 24 sizes per collective), one
rccl-tests invocation can run 30+ minutes. Both `rccl_regression` and
`rccl_perf` wrap their `mpirun` call in `shdl.exec(cmd, timeout=500)` (8min
20sec), which aborts perfectly healthy benchmarks well before they finish.
There is also no way today to (a) bound rccl-tests' own internal timer or
(b) ask rccl-tests to print which NCCL algorithm/protocol/channel
combination it selected per message size — both critical for triaging
real-world performance.

This change adds three optional `rccl_test_params` keys, applied
symmetrically to both `rccl_regression` and `rccl_perf`:

| Key                          | Default          | Effect                                           |
|------------------------------|------------------|--------------------------------------------------|
| `cvs_exec_timeout`           | `2400` (40 min)  | Replaces hardcoded `timeout=500` on `shdl.exec`. |
| `rccl_timeout`               | `None` (omitted) | When set, appended as `-T <value>`.              |
| `output_algo_proto_channels` | `False`          | When truthy, appends `-A 1`. Boolean toggle.     |

`-A` is a boolean per the rccl-tests help text:

    -A,--output_algo_proto_channels <0/1>
       enable algorithm/protocol/channels output (default: 0)

`output_algo_proto_channels` is therefore a Python boolean (`bool(...)`)
that becomes `-A 1` on the command line when truthy and is omitted otherwise.
Users do not need to know about the integer wire format; they write
`"output_algo_proto_channels": true` in their config.

The model config at `cvs/input/config_file/rccl/rccl_config.json` is updated
to demonstrate the three new keys with an inline `_comment_*` description,
matching the file's existing self-documenting style.

Both functions get the same key extractions, the same `extra_flags` build
block, the same splice point in `test_cmd` (immediately before `-Z json`),
and the same `timeout=cvs_exec_timeout` substitution. This ensures users
have a single uniform schema across `tests/rccl/rccl_perf.py` and
`tests/rccl/rccl_regression.py`.

Validated:
- 4-node smoke with `cvs_exec_timeout=600`, `rccl_timeout=300`,
  `output_algo_proto_channels=true`: rccl-tests command line in test.log
  contains ` -T 300 -A 1 ` immediately before `-Z json`; pytest passes.
- 62-node overnight: `cvs_exec_timeout=21600` and `rccl_timeout=1800` both
  exercised, rccl-tests invocations completing 25-35 minutes without CVS
  abort and without rccl-tests overrunning its own timer.

### DIFF
--- a/cvs/input/config_file/rccl/rccl_config.json
+++ b/cvs/input/config_file/rccl/rccl_config.json
@@ -23,6 +23,10 @@
             "no_of_iterations": "20",
             "no_of_cycles": "1",
             "check_iteration_count": "1",
+            "_comment_timeouts_and_diag": "Optional. cvs_exec_timeout = CVS-side outer cap on the SSH-exec call (seconds, default 2400). rccl_timeout = rccl-tests internal timer passed as -T (seconds; omit to not set). output_algo_proto_channels = boolean toggle for rccl-tests' -A 1 algorithm/protocol/channels diagnostic output.",
+            "cvs_exec_timeout": "2400",
+            "rccl_timeout": "1800",
+            "output_algo_proto_channels": false,
             "data_types": ["float"]
         },
 
--- a/cvs/input/config_file/rccl/rccl_config.json
+++ b/cvs/input/config_file/rccl/rccl_config.json
@@ -23,8 +23,7 @@
             "no_of_iterations": "20",
             "no_of_cycles": "1",
             "check_iteration_count": "1",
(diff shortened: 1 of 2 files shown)


speriaswamy-amd (commented):
Move to CVS Param

cijohnson (approved):

PR #157, opened 2026-05-07
feat(megatron_training_lib): make HCA-id verification regex configurable [AIMVT-160]

atnair-amd:
## Summary

Closes [AIMVT-160]. Makes the in-container HCA-id verification regex inside `exec_nic_setup_scripts` configurable via a new optional `hca_id_pattern` key in the training config (default `bnxt_|rocep`). Replaces the previously hardcoded `bnxt_` literal so users with Mellanox/RoCE or other RDMA NICs can extend the match without patching the lib.

## Why

`exec_nic_setup_scripts` runs a Broadcom-specific libbnxt copy inside each container, then verifies the copy by `re.search(r'hca_id:\s+bnxt_', ibv_devinfo_output)`. On clusters running different nics, that regex never matches. Today the only fix is editing the lib; this PR moves the pattern into config.

## What changed

- [`cvs/lib/megatron_training_lib.py`](cvs/lib/megatron_training_lib.py)
  - Add `tdict.setdefault('hca_id_pattern', 'bnxt_|rocep')` next to the existing `nic_type` setdefault.
  - Store on `self.hca_id_pattern` and thread it into the `re.search` call:

    ```python
    if not re.search(rf'hca_id:\s+({self.hca_id_pattern})', out_dict[node], re.I):
    ```

- [`cvs/input/config_file/training/megatron/mi3xx_megatron_llama_distributed.json`](cvs/input/config_file/training/megatron/mi3xx_megatron_llama_distributed.json) — add `_example_hca_id_pattern: "bnxt_|rocep|mlx5_"` and `hca_id_pattern: "bnxt_|rocep"` after the `nccl_ib_gid_index` block. **Distributed only** — the libbnxt workaround never runs in single-node mode, so the single-node JSONs intentionally don't get the key.

- [`docs/reference/configuration-files/megatron.rst`](docs/reference/configuration-files/megatron.rst) — distributed section only:
  - Update the embedded JSON dropdown to match the JSON above.
  - Add two rows to the distributed `config` parameters table.

- [`cvs/lib/unittests/test_megatron_training_lib.py`](cvs/lib/unittests/test_megatron_training_lib.py) — append `TestExecNicSetupHcaIdPattern` with two cases:
  - `test_default_pattern_matches_rocep` — default config + `hca_id:\trocep1s0f0`, asserts `fail_test` is *not* called. Catches a regression where the default value loses `rocep`.
  - `test_override_pattern_matches_mlx5` — `hca_id_pattern='mlx5_'` + `hca_id:\tmlx5_0`, asserts `fail_test` is *not* called. Catches missing `setdefault`, missing f-string interpolation, and a stale hardcoded literal.

### DIFF
--- a/cvs/input/config_file/training/megatron/mi3xx_megatron_llama_distributed.json
+++ b/cvs/input/config_file/training/megatron/mi3xx_megatron_llama_distributed.json
@@ -22,6 +22,8 @@
         "gloo_socket_ifname": "<changeme>",
         "_example_nccl_ib_gid_index": "3",
         "nccl_ib_gid_index": "<changeme>",
+        "_example_hca_id_pattern": "bnxt_|rocep|mlx5_",
+        "hca_id_pattern": "bnxt_|rocep",
         "nccl_debug": "ERROR",
         "hf_token_file": "/home/{user-id}/.hf_token",
         "shm_size": "128G",
--- a/cvs/lib/megatron_training_lib.py
+++ b/cvs/lib/megatron_training_lib.py
@@ -162,6 +162,7 @@ class MegatronLlamaTrainingJob:
         tdict.setdefault('training_iterations', 10)
         tdict.setdefault('nnodes', 2)
         tdict.setdefault('nic_type', 'thor2')
+        tdict.setdefault('hca_id_pattern', 'bnxt_|rocep')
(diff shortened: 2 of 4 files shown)


anujmittal-amd:
The changes are good but want to address one things wrt regex:
Can we create the hca_id_pattern based on the input from the config.json file; instead of directly inserting the pattern in the regex? This way we can avoid input user error.

anujmittal-amd (approved):
LGTM

PR #158, opened 2026-05-07
refactor(megatron_training_lib): add ordered fallback chains for log parsing [AIMVT-161]

atnair-amd:
## Summary

Closes [AIMVT-161]. Refactors the inline regex parsing/progress/NaN checks in `cvs/lib/megatron_training_lib.py` into ordered fallback chains and pure module-level helpers. Each chain is seeded with `[new, old]` so newer Megatron output (`throughput per GPU (TFLOP/s/GPU): N`) is preferred but older output (`throughput per GPU: N`) still parses correctly. No config or doc changes — pattern lists are intentionally library-internal.

## Why

The parsing regex was edited in place when newer Megatron builds changed the output format (`throughput per GPU:` to `throughput per GPU (TFLOP/s/GPU):`), losing the ability to parse logs from the older builds. A fallback chain is the structural fix: future format changes append a new pattern instead of overwriting an old one, and old logs keep parsing.

The progress regex was also buried inside `poll_for_training_completion`'s `time.sleep(80)`-then-loop machinery, which made it un-unit-testable without monkeypatching sleeps + `phdl.exec` + `scan_for_training_errors` + `get_training_results_dict`. Extracting pure helpers makes both the parsing and progress/NaN checks testable with a single string fixture.

## What changed

**`cvs/lib/megatron_training_lib.py`:**

Three new module-level constants — ordered pattern chains, `[new, old]`:

```python
TRAINING_RESULT_PATTERNS = {
    'throughput_per_gpu': [
        r'throughput per GPU(?:\s*\([^)]*\))?\s*:\s+([0-9\.]+)',
        r'throughput per GPU:\s+([0-9\.]+)',
    ],
    'tokens_per_gpu': [r'tokens\/GPU\/s:\s+([0-9]+)'],
    'mem_usage': [r'mem usages:\s+([0-9\.]+)'],
    'elapsed_time_per_iteration': [r'elapsed time per iteration \(ms\):\s+([0-9\.]+)'],
}

TRAINING_PROGRESS_PATTERNS = [
    r'throughput per GPU(?:\s*\([^)]*\))?\s*:|tokens\/GPU\/s\s+[0-9]+',
    r'throughput per GPU:|tokens\/GPU\/s\s+[0-9]+',
]

TRAINING_NAN_PATTERNS = [...]   # same shape; behavior-preserving
```

Three new pure helpers — first non-empty match wins:

```python
def _parse_training_results(output: str) -> dict: ...
def _is_training_complete(output: str) -> bool: ...
def _has_nan_inf_results(output: str) -> bool: ...
```

- `get_training_results_dict` reduces to a thin wrapper: fetch the log via `phdl.exec`, then `_parse_training_results(output)`.
- `poll_for_training_completion` replaces its inline `re.search(...)` calls with `_is_training_complete(output)` and `_has_nan_inf_results(output)`.

**`cvs/lib/unittests/test_megatron_training_lib.py`** — append two test classes (4 tests total):

- `TestTrainingLogParsing.test_parse_new_format` — parser extracts every metric from new-format input.
- `TestTrainingLogParsing.test_parse_old_format_falls_back` — parser extracts every metric from old-format input. 
- `TestProgressDetection.test_handles_new_format` and `test_handles_old_format` — exercise the *separate* `TRAINING_PROGRESS_PATTERNS` chain (different code path from parsing).

### DIFF
--- a/cvs/lib/megatron_training_lib.py
+++ b/cvs/lib/megatron_training_lib.py
@@ -89,7 +89,9 @@ class MegatronLlamaTrainingJob:
           - phdl.exec(cmd: str) -> Dict[node, str] or str, depending on implementation
           - phdl.exec_cmd_list(cmd_list: List[str]) -> Dict[node, str]
       - Docker container is pre-deployed and accessible on each node.
-      - Training scripts exist under /workspace/Megatron-LM/examples/llama/.
+      - Training scripts exist under {megatron_root}/examples/llama/ (default
+        `/workspace/Megatron-LM/`; configurable via the `megatron_root` and
+        `training_scripts` keys in the training config).
       - External helpers referenced in the methods are available in scope:
           - linux_utils.get_rdma_stats_dict, linux_utils.get_nic_ethtool_stats_dict
           - json_to_dict, fail_test, verify_dmesg_for_errors, log, training_err_dict
@@ -174,6 +176,14 @@ class MegatronLlamaTrainingJob:
         tdict.setdefault('master_address', '127.0.0.1')
         tdict.setdefault('verify_network_errors', 'False')
         tdict.setdefault('rocm_dir', '')
+        tdict.setdefault('megatron_root', '/workspace/Megatron-LM')
+        tdict.setdefault(
+            'training_scripts',
+            {
+                'llama-3': 'examples/llama/train_llama3.sh',
+                'llama-2': 'examples/llama/train_llama2.sh',
+            },
+        )
 
         self.container_image = tdict['container_image']
         self.container_name = tdict['container_name']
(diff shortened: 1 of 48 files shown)


cvs/lib/megatron_training_lib.py
```diff
@@ -25,6 +25,75 @@
...
+    r'throughput per GPU(?:\s*\([^)]*\))?\s*:|tokens\/GPU\/s\s+[0-9]+',
+    r'throughput per GPU:|tokens\/GPU\/s\s+[0-9]+',
+]
+
+# NOTE: the `[NaN|Inf]` character class is a long-standing bug in the
+# original code (it matches the literal characters `N`, `a`, `I`, `n`,
+# `f`, `|`, not the strings `NaN` or `Inf`). Preserved here behavior-
+# preserving — fixing the regex semantics is a separate ticket.
+TRAINING_NAN_PATTERNS = [
+    r'throughput per GPU(?:\s*\([^)]*\))?\s*:\s+[NaN|Inf]',
```
anujmittal-amd: Can you check this out? this regex will not catch NaN|Inf patterns.

This is a character class, so it matches a single character from that set rather than the strings NaN or Inf.
You can use this: (?:NaN|Inf)

anujmittal-amd (approved):
Other than the comment below along with the hca pattern comment; LGTM for merge.

cvs/lib/megatron_training_lib.py
atnair-amd: addressed this and added some test cases to validate Nan/Inf catching

PR #160, opened 2026-05-11
docs: consolidate orchestrator-scope note into Sphinx include

atnair-amd:
Minor docs improvement so that it is easy to update ongoing orchestrator support


urtiwari (approved):
looks fine to me

PR #161, opened 2026-05-11
Add upload_file and download_file support

cijohnson:
to MultiProcessPssh with operation registry

This commit addresses API incompatibility issues between Pssh and MultiProcessPssh classes and implements a robust operation registry system to prevent future drift.

Key Changes:
- Add missing upload_file and download_file methods to MultiProcessPssh class
- Implement operation registry (SUPPORTED_OPERATIONS) as single source of truth
- Use method names directly as operation names to eliminate mapping errors
- Add missing detailed parameter support to MultiProcessPssh.exec()

API Parity Testing:
- Add comprehensive test suite to catch missing methods automatically
- Validate method signatures match between Pssh and MultiProcessPssh
- Verify operation registry completeness and consistency
- Test runtime execution of all sharder operations with proper parameters

Operations Updated:
- exec -> exec (added detailed parameter support)
- upload_file -> upload_file (new operation, matches method name)
- download_file -> download_file (new operation, matches method name)

This ensures MultiProcessPssh maintains full API compatibility with Pssh and prevents the class of issues that led to missing upload_file/download_file methods.

## Submission Checklist

### DIFF
--- a/cvs/lib/parallel/multiprocess_pssh.py
+++ b/cvs/lib/parallel/multiprocess_pssh.py
@@ -171,10 +171,10 @@ class MultiProcessPssh(Pssh):
             for line in cmd_output[host].splitlines():
                 self.log.info("%s", line)
 
-    def exec(self, cmd, timeout=None, print_console=True):
+    def exec(self, cmd, timeout=None, print_console=True, detailed=False):
         """Execute command with automatic sharding if needed."""
         if not getattr(self, '_use_process_sharding', False):
-            return super().exec(cmd, timeout=timeout, print_console=print_console)
+            return super().exec(cmd, timeout=timeout, print_console=print_console, detailed=detailed)
 
         if self.env_prefix:
             full_cmd = f"{self.env_prefix} ; {cmd}"
@@ -195,7 +195,7 @@ class MultiProcessPssh(Pssh):
         # Use sharder for sharded execution
         host_chunks = list(self.sharder.chunk_hosts(self.reachable_hosts))
         payloads = self.sharder.create_payloads(
-            'exec', host_chunks, self._shard_init_kwargs(), cmd=cmd, timeout=timeout
+            'exec', host_chunks, self._shard_init_kwargs(), cmd=cmd, timeout=timeout, detailed=detailed
         )
         shard_returns = self.sharder.execute_sharded(payloads)
 
@@ -256,7 +256,7 @@ class MultiProcessPssh(Pssh):
             offset += k
             payloads.append(
                 self.sharder.create_payloads(
-                    'cmd_list', [shard_hosts], self._shard_init_kwargs(), cmd_list=shard_commands, timeout=timeout
+                    'exec_cmd_list', [shard_hosts], self._shard_init_kwargs(), cmd_list=shard_commands, timeout=timeout
                 )[0]
             )
 
@@ -283,7 +283,7 @@ class MultiProcessPssh(Pssh):
 
         host_chunks = list(self.sharder.chunk_hosts(self.reachable_hosts))
         payloads = self.sharder.create_payloads(
-            'scp',
+            'scp_file',
             host_chunks,
             self._shard_init_kwargs(),
             local_file=local_file,
@@ -293,6 +293,54 @@ class MultiProcessPssh(Pssh):
         shard_returns = self.sharder.execute_sharded(payloads)
         return self._merge_shard_returns(shard_returns, merge_unreachable=False)
 
+    def upload_file(self, local_file, remote_file, recurse=False):
+        """Upload file with automatic sharding if needed."""
+        if not getattr(self, '_use_process_sharding', False):
+            return super().upload_file(local_file, remote_file, recurse=recurse)
+
+        self.log.info('SFTP upload %s -> %s on %s', local_file, remote_file, self.reachable_hosts)
+
+        host_chunks = list(self.sharder.chunk_hosts(self.reachable_hosts))
+        payloads = self.sharder.create_payloads(
+            'upload_file',
+            host_chunks,
+            self._shard_init_kwargs(),
+            local_file=local_file,
+            remote_file=remote_file,
+            recurse=recurse,
+        )
+        shard_returns = self.sharder.execute_sharded(payloads)
+        return self._merge_shard_returns(shard_returns, merge_unreachable=False)
+
+    def download_file(self, remote_file, local_file, recurse=False, suffix_separator='_'):
+        """Download file with automatic sharding if needed."""
+        if not getattr(self, '_use_process_sharding', False):
+            return super().download_file(remote_file, local_file, recurse=recurse, suffix_separator=suffix_separator)
+
+        self.log.info('SFTP download %s -> %s from %s', remote_file, local_file, self.reachable_hosts)
+
+        host_chunks = list(self.sharder.chunk_hosts(self.reachable_hosts))
+        payloads = self.sharder.create_payloads(
+            'download_file',
+            host_chunks,
+            self._shard_init_kwargs(),
+            remote_file=remote_file,
+            local_file=local_file,
+            recurse=recurse,
+            suffix_separator=suffix_separator,
+        )
+        shard_returns = self.sharder.execute_sharded(payloads)
+
+        # For download_file, we need to merge the returned dictionaries from each shard
+        # The result is a dict mapping {host: actual_local_path}
+        merged_result = {}
+        for shard in shard_returns:
+            result = shard.get('result') or {}
+            if isinstance(result, dict):
+                merged_result.update(result)
+
+        return merged_result
+
     def reboot_connections(self):
         """Reboot connections with automatic sharding if needed."""
         if not getattr(self, '_use_process_sharding', False):
@@ -301,7 +349,7 @@ class MultiProcessPssh(Pssh):
         self.log.info('Rebooting Connections')
 
         host_chunks = list(self.sharder.chunk_hosts(self.reachable_hosts))
-        payloads = self.sharder.create_payloads('reboot', host_chunks, self._shard_init_kwargs())
+        payloads = self.sharder.create_payloads('reboot_connections', host_chunks, self._shard_init_kwargs())
         shard_returns = self.sharder.execute_sharded(payloads)
         return self._merge_shard_returns(shard_returns, merge_unreachable=False)
 
--- a/cvs/lib/parallel/multiprocess_pssh.py
+++ b/cvs/lib/parallel/multiprocess_pssh.py
@@ -9,12 +9,13 @@ from cvs.lib.env_lib import build_env_prefix
 from cvs.lib.parallel.pssh import Pssh
 from cvs.lib.parallel.config import ParallelConfig
 from cvs.lib.parallel.pssh_sharder import PsshSharder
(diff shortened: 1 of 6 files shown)


cvs/lib/parallel/unittests/test_multiprocess_pssh.py
```diff
@@ -405,6 +405,104 @@ def test_merge_shard_returns(self):
...
+
+        with patch('cvs.lib.parallel.multiprocess_pssh.MultiProcessPssh._init_sharded'):
+            with patch('cvs.lib.parallel.multiprocess_pssh.PsshSharder') as mock_sharder_class:
+                mock_sharder = MagicMock()
+                mock_sharder_class.return_value = mock_sharder
+
+                # Set up the mock sharder behavior
+                mock_sharder.chunk_hosts.return_value = [["host1"], ["host2"]]
+                mock_sharder.create_payloads.return_value = [
+                    {"operation": "upload", "init": {}, "local_file": "test.txt", "remote_file": "/tmp/test.txt"},
```
speriaswamy-amd: This works since it's mocked, but the method we implemented is `upload_file`. I think this should be renamed

cvs/lib/parallel/unittests/test_multiprocess_pssh.py
```diff
@@ -405,6 +405,104 @@ def test_merge_shard_returns(self):
...
+        with patch('cvs.lib.parallel.multiprocess_pssh.MultiProcessPssh._init_sharded'):
+            with patch('cvs.lib.parallel.multiprocess_pssh.PsshSharder') as mock_sharder_class:
+                mock_sharder = MagicMock()
+                mock_sharder_class.return_value = mock_sharder
+
+                # Set up the mock sharder behavior
+                mock_sharder.chunk_hosts.return_value = [["host1"], ["host2"]]
+                mock_sharder.create_payloads.return_value = [
+                    {
+                        "operation": "download",
```
speriaswamy-amd: Similar to upload this should be renamed to `download_file`

cvs/lib/parallel/unittests/test_pssh_sharder.py
```diff
@@ -319,6 +319,294 @@ def test_pssh_shard_worker_unknown_operation(self, mock_pssh_class):
...
+                    elif operation_name == 'cmd_list':
+                        mock_shard.exec_cmd_list.assert_called()
+                    elif operation_name == 'scp':
+                        mock_shard.scp_file.assert_called()
+                    elif operation_name == 'upload':
+                        mock_shard.upload_file.assert_called()
+                    elif operation_name == 'download':
+                        mock_shard.download_file.assert_called()
+                    elif operation_name == 'reboot':
+                        mock_shard.reboot_connections.assert_called()
```
speriaswamy-amd: This is off, we renamed operations from `cmd_list -> exec_cmd_list`, `scp -> scp_file`, `reboot to reboot_connection` but the assertion for this test is still using the old method naming convetion which should fail, but ironically I see it passing, not sure why, I think something in the try, catch error handling ? Maybe the whole thing could be simplified with something like - `getattr(mock_shard, operation_name).assert_called()`

cvs/lib/parallel/pssh_sharder.py
```diff
@@ -90,23 +126,74 @@ def run_shard(payload):
...
 
         try:
-            # Direct operation calls - no registry needed!
+            # Use operation registry for parameter validation and documentation
+            op_config = SUPPORTED_OPERATIONS.get(operation)
+            if not op_config:
+                # Let the else clause in the operation handler catch unknown operations
+                op_config = {'required_params': [], 'optional_params': [], 'returns_result': False}
+            method_name = operation  # Operation name IS the method name!
+            shard_method = getattr(shard, method_name)
```
speriaswamy-amd: If there's a unknown operation here, we set `op_config` to empty values, but later we try to do `shared_method = getattr(shard, method_name)` this will raise a Error instantly, preventing us from handling this in the else block. We should instead throw exception here

cvs/lib/parallel/multiprocess_pssh.py
```diff
@@ -293,6 +293,54 @@ def scp_file(self, local_file, remote_file, recurse=False):
...
+
+        # For download_file, we need to merge the returned dictionaries from each shard
+        # The result is a dict mapping {host: actual_local_path}
+        merged_result = {}
+        for shard in shard_returns:
+            result = shard.get('result') or {}
+            if isinstance(result, dict):
+                merged_result.update(result)
+
+        return merged_result
```
speriaswamy-amd: `upload_file`, `scp_file` all uses `self._merge_shard_returns(shard_returns, merge_unreachable=False)` which updates reachable/unreachable host lists and then merges result shards, but `download_file` follows a different pattern ?

cvs/lib/parallel/pssh_sharder.py
```diff
@@ -90,23 +126,74 @@ def run_shard(payload):
...
+                    # Case #2: Implementation Error - Operation in registry but no handler
+                    raise RuntimeError(
+                        f'Implementation bug: Operation \'{operation}\' is supported (in registry) '
+                        f'but handler is missing in run_shard().'
+                    )
+                else:
+                    # Case #1: User Error - Operation not supported at all
+                    raise ValueError(
+                        f'Unknown operation: {operation}. Supported operations: {list(SUPPORTED_OPERATIONS.keys())}'
+                    )
```
speriaswamy-amd: Suggestion from Claude:


  The registry + regex machinery, and why I think it's heavy

  Here's what gets added to pssh_sharder.py in this PR:

   cvs/lib/parallel/pssh_sharder.py lines 16-33
```python
  SUPPORTED_OPERATIONS = {
      'exec': {
          'required_params': ['cmd'],
          'optional_params': ['timeout', 'detailed'],
          'returns_result': True,
      },
      'exec_cmd_list': {
          'required_params': ['cmd_list'],
          'optional_params': ['timeout'],
          'returns_result': True,
      },
      'scp_file': {
          'required_params': ['local_file', 'remote_file'],
          'optional_params': ['recurse'],
          'returns_result': False,
      },
      ... # 3 more entries
  }
```

  And then run_shard walks this registry:

   cvs/lib/parallel/pssh_sharder.py lines 128-146
```python
              op_config = SUPPORTED_OPERATIONS.get(operation)
              if not op_config:
                  op_config = {'required_params': [], 'optional_params': [], 'returns_result': False}
              method_name = operation  # Operation name IS the method name!
              shard_method = getattr(shard, method_name)
              # Build arguments from payload based on operation configuration
              args = {}
              for param in op_config['required_params']:
                  if param not in payload:
                      raise ValueError(f"Missing required parameter '{param}' for operation '{operation}'")
                  args[param] = payload[param]
              for param in op_config['optional_params']:
                  if param in payload:
                      args[param] = payload[param]
              # Add common parameters for operations that support them
              if method_name in ['exec', 'exec_cmd_list']:
                  args['print_console'] = False  # Always False for sharded operations
```
  And then a 40-line if/elif tree that ignores args for exec and exec_cmd_list (it rebuilds kwargs by hand from payload directly) and only really uses args for the file/reboot operations. So the registry is doing
   ~30% of its declared job.

  Plus three regex-based tests:

  • test_sharder_supports_all_multiprocess_operations — re.findall("create_payloads(...'(\w+)'", multiprocess_source) to detect operations used by MultiProcessPssh but not in the registry.
  • test_operation_registry_completeness — uses inspect.signature(Pssh.<op>) to verify the params in the registry exist on the method.
  • test_registry_handler_consistency — re.findall("operation == '(\w+)'", run_shard_source) to detect operations in the registry without a handler in run_shard.

  What this delivers:

  1. Catches "you renamed an operation in MultiProcessPssh but forgot to rename it in the registry."
  2. Catches "you added an operation to the registry but forgot to declare a handler in run_shard."
  3. Catches "you declared required_params: ['nonexistent_param'] for an op."

  What it doesn't catch:

  4. Anyone who writes create_payloads(self._operation_var, ...) instead of create_payloads('exec', ...) (regex requires literal string).
  5. Anyone who refactors run_shard from if/elif operation == to match operation: or to a dispatch dict (regex won't find handlers).

  The deeper issue: the design principle that the registry is "single source of truth" only works if the operation tag equals the method name (operation_name == method_name), which this PR establishes. Given that
   invariant, you can drop the registry entirely and the if/elif tree mostly collapses to:
```python
  method = getattr(shard, operation, None)
  if method is None:
      raise ValueError(f"Unknown operation: {operation}")
  # build kwargs from payload, filter out 'operation' and 'init'
  call_kwargs = {k: v for k, v in payload.items() if k not in ('operation', 'init')}
  # print_console handling for the two ops that support it
  if operation in ('exec', 'exec_cmd_list'):
      call_kwargs['print_console'] = False
  result = method(**call_kwargs)
```
  That's ~10 lines instead of ~80 (registry + handler if/elif + the 3 regex tests). The drift detection becomes:

  def test_all_multiprocess_operations_have_matching_pssh_methods():
      # extract operation names from MultiProcessPssh.create_payloads(...) calls
      # for each, assert hasattr(Pssh, op_name)

  One test instead of three, and it doesn't depend on parsing run_shard's source for == strings.

  I'm raising this as a strong suggestion, not a hard block. The current code works. But the registry adds a layer of indirection that's bigger than the problem it solves, and the regex tests are themselves
  fragile (the very thing they're trying to fight).

speriaswamy-amd (changes requested):

cijohnson:
@speriaswamy-amd  addressed all the review comments, introduced ABC and class composition instead of inheritance to ensure all shardable methods of Pssh are implemented in MultiprocessPssh, also dynamically handling shardable methods in run_shard. Removed unnecessary UT's to cover old behaviour

cvs/lib/parallel/multiprocess_pssh.py
```diff
@@ -44,31 +45,32 @@ def __init__(
         # Always ensure self.log is set for backward compatibility
         self.log = global_log
 
+        # Always create Pssh instance for utility methods and delegation
+        self.pssh = Pssh(
```
speriaswamy-amd: If we always create PSSH instance, wouldn't we create the same bottleneck sharding solved? i.e we open multiple connections to hosts parallelly?

cvs/lib/parallel/multiprocess_pssh.py
```diff
@@ -275,40 +277,71 @@ def exec_cmd_list(self, cmd_list, timeout=None, print_console=True):
...
 
         self.log.info('Rebooting Connections')
 
         host_chunks = list(self.sharder.chunk_hosts(self.reachable_hosts))
-        payloads = self.sharder.create_payloads('reboot', host_chunks, self._shard_init_kwargs())
+        payloads = self.sharder.create_payloads('reboot_connections', host_chunks, self._shard_init_kwargs())
         shard_returns = self.sharder.execute_sharded(payloads)
         return self._merge_shard_returns(shard_returns, merge_unreachable=False)
 
     def destroy_clients(self):
```
speriaswamy-amd, line 333: Claude Spotted this: 

`destroy_clients()` in sharded mode does `self.client = None` (a dead attribute) and never calls `self.pssh.destroy_clients()`. Net effect: leaked ParallelSSHClient until GC.

speriaswamy-amd (approved):
LGTM

PR #169, opened 2026-05-14
cvs/rccl: surface oversubscription as WARN, per-test dmesg window, per-user hostfile

speriaswamy-amd:
## Summary

Three small fixes from debugging a 28-node `rccl_perf` run on a shared cluster:

- **`cvs/lib/verify_lib.py`** — surface amdgpu runlist oversubscription as a non-fatal **WARN**. Per [AMD docs](https://instinct.docs.amd.com/projects/amdgpu-docs/en/latest/conceptual/oversubscription.html) (`Runlist is getting oversubscribed`, `Expect reduced ROCm performance`), this is a real perf-degrading state — the HW scheduler round-robins queues and inactive queues can block the GPU for ms-scale windows. The collective itself completes correctly, so we don't fail the test, but we do log a `WARN` line so reviewers know the perf numbers from this run were measured under a degraded HW state and shouldn't be trusted blindly. Implemented via a new module-level `warn_patterns_dict` scanned alongside `err_patterns_dict` inside `verify_dmesg_for_errors`. Backward-compatible — function signature and return shape unchanged, all 7+ existing callers untouched.

- **`cvs/tests/rccl/rccl_perf.py`** — flip `verify_dmesg_for_errors(..., till_end_flag=False)` so the dmesg scan is bounded by each parametrized test's own `start_time`/`end_time` window. With `till_end_flag=True` the scan ran from `start_time` to the end of the dmesg buffer, so a kernel event from one test (e.g. a `scatter_perf` segfault) was re-flagged as a failure for every subsequent parametrized test in the run.

- **`cvs/lib/rccl_lib.py`** — write the mpirun hostfile to `/tmp/rccl_hosts_file_<USER>.txt` instead of the shared `/tmp/rccl_hosts_file.txt`. A leftover hostfile owned by another user blocks the run with `Operation not permitted` / `Permission denied` on shared cluster nodes.

## Why the WARN approach (not drop-from-failures)

Earlier rev of this PR dropped the oversubscription patterns from the failure regex outright. That made the run "green" but threw away the signal — a perf run in a degraded HW state would now be marked PASS and its bandwidth deltas silently trusted. Demoting to a WARN bucket separates the two questions:
- _Did the collective complete correctly?_ → still a PASS (correct behavior).
- _Is this a valid perf measurement?_ → flagged via WARN in `std.log`.

Real driver-side errors (`Queue preemption failed`, `Failed to evict process queues`, `No more SDMA queue to allocate`, `amdgpu: process pid`) still match the FAIL regex and behave as before.

## Test plan

- [x] 28-node validation on RCCL 2.27.7 / ROCm 7.2 and RCCL 2.28.3 / ROCm 7.13: previously 7-9 false-positive failures per run, now 14/14 collectives PASS, oversubscription events emit `WARN` lines in `std.log`.
- [x] Confirmed `Test failure` and segfault still fail the correct (and only the correct) test after the per-test window change.
- [x] Confirmed two concurrent users on the same nodes no longer collide on the hostfile.

Refs: AIMVT-175

### DIFF
--- a/cvs/tests/rccl/rccl_perf.py
+++ b/cvs/tests/rccl/rccl_perf.py
@@ -350,7 +350,11 @@ def test_rccl_perf(phdl, shdl, cluster_dict, config_dict, rccl_collective):
 
     end_time = phdl.exec('date +"%a %b %e %H:%M"')
     if can_use_sudo:
-        verify_dmesg_for_errors(phdl, start_time, end_time, till_end_flag=True)
+        # Bound dmesg scan to this test's own start..end window (per-test).
+        # till_end_flag=True scans from start_time to the end of the dmesg
+        # buffer, which causes earlier-test kernel events (e.g. a scatter_perf
+        # segfault) to repeatedly fail every subsequent parametrized test.
+        verify_dmesg_for_errors(phdl, start_time, end_time, till_end_flag=False)
 
     # Get new cluster snapshot and compare ..
     if can_use_sudo and re.search(
--- a/cvs/tests/rccl/rccl_perf.py
+++ b/cvs/tests/rccl/rccl_perf.py
@@ -309,7 +309,10 @@ def test_rccl_perf(phdl, shdl, cluster_dict, config_dict, rccl_collective):
         phdl.exec(f'sudo echo "Starting Test {rccl_collective}" | sudo tee /dev/kmsg')
 
     # start_time = phdl.exec('date')
-    start_time = phdl.exec('date +"%a %b %e %H:%M"')
+    # Seconds precision matters: verify_dmesg_for_errors' node-scraper path
+    # treats analysis_range_end as an exclusive cutoff, so a minute-truncated
+    # end time silently drops the final minute of this test's dmesg window.
+    start_time = phdl.exec('date +"%a %b %e %H:%M:%S"')
 
     # Build list of nodes and their VPC IPs (used by the RCCL test)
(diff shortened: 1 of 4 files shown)


cvs/tests/rccl/rccl_perf.py
```diff
@@ -350,7 +350,11 @@ def test_rccl_perf(phdl, shdl, cluster_dict, config_dict, rccl_collective):
 
     end_time = phdl.exec('date +"%a %b %e %H:%M"')
     if can_use_sudo:
-        verify_dmesg_for_errors(phdl, start_time, end_time, till_end_flag=True)
+        # Bound dmesg scan to this test's own start..end window (per-test).
+        # till_end_flag=True scans from start_time to the end of the dmesg
+        # buffer, which causes earlier-test kernel events (e.g. a scatter_perf
+        # segfault) to repeatedly fail every subsequent parametrized test.
+        verify_dmesg_for_errors(phdl, start_time, end_time, till_end_flag=False)
```
cijohnson, line 360: Do we need to do the same for rccl_regression.py as well?

cijohnson (changes requested):

cvs/tests/rccl/rccl_perf.py
speriaswamy-amd: Good catch — yes, same bug was present in rccl_regression.py:427. Fixed in 3d66b16 (till_end_flag=True -> False), same as rccl_perf.py.

cijohnson (approved):

PR #170, opened 2026-05-14
[release/cvs-0.2.0] cvs/rccl: fix dmesg false positives and per-user RCCL hostfile

speriaswamy-amd:
## Summary

Cherry-pick of #169 onto `release/cvs-0.2.0`. Three small fixes from debugging a 28-node `rccl_perf` run on a shared cluster:

- **`cvs/lib/verify_lib.py`** — surface amdgpu runlist oversubscription as a non-fatal **WARN**. Per [AMD docs](https://instinct.docs.amd.com/projects/amdgpu-docs/en/latest/conceptual/oversubscription.html), this is a real perf-degrading state but the collective itself completes correctly. New module-level `warn_patterns_dict` scanned alongside `err_patterns_dict` inside `verify_dmesg_for_errors`. Backward-compatible signature & return shape; existing callers untouched.
- **`cvs/tests/rccl/rccl_perf.py`** — `verify_dmesg_for_errors(..., till_end_flag=False)` so each parametrized test's dmesg scan is bounded by its own start/end window (prevents one test's kernel event from cascading into N reported failures).
- **`cvs/lib/rccl_lib.py`** — per-`$USER` mpirun hostfile path (`/tmp/rccl_hosts_file_<USER>.txt`); replaces the `sudo rm -f` workaround that was on this branch.

## Conflict resolution notes

- `cvs/lib/rccl_lib.py`: this branch had a `sudo rm -f /tmp/rccl_hosts_file.txt` workaround for the same multi-user collision. The per-user path supersedes it; workaround removed.
- `cvs/tests/rccl/rccl_perf.py`: `main` gates `verify_dmesg_for_errors` on `if can_use_sudo:` (passwordless-sudo guard not yet on this branch). For this cherry-pick the call stays unconditional — only the `till_end_flag` value is changed — to avoid introducing an undefined `can_use_sudo` reference.
- `cvs/lib/verify_lib.py`: auto-merged cleanly.

## Test plan

- [x] Cherry-pick produces the same final diff as #169 on `main`.
- [x] `python -c "import ast; ast.parse(...)"` on all three files passes.
- [ ] Smoke-run `rccl_perf` from this branch on the validation cluster.

Refs: AIMVT-175

### DIFF
--- a/cvs/lib/verify_lib.py
+++ b/cvs/lib/verify_lib.py
@@ -17,7 +17,11 @@ err_patterns_dict = {
     'crash': 'crashed|Traceback|cut here|Bug:|Call Trace|RIP:|end trace|amdgpu: Fatal error|segfault|show_stack|dump_stack|fault ',
     'test_fail': 'Test failure',
     'fault': 'no-retry page fault|Illegal register access|PROTECTION_FAULT_STATUS',
-    'driver': 'Queue preemption failed for queue|Failed to evict process queues|Runlist is getting oversubscribed|No more SDMA queue to allocate|Expect reduced ROCm performance|amdgpu: process pid',
+    # Note: 'Runlist is getting oversubscribed' and 'Expect reduced ROCm performance'
+    # are amdgpu kernel info-level messages (not errors). They fire routinely on
+    # large multi-rank RCCL runs whenever HSA queue count exceeds the runlist
+    # size, even when the run itself is healthy. Excluded from failure matching.
+    'driver': 'Queue preemption failed for queue|Failed to evict process queues|No more SDMA queue to allocate|amdgpu: process pid',
     'hardware': 'hardware error|hardware fail|ras error|uncorrectable|correctable err',
     'network': 'NIC Link is Down|link is down|ib_uverb|CQE|queue catastrophic|CQ error',
 }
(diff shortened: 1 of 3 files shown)


cvs/lib/verify_lib.py
```diff
@@ -17,7 +17,11 @@
     'crash': 'crashed|Traceback|cut here|Bug:|Call Trace|RIP:|end trace|amdgpu: Fatal error|segfault|show_stack|dump_stack|fault ',
     'test_fail': 'Test failure',
     'fault': 'no-retry page fault|Illegal register access|PROTECTION_FAULT_STATUS',
-    'driver': 'Queue preemption failed for queue|Failed to evict process queues|Runlist is getting oversubscribed|No more SDMA queue to allocate|Expect reduced ROCm performance|amdgpu: process pid',
+    # Note: 'Runlist is getting oversubscribed' and 'Expect reduced ROCm performance'
+    # are amdgpu kernel info-level messages (not errors). They fire routinely on
+    # large multi-rank RCCL runs whenever HSA queue count exceeds the runlist
+    # size, even when the run itself is healthy. Excluded from failure matching.
```
cijohnson, line 23: please add links to the reference doc in the comment for future reference

cijohnson (approved):

PR #175, opened 2026-05-18
fix(install_rvs): single-command strings for container-mode source build [AIMVT-168]

atnair-amd:
## Summary

In container orchestrator mode, `install_rvs.test_install_rvs`'s source-build path passes multi-command shell strings like `cd X; cmd Y` to `orch.exec`. The docker runtime (`docker.py:196`) expands that to `sudo docker exec <name> cd X; cmd Y`, the host shell parses `;` first, so `docker exec ... cd X` runs and dies with `OCI runtime exec failed: "cd": executable file not found in $PATH` (`cd` is a shell builtin, not an executable), and `cmd Y` then runs on the host instead of inside the container.

Rewrites each affected `orch.exec` call in the source-build branch (`install_rvs.py:302-318`) as a single command:

| Old | New |
|---|---|
| `cd X; git clone Y` | `git clone <url> <dest>` |
| `cd X; cmake -B ./build` | `cmake -S <src> -B <build>` |
| `cd X; make -j$(nproc)` | `cmake --build <build> --parallel` |
| `cd X; make -j$(nproc) package` | `cmake --build <build> --target package --parallel` |
| `cd X; sudo make install; echo "RVS_INSTALL_STATUS:$?"` | `sudo cmake --install <build>` |

Also adds `-DRVS_BUILD_TESTS=OFF` to the cmake invocation: the bundled googletest in `release/rocm-rel-7.2` declares `cmake_minimum_required(VERSION 2.6.4)`, which newer cmake versions reject. RVS's own unit tests are not needed for `install_rvs`'s verification.

The `RVS_INSTALL_STATUS:0` regex check on the old install line is dropped: `cmake --install` exits non-zero on failure, and the existing `which rvs` + `gst_single.conf` existence checks below (lines 326-340) already gate install success.

### DIFF
--- a/cvs/tests/health/install/install_rvs.py
+++ b/cvs/tests/health/install/install_rvs.py
@@ -298,28 +298,29 @@ def test_install_rvs(orch, config_dict):
                     orch.exec(f'mkdir -p {git_install_path}')
 
             # Remove any existing RVS directory and clone fresh
-            out_dict = orch.exec(f'rm -rf {git_install_path}/ROCmValidationSuite')
-            out_dict = orch.exec(f'cd {git_install_path};git clone {git_url}', timeout=300)
+            src_dir = f'{git_install_path}/ROCmValidationSuite'
+            build_dir = f'{src_dir}/build'
+            out_dict = orch.exec(f'rm -rf {src_dir}')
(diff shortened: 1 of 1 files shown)


urtiwari (approved):
LGTM.

PR #178, opened 2026-05-19
exec-plumbing fixes: DockerRuntime shell-wrap + BaremetalOrchestrator detailed kwarg

atnair-amd:
## Summary

Two pre-existing defects in the orchestrator/runtime exec layer, surfaced by the in-flight orch-port branches. Each is its own commit so they can be reviewed independently.

### `286e173` runtimes/docker: fix two defects in DockerRuntime.exec

Both surface only under `ContainerOrchestrator` and were inert before the orch-port branches started exercising container mode end-to-end against non-pre-baked images.

- **No shell wrap around `cmd`.** `exec_cmd = f"sudo docker exec {container_name} {cmd}"` -- `docker exec` runs argv via `execve` with no implicit shell, so any `cmd` containing `cd`, `;`, `&&`, `||`, `|`, globs or redirects gets parsed by the host SSH shell first. Empirical proof: `install_agfhc.py:118`'s `cd {install_dir}; cp {tar} . && tar -xvf {tar}` extracted AGFHC ONTO THE HOST (`/home/atnair/agfhc-mi300...`) instead of inside the container; the install output even printed the host's ROCm 7.2.0 instead of the container's 7.13.0 as a smoking gun. Fix: wrap with `bash -c` using `shlex.quote` so shell semantics are evaluated inside the container and embedded quotes survive the SSH-host shell unwrap.
- **Phantom `exec_on_host` call on the `hosts=` branch.** `results.update(self.orchestrator.all.exec_on_host(exec_cmd, host, ...))` -- `MultiProcessPssh` (the type behind `self.orchestrator.all`) has no `exec_on_host` method. Any test passing `hosts=[...]` raises `AttributeError`. Empirical proof: `csp_qual_agfhc`'s per-host log collection via `orch.exec(cmd, hosts=[host])` failed on `test_agfhc_xgmi_lvl1` / `test_agfhc_pcie_lvl2` after the AGFHC run itself succeeded. Fix: build a fresh `Pssh` handle for the requested host subset, mirroring `BaremetalOrchestrator.exec`'s subset branch. Cannot reuse `self.orchestrator.exec()` because that re-enters `ContainerOrchestrator` and recurses through `self.runtime.exec`.

`exec_on_head` gets the same `bash -c` wrap for symmetry.

### `3aa4a03` orchestrators/baremetal: forward `detailed` kwarg in `exec` / `exec_on_head`

`BaremetalOrchestrator.exec` and `exec_on_head` omit the `detailed` kwarg that the underlying `Pssh.exec` already supports and that `ContainerOrchestrator.exec` already accepts. Callers that branch on `output['exit_code']` work under the container orchestrator but raise

```
TypeError: BaremetalOrchestrator.exec() got an unexpected keyword argument 'detailed'
```

under the baremetal orchestrator. Surfaced by the `port-install-transferbench` branch (`AIMVT-171`, commit `0aca15c`) which switched `detect_rocm_path` / `detect_hip_compiler` from the legacy `test ... && echo Y` pattern to `orch.exec('test ...', detailed=True)`. That commit was validated container-only on a06u43; the baremetal path failed at the first `detailed=True` call site (`install_transferbench.py:58`).

Fix: add `detailed=False` to both methods, forward to the Pssh handle on all three branches (`self.all`, `self.head`, and the subset-hosts fresh `Pssh`). No new behavior in Pssh -- that layer already handles `detailed` end-to-end.

## Sequencing

These fixes are independent of the orch-port branches by design. After this lands on `main`, the port branches (`port-install-transferbench`, `port-install-agfhc`, `port-csp-qual-agfhc`, `port-agfhc-cvs`, `port-transferbench-cvs`) rebase on top to pick up correct container-mode + baremetal-mode `detailed` support.

### DIFF
--- a/cvs/core/orchestrators/baremetal.py
+++ b/cvs/core/orchestrators/baremetal.py
@@ -71,7 +71,7 @@ class BaremetalOrchestrator(Orchestrator):
             stop_on_errors=self.stop_on_errors,
         )
 
-    def exec(self, cmd, hosts=None, timeout=None):
+    def exec(self, cmd, hosts=None, timeout=None, detailed=False):
         """
         Execute command across hosts via SSH (baremetal execution).
 
@@ -79,6 +79,8 @@ class BaremetalOrchestrator(Orchestrator):
             cmd: Command to execute
             hosts: Target hosts (if None, uses all hosts)
             timeout: Command timeout
+            detailed: If True, return detailed execution info including
+                exit_code (mirrors ContainerOrchestrator.exec).
 
         Returns:
             Dictionary mapping hosts to execution results
@@ -88,7 +90,7 @@ class BaremetalOrchestrator(Orchestrator):
 
         # Use appropriate handle based on target hosts
         if set(hosts) == set(self.hosts):
-            return self.all.exec(cmd, timeout=timeout)
+            return self.all.exec(cmd, timeout=timeout, detailed=detailed)
         else:
             # For arbitrary subset (including head node), create temporary handle
             pssh = Pssh(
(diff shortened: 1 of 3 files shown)


anujmittal-amd:
Minor suggestion: should this use /bin/bash -c instead of bash -c? Using the absolute path avoids relying on PATH resolution inside the container and makes the dependency a bit more explicit.

anujmittal-amd (approved):
Minor suggestion: should this use /bin/bash -c instead of bash -c? Using the absolute path avoids relying on PATH resolution inside the container and makes the dependency a bit more explicit.

PR #181, opened 2026-05-20
Port install_agfhc to the orch fixture; drop nfs_install [AIMVT-170]

atnair-amd:
Tracking: [AIMVT-170](https://amd-hub.atlassian.net/browse/AIMVT-170).

## Changes

`cvs/tests/health/install/install_agfhc.py`:

- Delete `from cvs.lib.parallel_ssh_lib import *` and the local `phdl`/`shdl`
  fixtures. The test signature becomes `test_install_agfhc(orch, config_dict)`.
- Remove the `nfs_install` head-vs-fleet conditional. Install runs
  unconditionally on every node via `orch.exec` (matches the merged
  `install_rvs.py` behavior). The `package_tar_ball` must exist at the same
  local path on every node.
- Split multi-statement shell strings (`cd X; cp ... && tar -xvf ...`) into
  separate single-binary `orch.exec` calls. The single `./install` invocation
  that genuinely needs `cwd = install_dir` (AGFHC's installer reads sibling
  files via relative paths) is wrapped in an explicit
  `sudo bash -c 'cd X && ./install'` so the cwd dependency is visible at the
  call site instead of relying on an implicit shell. Each call now invokes one
  binary, which is what `ContainerOrchestrator`'s docker-exec transport
  requires.
- Pass `--rocm-tar` to AGFHC's installer so it bypasses apt-based dependency
  resolution in favour of `dpkg-deb` extraction. Required on TheRock-rocm
  hosts (libs not tracked by dpkg); no-op on apt-rocm hosts.

`cvs/input/config_file/health/mi300_health_config.json`:

- Remove `"nfs_install": "True"` from the `agfhc` block.

`docs/reference/configuration-files/health.rst`:

- Remove the matching `"nfs_install"` entry from the inline JSON dropdown and
  the `nfs_install` row from the AGFHC parameter table.

### DIFF
--- a/cvs/input/config_file/health/mi300_health_config.json
+++ b/cvs/input/config_file/health/mi300_health_config.json
@@ -5,7 +5,6 @@
         "install_dir": "/home/{user-id}/INSTALL/agfhc/",
         "_comments_log_dir": "log_dir has to be a NON NFS local file system",
         "log_dir": "/root/agfhc_logs",
-        "nfs_install": "True",
         "hbm_test_duration": "00:01:30"
     },
     "transferbench": {
--- a/cvs/tests/health/install/install_agfhc.py
+++ b/cvs/tests/health/install/install_agfhc.py
@@ -7,6 +7,7 @@ All code contained here is Property of Advanced Micro Devices, Inc.
 
 import pytest
 
+import os
 import re
 import time
 import json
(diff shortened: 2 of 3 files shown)


urtiwari (approved):
LGTM

PR #182, opened 2026-05-20
Port install_transferbench to the orch fixture; drop nfs_install [AIMVT-171]

atnair-amd:
Tracking: [AIMVT-171](https://amd-hub.atlassian.net/browse/AIMVT-171).

## Changes

`cvs/tests/health/install/install_transferbench.py`:

- Delete `from cvs.lib.parallel_ssh_lib import *` and the local `phdl`/`shdl`
  fixtures. The test signature becomes
  `test_install_transferbench(orch, config_dict)`.
- Rename `phdl` → `orch` in the suite-local `detect_rocm_path` and
  `detect_hip_compiler` helpers (signatures, bodies, docstrings). Post-rename
  these match `install_rvs.py`'s helpers byte-for-byte.
- Remove the `nfs_install` head-vs-fleet conditional. Clone, build, and verify
  now run on every node via `orch.exec`.
- Split multi-statement shell strings into single-binary calls so each
  `orch.exec` works under `ContainerOrchestrator`'s docker-exec transport:
  - `detect_rocm_path`: glob+pipe pipeline wrapped in explicit `bash -c`.
  - `detect_hip_compiler`: probes both `hipcc` and `amdclang++` via the
    command's exit code (`detailed=True`) instead of
    `test -d X && echo Y`.
  - TransferBench clone: `git clone URL DEST` with explicit destination drops
    the `cd X; git clone URL` cwd dependency.
  - TransferBench Makefile build: Makefile genuinely needs cwd in the source
    tree, so the call is wrapped explicitly as
    `sudo bash -c 'cd X && ROCM_PATH=v make'`.

`cvs/input/config_file/health/mi300_health_config.json`:

- Remove `"nfs_install": "True"` from the `transferbench` block.

`docs/reference/configuration-files/health.rst`:

- Remove the matching `"nfs_install"` entry from the inline JSON dropdown and
  the `nfs_install` row from the TransferBench parameter table.

### DIFF
--- a/cvs/input/config_file/health/mi300_health_config.json
+++ b/cvs/input/config_file/health/mi300_health_config.json
@@ -13,7 +13,6 @@
         "example_tests_path": "/home/{user-id}/INSTALL/TransferBench/examples",
         "git_install_path": "/home/{user-id}/INSTALL/",
         "git_url": "https://github.com/ROCm/TransferBench.git",
-        "nfs_install": "True",
         "_comment_rocm_path": "ROCm installation path. Set the placeholder changeme to auto-detect from /opt/rocm or /opt/rocm/core-*",
         "rocm_path": "<changeme>",
         "results": {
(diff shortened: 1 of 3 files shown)


urtiwari (approved):
LGTM

PR #195, opened 2026-05-31
container backend: container.lifetime schema, setup_script provisioning, and persistent mode

atnair-amd:
## Summary

Reworks the container execution backend around a single container lifecycle policy and adds in-container provisioning + a persistent (install-then-run) mode.

- **Schema**: replaces the two-axis `container.enabled` + `container.launch` with one tri-valued `container.lifetime` (`external` / `per_run` / `persistent`). Both legacy keys now hard-error with a migration message (no silent mapping), so a stale flag can't override an explicit `lifetime`.
- **Provisioning**: adds `container.setup_script`, run inside each freshly-launched container before sshd setup (base64-delivered over the existing `docker exec` channel, with `set -o pipefail` and a size guard). The packaged default installs `openssh-server`, so the base image no longer needs sshd baked in.
- **Persistent mode**: `setup_containers` attaches when the container runs on every host, cold-starts when it runs on none, and **hard-fails on a partial set** (relaunching would force-remove and wipe the still-running hosts' overlay). Includes a per-host image-SHA consistency check (cross-host skew or unreadable SHA is a hard error) and an idempotent `setup_sshd` (skips when sshd is already on 2224). Teardown is a no-op, enabling `cvs run install_rvs` then `cvs run rvs_cvs` in separate invocations.
- **Launch pull**: the container-start `docker run` timeout is raised 60s -> 900s (`CONTAINER_START_TIMEOUT_S`) so the orchestrator can pull a multi-GB image on a cold node instead of timing out.
- Docs (`cluster-file`, `run-with-containers`, cluster_file README) and `cluster_container.json` sample updated; `enabled`/`launch` references removed.

## Test plan

- [x] Unit: `test_container.py`, `test_factory.py`, `test_docker.py` (lifetime resolution, per-lifetime setup/teardown, persistent attach/partial/skew/unreadable-SHA, provisioning dispatch + size guard + pipefail payload, sshd pgrep self-match) — 72 pass.
- [x] Lint: `ruff check` clean on changed files.
- [x] End-to-end on an MI300X node (container orchestrator, `lifetime: persistent`):
  - `install_rvs` cold-starts a container, provisions sshd into a no-sshd TheRock image, installs RVS into the overlay; container survives teardown.
  - `rvs_cvs` attaches to the same container (verified identical container ID + StartedAt, no relaunch), skips sshd setup, and runs `rvs` from the persisted overlay.
  - Verified the launch path pulls a fresh 61GB image on a cold node (~9 min) within the new 900s timeout (the old 60s timeout aborted the pull).


cvs/core/orchestrators/factory.py
```diff
@@ -72,7 +179,13 @@ def __init__(self, **kwargs):
         self.priv_key_file = kwargs['priv_key_file']
         self.password = kwargs.get('password')
         self.head_node_dict = kwargs.get('head_node_dict', {})
-        self.container = kwargs.get('container', {})
+        # Normalize the container block. This is the single chokepoint:
```
cijohnson: Do we need this comment?

cvs/core/runtimes/unittests/test_docker.py
```diff
@@ -43,11 +43,13 @@ def _fake_exec(cmd, timeout=None, detailed=False, print_console=True):
...
-def _container_config(launch=True, image="img:test", extra_runtime_args=None, **extra):
-    """Minimal container config dict for setup_containers."""
+def _container_config(image="img:test", extra_runtime_args=None, **extra):
+    """Minimal container config dict for setup_containers.
+
+    The runtime no longer reads enabled/launch -- lifetime resolution lives in
+    OrchestratorConfig -- so neither key is present here.
+    """
     cfg = {
-        "enabled": True,
```
cijohnson, line 49: i dont see lifetime arg

cvs/core/runtimes/docker.py
```diff
@@ -174,6 +178,41 @@ def is_running(self, container_name):
             }
         return out
 
+    def image_sha_status(self, container_name, image_name):
```
cijohnson: i see i used only in unittests, do we need this?

cvs/core/orchestrators/factory.py
atnair-amd: removed

cvs/core/runtimes/unittests/test_docker.py
atnair-amd: added lifetime arg

cvs/core/runtimes/docker.py
atnair-amd: removed

cijohnson (approved):

PR #221, opened 2026-06-11
AIMVT-139: node-scraper offline dmesg adapter + full_dmesg_scan toggle

speriaswamy-amd:
## Summary
- Add `cvs/lib/node_scraper_adapter.py` to parse collected dmesg offline via AMD node-scraper's `DmesgAnalyzer` (`parse_dmesg` normalizes events; `event_match_lines` flattens to legacy `{node: [lines]}` strings).
- Wire `full_dmesg_scan()` behind `CVS_DMESG_PARSER` (default `node-scraper`; `legacy`/`false` selects `err_patterns_dict` regex). Auto-falls back to legacy when `amd-node-scraper` is not installed.
- Add `amd-node-scraper>=1.1.4` to `requirements.txt` and unit tests for adapter + toggle behavior.

## Test plan
- [x] `python -m unittest cvs.lib.unittests.test_node_scraper_adapter` (requires `amd-node-scraper` installed)
- [x] `python -m unittest cvs.lib.unittests.test_verify_lib.TestFullDmesgScan`
- [x] Run a CVS job with `CVS_DMESG_PARSER=node-scraper` (default) and confirm `full_dmesg_scan` collects `dmesg --time-format iso -x`
- [x] Run with `CVS_DMESG_PARSER=legacy` and confirm legacy `dmesg -T` + regex path still works

### DIFF
--- a/cvs/lib/node_scraper_adapter.py
+++ b/cvs/lib/node_scraper_adapter.py
@@ -24,19 +24,11 @@ from __future__ import annotations
 import logging
 from typing import Any, Dict, List, Optional, Union
 
+from nodescraper.models import SystemInfo
+from nodescraper.plugins.inband.dmesg.analyzer_args import DmesgAnalyzerArgs
+from nodescraper.plugins.inband.dmesg.dmesg_plugin import DmesgPlugin
+
 log = logging.getLogger(__name__)
-
-try:
-    from nodescraper.models import SystemInfo
-    from nodescraper.plugins.inband.dmesg.analyzer_args import DmesgAnalyzerArgs
-    from nodescraper.plugins.inband.dmesg.dmesg_plugin import DmesgPlugin
-
-    NODE_SCRAPER_AVAILABLE = True
-    _IMPORT_ERROR: Optional[Exception] = None
-except Exception as exc:  # pragma: no cover - only hit when dependency is absent
-    NODE_SCRAPER_AVAILABLE = False
-    _IMPORT_ERROR = exc
-
 
 DEFAULT_NODE_NAME = "cvs-node"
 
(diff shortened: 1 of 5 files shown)


cvs/lib/node_scraper_adapter.py
```diff
@@ -0,0 +1,136 @@
...
+        source.
+    """
+    _require_node_scraper()
+
+    plugin = DmesgPlugin(system_info=SystemInfo(name=node_name or DEFAULT_NODE_NAME))
+    result = plugin.analyze(
+        data=DmesgData(dmesg_content=dmesg_content or ""),
+        analysis_args=analysis_args,
+    )
+    return [_normalize_event(event) for event in result.events]
```
alexandraBara: FYI you can also call plugin.run(collection=False, data=dmesg.log, analysis_args=analysis_args) this way the plugin returns the common DataPluginResult which you can later parse similarly for NicPlugin and other plugins you might be interested in.

alexandraBara (approved):
Integration looks good

cvs/lib/verify_lib.py
```diff
@@ -29,6 +31,26 @@
 threshold_counter_val = 1000
 
 
+# Environment toggle selecting the dmesg parser backend:
+#   CVS_DMESG_PARSER=node-scraper (default) -> AMD node-scraper analyzer
+#   CVS_DMESG_PARSER=legacy                 -> historical err_patterns_dict regex
+# Falls back to legacy automatically when node-scraper is not installed.
```
cijohnson: Do we need fall back mechanism? if we add node-scrapper to requiremets, all cvs installations will have it.

requirements.txt
```diff
@@ -6,10 +6,13 @@ pytest-html
 pytest-repeat
 pytest-dependency
 xlsxwriter
-pydantic >= 2.0
+pydantic >= 2.8.2
 pandas
 tabulate
 
+# AMD node-scraper: offline dmesg/log parsing (see cvs/lib/node_scraper_adapter.py)
+amd-node-scraper >= 1.1.4
```
cijohnson: 1.1.7 is the latest version available

cvs/lib/node_scraper_adapter.py
```diff
@@ -0,0 +1,136 @@
...
+    "category",
+    "description",
+    "match_content",
+    "count",
+    "timestamps",
+    "source",
+)
+
+
+def is_available() -> bool:
```
cijohnson: we can remove this, because node-scrapper will be a mandatory  requirement for cvs

cvs/lib/node_scraper_adapter.py
```diff
@@ -0,0 +1,136 @@
...
+    """Return True if the amd-node-scraper package is importable."""
+    return NODE_SCRAPER_AVAILABLE
+
+
+def _require_node_scraper() -> None:
```
cijohnson: we can remove this, because node-scrapper will be a mandatory requirement for cvs

cijohnson (changes requested):

requirements.txt
alexandraBara: is there a reason to hardcode the version? we are at 1.1.8 as of yesterday

speriaswamy-amd: removed pinned versions, we should be just picking up latest

cijohnson (approved):

PR #223, opened 2026-06-11
vLLM single node refactor

atnair-amd:
## Summary

Remediates the lifecycle-review findings on the `vllm_single` PoC suite. Two
classes of fix: **silent-pass eliminations** (a green PASS that asserted
nothing) and **lifecycle/observability hardening** (readiness detection, HTML
timing rows, shell-injection safety). Also removes the now-unused legacy
`VllmJob`.

## Changes

**`config_loader.py` — threshold coverage is now load-time enforced**
- New `cell_key()` / `expected_cells()` give the loader and the test a single
  shared key builder, so the coverage check and the verdict lookup can't drift
  on whitespace/ordering/field names.
- New `_check_thresholds_cover_sweep` validator fails at config load if any
  sweep cell lacks a `threshold.json` entry (or vice-versa). Previously a
  mistyped key made the test silently skip its verdict and report PASS with
  zero assertions.
- Adds `client_poll_count` param (completion-poll budget; large-OSL cells need
  a bigger budget, and raising it never slows fast cells).

**`vllm_orch.py` — readiness, failure detection, cache, quoting**
- `is_ready()` now runs `grep -qiE` *inside the container* and ships back only
  an exit code, instead of `tail -30` + cat-ing the log over the wire. The
  one-shot startup banner scrolls out of a fixed tail once vLLM gets chatty, so
  the old check could miss a server that did come up.
- Client failure is now a **nonzero** failed-request count, not a bare "Failed"
  substring — the bench summary always prints `Failed requests: N`, so the old
  check false-failed every successful run. A pre-summary traceback is still a
  hard failure.
- Narrowed `EARLY_FAILURE_RE`: dropped bare `error:`/`exception:`/`traceback`,
  which match benign ROCm/vLLM startup lines and aborted servers that would
  have come up.
- HF cache pinned to the mounted models dir so weights survive teardown and
  aren't re-downloaded each run.
- `shlex.quote` on all shell interpolation (replaces the hand-rolled
  `_shquote`).
- Readiness budget raised 30→60min to cover a remote (~152GB) model pull on the
  first cell; free on the happy path (the poll exits on ready).

**`conftest.py` / `vllm_single.py` — lifecycle-as-tests + report fixes**
- Container launch / sshd / model-fetch / teardown are now individual timed,
  pass/fail tests rather than fixture body code, each asserting independent
  observation (container running, sshd listening on 2224, bytes present,
  container gone).
- HTML timing table is per-test (keyed by nodeid) with an explicit
  `stage | value | unit` header — fixes timings leaking across rows and the
  size value rendering under a "seconds" header.
- Per-cell verdict miss is now a hard `AssertionError` instead of a logged
  skip.
- `orch` fixture keeps a leak-guard teardown finalizer; explicit teardown sets
  a flag so it no-ops (no double teardown). Deterministic collection order
  pins the lifecycle sequence.

**Removed**
- `cvs/lib/inference/vllm.py` (legacy `VllmJob`) — no remaining importers.

## Validation

A full 14/14 `vllm_single` run completed on MI300X (Qwen3-Next-80B, TP=8). The
remediated code was confirmed loaded (editable install): `is_ready` grep path,
per-row 3-column timing tables, and the threshold coverage check all exercised.

## Test plan

- [ ] `cvs run vllm_single` against an MI300X node with matching
      `threshold.json` → 14/14 pass, HTML shows per-stage 3-column timing tables
- [ ] Config with a deliberately missing threshold cell → fails at load, not
      mid-run
- [ ] Confirm no remaining importers of the deleted `vllm.py`


cvs/lib/dtni/config_loader.py
```diff
@@ -0,0 +1,301 @@
...
+Loads a per-variant `config.json` + sibling `*_threshold.json`, validates
+shape with pydantic v2 (extra="forbid"), and runs a 3-pass placeholder
+substitution:
+  1. cluster placeholders (`{user-id}`) anywhere
+  2. self-reference within `paths` (e.g. `{shared_fs}`)
+  3. cross-block (`{paths.models_dir}`, etc.) into the rest of the doc
+
+A loaded variant is returned as a `VariantConfig` instance whose `container`
+field matches the dict shape that `cvs.core.orchestrators.factory.OrchestratorConfig`
+already understands (`lifetime`, `name`, `image`, `runtime{name,args}`),
```
amd-droy: ContainerSpec seems to have a runtime as RuntimeSpec, not runtime{name,args} as a flat dict

cvs/lib/dtni/config_loader.py
```diff
@@ -0,0 +1,301 @@
...
+    num_prompts: str = "3200"
+    # Completion-poll budget for the bench client = client_poll_count * 60s
+    # (plus a 120s initial wait). Large-output cells (high osl) need a bigger
+    # budget; the poll loop exits as soon as the client finishes, so raising
+    # this never slows down fast cells. See regressions REG-20260609-001.
+    client_poll_count: str = "20"
+
+
+class BenchmarkParams(_Allow):
+    pass
```
amd-droy: will this class be implemented later? what is this intended to do?

cvs/lib/dtni/config_loader.py
```diff
@@ -0,0 +1,301 @@
...
+    # numbers are curves, not tabulated values). Default true keeps the gate
+    # strict for calibrated configs -- no regression to the remediation work.
+    enforce_thresholds: bool = True
+    paths: Paths
+    model: ModelSpec
+    image: ImageSpec
+    container: ContainerSpec
+    roles: Roles
+    params: Params
+    benchmark_params: Dict[str, Any] = Field(default_factory=dict)
```
amd-droy: is this what the BechmarkParams is intended to do?
benchmark_params: BenchmarkParams = Field(default_factory=BenchmarkParams)

cvs/lib/dtni/config_loader.py
```diff
@@ -0,0 +1,301 @@
...
+            problems.append(f"sweep cells with no threshold entry: {missing}")
+        if extra:
+            problems.append(f"threshold keys matching no sweep cell (typo?): {extra}")
+        if problems:
+            msg = "threshold.json does not match the sweep matrix; " + "; ".join(problems)
+            if self.enforce_thresholds:
+                raise ValueError(msg)
+            warnings.warn(f"{msg} (enforce_thresholds=false -> record-only)", stacklevel=2)
+        return self
+
```
amd-droy, line 211: Python decorators (@model_validator(mode="after") ) runs in the definite order, if _check_remote_not_implemented fails, _check_thresholds_cover_sweep will be skipped. But if that's what is expected, then its fine.

cvs/lib/dtni/config_loader.py
```diff
@@ -0,0 +1,301 @@
...
+            out[key] = str(v)
+    return out
+
+
+def _resolve_cluster_mapping(cluster_dict):
+    user = cluster_dict.get("username") or getpass.getuser()
+    if user == "{user-id}":
+        user = getpass.getuser()
+    return {"user-id": user}
+
```
amd-droy, line 250: can be collapsed -
raw = cluster_dict.get("username") or "{user-id}"
user = getpass.getuser() if raw == "{user-id}" else raw
return {"user-id": user}

cvs/lib/dtni/config_loader.py
```diff
@@ -0,0 +1,301 @@
...
+    if not config_path.is_file():
+        raise AssertionError(f"variant config not found: {config_path}")
+
+    raw = json.loads(config_path.read_text())
+
+    threshold_candidates = sorted(config_path.parent.glob("*threshold.json"))
+    if not threshold_candidates:
+        raise AssertionError(f"no *threshold.json next to config: {config_path.parent}")
+    if len(threshold_candidates) > 1:
+        raise AssertionError(f"multiple *threshold.json files next to config (ambiguous): {threshold_candidates}")
```
amd-droy: or
raise FileNotFoundError(f"variant config not found: {config_path}")
raise FileNotFoundError(f"no *threshold.json next to config: {config_path.parent}")
raise ValueError(f"multiple *threshold.json files — ambiguous: {threshold_candidates}")

cvs/lib/inference/vllm_orch.py
```diff
@@ -0,0 +1,317 @@
...
+            f"--tokenizer-mode {self.tokenizer_mode} "
+            f"--seed {self.seed} "
+            f"--random-range-ratio {self.random_range_ratio} "
+            f"--random-prefix-len {self.random_prefix_len} "
+            f"--percentile-metrics {self.percentile_metrics} "
+            "--ignore-eos --save-result "
+            f"--result-dir {self.out_dir} --result-filename results "
+            f"> {self.client_log} 2>&1 &"
+        )
+        self.orch.exec("bash -c " + shlex.quote(client_cmd))
```
amd-droy, line 306: this way might be better, if anything ever contains a space or a $ the inner bash invocation might break silently.

p = self.variant.params
args = [
    "python3", f"bench_serving/{p.bench_serv_script}",
    "--model",              self.variant.model.id,
    "--backend",            p.backend,
    "--base-url",           f"{p.base_url}:{p.port_no}",
    "--dataset-name",       p.dataset_name,
    "--num-prompts",        self.num_prompts,
    "--random-input-len",   self.isl,
    "--random-output-len",  self.osl,
    "--max-concurrency",    self.concurrency,
    "--request-rate",       p.request_rate,
    "--burstiness",         p.burstiness,
    "--tokenizer-mode",     p.tokenizer_mode,
    "--seed",               p.seed,
    "--random-range-ratio", p.random_range_ratio,
    "--random-prefix-len",  p.random_prefix_len,
    "--percentile-metrics", p.percentile_metrics,
    "--metric-percentiles", p.metric_percentiles,
    "--ignore-eos",
    "--save-result",
    "--result-dir",         self.out_dir,
    "--result-filename",    "results",
]
bench_cmd = " ".join(shlex.quote(a) for a in args)
client_cmd = (
    f"source /tmp/server_env_script.sh && cd /app && "
    f"{bench_cmd} > {shlex.quote(self.client_log)} 2>&1 &"
)
self.orch.exec("bash -c " + shlex.quote(client_cmd))

cvs/lib/inference/vllm_orch.py
```diff
@@ -0,0 +1,317 @@
...
+        log.info("stopping vllm server")
+        self.orch.exec("bash -c 'pkill -f \"vllm serve\" || true'")
+        time.sleep(5)
+
+    # ---------- client side ----------
+
+    def _clone_bench_serving(self, clone_dir="/app"):
+        cmd = (
+            f"bash -c 'mkdir -p {clone_dir} && cd {clone_dir} && "
+            f"(test -d bench_serving || git clone https://github.com/kimbochen/bench_serving.git)'"
```
amd-droy, line 248: is this hard-coded -  https://github.com/kimbochen/bench_serving.git?

cvs/tests/inference/vllm/conftest.py
```diff
@@ -0,0 +1,172 @@
...
+@pytest.fixture(scope="module")
+def hf_token(variant_config):
+    path = variant_config.paths.hf_token_file
+    if not os.path.isfile(path):
+        pytest.skip(f"hf_token file missing: {path}")
+    with open(path) as fp:
+        return fp.read().strip()
+
+
+@pytest.fixture(scope="session")
```
amd-droy: shouldn't the scope be "module"?

cvs/tests/inference/vllm/vllm_single.py
```diff
@@ -0,0 +1,237 @@
...
+        log.info("[fetch poll %d] size=%.1fGB", it, cur / 1e9)
+        if cur > 0 and cur == prev:
+            stable += 1
+            if stable >= 2 or not remote:
+                break
+        else:
+            stable = 0
+        prev = cur
+        if not remote:
+            break
```
amd-droy: you can make the non-remote path explicit and separate from the download-polling loop. if remote=0, let's say for slow mount, the loop breaks on the first iteration, and the test fails even though the model is present

cvs/lib/inference/vllm_orch.py
atnair-amd: Yes, it's hardcoded. It came over from the legacy inference path — on `main`, `base.py` clones the same repo (`benchmark_script_repo`, defaulted to `kimbochen/bench_serving.git` at `base.py:93`, and set explicitly in the inferencemax configs). So this isn't new to the PoC; what's new is that `vllm_orch.py` inlines the URL instead of keeping it config-overridable, so we actually lost the legacy knob here.

Worth flagging that it's more than a hardcoded URL: it's a personal fork (`kimbochen`) of a personal extraction (`kedarpotdar-nv`) of vLLM's Apache-2.0 `benchmark_serving.py`, cloned at HEAD with no pin. And it's *not* stock vLLM — the fork carries calibration-bearing changes (a warmup phase, redefined random range-ratio/seq-length) that our thresholds are tuned against, so we can't just point at the in-image vLLM scripts.

cvs/lib/dtni/config_loader.py
atnair-amd: Good catch on the wording — `runtime` is a nested `RuntimeSpec`, not a flat dict. The code is correct (`container.model_dump()` serialises it to `runtime: {name, args}`, which the conftest does before building the orchestrator), but the docstring read as if it were flat. Reworded the docstring to say so explicitly. (bdac9d0)

cvs/lib/dtni/config_loader.py
atnair-amd: It was leftover scaffolding — never referenced anywhere. Removed the class. (bdac9d0)

cvs/lib/dtni/config_loader.py
atnair-amd: The field was unused in the vllm_single path (only the other suites consume a `benchmark_params` key). Rather than wire in the empty `BenchmarkParams`, I removed both the field and the matching `benchmark_params: {}` in the w1 config — keeping with "don't generalise speculatively." Easy to add back as a typed model when a real consumer lands. (bdac9d0)

cvs/lib/dtni/config_loader.py
atnair-amd: Yes, that's intended — the remote check is deliberately first so an unimplemented remote config fails fast before the threshold-coverage check (which is meaningless for a config we're going to reject). Added a comment documenting the ordering so it's not mistaken for incidental. (bdac9d0)

cvs/lib/dtni/config_loader.py
atnair-amd: Done, adopted your version. (bdac9d0)

cvs/lib/dtni/config_loader.py
atnair-amd: Agreed — switched to `FileNotFoundError` for the missing-config / missing-threshold cases and `ValueError` for the ambiguous-multiple case. (`AssertionError` was also wrong since it's stripped under `python -O`.) (bdac9d0)

cvs/lib/inference/vllm_orch.py
atnair-amd: Adopted — rebuilt the client command as a `shlex.quote`-d arg list, matching the per-field quoting on the server side. (bdac9d0)

cvs/lib/inference/vllm_orch.py
```diff
@@ -0,0 +1,317 @@
...
+        env_script = "\n".join(env_lines) + "\n"
+        # printf the script body verbatim; shlex.quote protects the outer bash layer.
+        self.orch.exec("bash -c " + shlex.quote(f"printf '%s' {shlex.quote(env_script)} > /tmp/server_env_script.sh"))
+        self.orch.exec(f"mkdir -p {shlex.quote(self.out_dir)}")
+
+    def start_server(self):
+        cmd = (
+            f"bash -c 'cd {self.scripts_dir} && source /tmp/server_env_script.sh && "
+            f"nohup /bin/bash {self.server_script} > {self.server_log} 2>&1 &'"
+        )
```
hnimra-amd, line 195: Quote scripts_dir, server_script, and server_log in the bash -c string (same style as shlex.quote elsewhere). Unquoted paths break or are unsafe if they ever contain spaces or shell characters.

hnimra-amd: OK to keep hardcoded for parity with legacy. Can you add a short comment next to the URL in vllm_orch.py pointing to the same rationale you gave here (calibration fork, not stock vLLM, HEAD unpinned risk) so the next reader does not re-litigate it?

cvs/tests/inference/vllm/vllm_single.py
```diff
@@ -0,0 +1,237 @@
...
+    (avoids a double teardown). Runs even if an earlier stage failed -- teardown
+    must happen regardless -- so it does NOT skip on lifecycle.failed.
+    """
+    name = orch.get_container_name(orch.container_config, orch.container_config["image"])
+    t = time.monotonic()
+    orch.teardown_containers()
+    lifecycle.record(request.node.nodeid, "teardown", time.monotonic() - t)
+    lifecycle.torn_down = True
+    if orch.verify_containers_running(name):
+        pytest.fail(f"container {name} still running after teardown_containers()")
```
hnimra-amd, line 248: test_teardown: set lifecycle.torn_down after verify container is gone (or clear flag if verify fails) so finalizer can retry if teardown was incomplete.

cvs/tests/inference/vllm/conftest.py
atnair-amd: Agreed — switched to `scope="module"`. Every sibling fixture here is module-scoped, and the results dict is logically per-module (the table iterates the whole dict), so a session scope would let a second inference module's rows bleed into this one's table. (ad2a3b0)

cvs/tests/inference/vllm/vllm_single.py
atnair-amd: Good catch — fixed. Split the offline/pre-staged path out of the download poll loop: `remote=0` now does an explicit presence check that retries a few times (`_FETCH_PRESENCE_RETRIES`) so a cold/slow mount reading 0 on the first `du` no longer false-fails a model that's actually there. The remote path keeps the kick-download-then-poll-until-stable loop. (ad2a3b0)

cvs/lib/inference/vllm_orch.py
atnair-amd: Done — `shlex.quote`d `scripts_dir`, `server_script`, and `server_log`. Also restructured to `bash -c " + shlex.quote(inner)` so the quoting doesn't nest wrongly inside the old single-quoted `bash -c '...'`, matching the style used in the sibling methods. (ad2a3b0)

atnair-amd: Added a comment next to the URL in `_clone_bench_serving` capturing the rationale: calibration-bearing fork (warmup + redefined range-ratio/seq-length our thresholds are tuned against), not stock vLLM, hardcoded for parity with the legacy `benchmark_script_repo` default, cloned at unpinned HEAD (pin if drift bites). (ad2a3b0)

cvs/tests/inference/vllm/vllm_single.py
atnair-amd: Fixed — `lifecycle.torn_down = True` now runs only after `verify_containers_running` confirms the container is gone. If verify fails the flag stays False so the orch finalizer retries the teardown instead of no-opping on an incomplete one. (ad2a3b0)

amd-droy (approved):
looks good to me, thanks @atnair-amd

PR #225, opened 2026-06-12
Hnimrama/inferencemax uplift

hnimra-amd:
## Summary

This PR uplifts **InferenceMax** on the DTNI track (`inferencemax_single`): suite layout, orchestration, and docs aligned with **`ContainerOrchestrator`**, while tightening **vLLM benchmark** wiring and a few related fixes. It targets **`dev/dtni`** and includes merges from upstream **`dev/dtni`** so the branch stays current.

### InferenceMax / vLLM

- **Suite refactor**: `inferencemax_single` module, thresholds / config loading (`inferencemax_benchmark_model_name`, etc.), discovery rules, and docstrings described as DTNI-style.
- **InferenceX integration**: clone/repo defaults (e.g. SemiAnalysisAI/InferenceX), server launch paths, host-mounted server scripts, `/workspace` handling, and log polling / error patterns for current upstream behavior.
- **Benchmark client**: resolve **`benchmark_serving.py`** (and peers) from the **installed vLLM** package (`vllm/benchmarks/…`) via `BENCH_SCRIPT` instead of maintaining a third-party **`bench_serving`** clone; **`benchmark_script_repo`** treated as legacy where documented.
- **Shared server wrapper**: canonical **`vllm_serve_mi300x.sh`** under `cvs.lib.dtni.vllm_benchmark_scripts` (model-agnostic `MODEL` + optional `vllm_enforce_eager` / CLI flags); sample **`server_script`** paths and **`host_benchmark_scripts_relpath`** updated accordingly.
- **Samples / docs**: InferenceMax JSON examples use **`<changeme>`** for container image/name where appropriate; **`inferencemax.rst`** / **`vllm_singlenode_mi355x.rst`** updated; optional lab-style config under `cvs/input/config_file/inference/inferencemax/`.
- **Docker volumes**: document and sample **`volume_dict`** so **`inferencemax_single`** does not stack two **`:/workspace`** binds with the orchestrator default (avoids **Duplicate mount point** on `docker run`).

### vLLM single-node (`vllm_orch`)

- Merged **`dev/dtni`** behavior: completion detection (**`Serving Benchmark Result`**), **`CLIENT_LAUNCH_FAIL_RE`**, **`client.log`**, while keeping the uplift path that runs **`python3 "$BENCH_SCRIPT"`** from the packaged benchmarks tree.

### Other

- **`run_plugin`**: `--log-file` is optional (no forced default path); tests adjusted.
- **`sglang_disagg_lib`**: parsed results use **`total_generated_tokens`** key consistently.
- Misc. review follow-ups (redaction, regex, ISL/OSL/max length, staging paths, `container_name` on deploy/exec, etc.).

---

## Test plan

- [ ] `cvs run` **`inferencemax_single`** against a 1-node cluster with valid `--cluster_file` / `--config_file` and HF token.
- [ ] Confirm container starts: no **duplicate `/workspace`** in `docker run` when using updated **`volume_dict`** (home bind only unless you intentionally own **`/workspace`** in config and orchestrator skips default).
- [ ] Confirm server uses staged/host **`vllm_serve_mi300x.sh`** (or your override) and client completes using **`benchmark_serving.py`** from the **vLLM image**.
- [ ] Optional: **`vllm_single`** / **`VllmJob`** smoke path if you touch shared **`vllm_orch`** / **`config_loader`** fields.
- [ ] **`cvs run`** / unit tests for **`run_plugin`** if you care about CLI logging defaults.

---

## Notes / follow-ups

- Requires a vLLM environment that ships the **`benchmarks/`** tree next to `vllm`; minimal wheels without it will fail at runtime until a **`vllm bench serve`** migration exists.
- Rebase/merge noise: branch history includes **`Merge … dev/dtni`** commits; review the **net diff vs `dev/dtni`** in the PR “Files changed” tab for the clearest picture.


cvs/cli_plugins/run_plugin.py
```diff
@@ -1,9 +1,12 @@
 import pytest
 import sys
 import os
+from pathlib import Path
 
 from .list_plugin import ListPlugin
 
+_DEFAULT_PYTEST_LOG_FILE = str(Path.home() / ".cvs" / "logs" / "pytest.log")
```
atnair-amd: is this needed?

cvs/lib/dtni/vllm_benchmark_scripts/vllm_serve_mi300x.sh
```diff
@@ -0,0 +1,42 @@
...
+exec vllm serve "${MODEL}" \
+  --host 0.0.0.0 \
+  --port "${PORT}" \
+  --tensor-parallel-size "${TP}" \
+  --max-model-len "${MAX_MODEL_LEN}" \
+  --gpu-memory-utilization "${GPU_MEM}" \
+  --block-size 64 \
+  --no-enable-prefix-caching \
+  "${EAGER_FLAG[@]}" \
+  "$@"
```
atnair-amd, line 43: I like this script, but can we make it more general? I did a way such that it is integrated into vllm_orch (https://github.com/ROCm/cvs/pull/227) maybe we can create a shared folder in dtni called vllm_benchmark scripts? This is so that we do not drift in terms of our benchserving scripts

cvs/input/config_file/inference/inferencemax_single/mi300x_gpt_oss_120b_single/mi300x_gpt_oss_120b_single_config.json
```diff
@@ -0,0 +1,57 @@
+{
+    "_comment": "ISL+OSL must fit 8192 context (7168+1024=8192). CVS expects percentile_metrics (not percentiles_metrics). For vLLM 0.21+ nightlies, InferenceX upstream gptoss_fp4_mi300x.sh calls vllm serve without --enforce-eager, so CUDA graph capture still runs and VLLM_ENFORCE_EAGER alone does not help; this suite defaults use_host_mounted_server_script to the CVS wrapper (CLI --enforce-eager). benchmark_server_script_path \"auto\" uses checkout cvs/input/.../benchmark_server_scripts when present (see host_benchmark_scripts_relpath), else stages under {log_dir}/inference-max/host_scripts_staged/<variant>/ (copy from checkout or bundled). Override with an absolute path if needed. volume_dict maps LOGS/.../workspace to /workspace for server.log on the host. env_dict mirrors key ROCm exports from the InferenceX script.",
+    "config": {
+        "container_image": "rocm/utd-private:vllm-0.21.0-ubuntu24.04-py3.13-nightlies-device-all-cdna-rocm7.14.0a20260602-ad7125a43",
```
atnair-amd: remove internal references

cvs/input/config_file/inference/inferencemax_single/mi300x_gpt_oss_120b_single/mi300x_gpt_oss_120b_single_config.json
```diff
@@ -0,0 +1,57 @@
...
+        "container_name": "inference_max_rocm",
+        "use_host_mounted_server_script": true,
+        "benchmark_server_script_path": "auto",
+        "host_benchmark_scripts_relpath": "input/config_file/inference/inferencemax_single/mi300x_gpt_oss_120b_single/benchmark_server_scripts",
+        "vllm_enforce_eager": true,
+        "nnodes": "1",
```
atnair-amd: we should not specify the node count. later when we implement disagg we can have fields for specifying the prefill and decode sets. For the purpose of single node, the user should not have to specify this

cvs/input/config_file/inference/inferencemax_single/mi300x_gpt_oss_120b_single/mi300x_gpt_oss_120b_single_config.json
```diff
@@ -0,0 +1,57 @@
...
+        "inferencemax_repo": "https://github.com/SemiAnalysisAI/InferenceX.git",
```
atnair-amd: why is this needed

cvs/input/config_file/inference/inferencemax_single/mi300x_gpt_oss_120b_single/mi300x_gpt_oss_120b_single_config.json
```diff
@@ -0,0 +1,57 @@
...
+        "benchmark_script_repo": "https://github.com/kimbochen/bench_serving.git",
```
atnair-amd: this should do nothing given ur benchmark script right

cvs/input/config_file/inference/inferencemax_single/mi355x_inferencemax_gpt_oss_120b_single/mi355x_inferencemax_gpt_oss_120b_single_config.json
```diff
@@ -1,9 +1,9 @@
 {
+    "_comment": "Restored from removed mi355x_inferencemax_gpt_oss_120b_single.json (single node). ISL+OSL within 8192; CVS uses percentile_metrics. Confirm server_script exists in your InferenceX revision (e.g. gptoss_fp4_mi355x.sh under benchmarks/single_node/).",
     "config": {
         "container_image": "rocm/7.0:rocm7.0_ubuntu_22.04_vllm_0.10.1_instinct_20250927_rc1",
         "container_name": "inference_max_rocm",
```
atnair-amd: remove internal references

cvs/tests/inference/inferencemax/conftest.py
```diff
@@ -0,0 +1,224 @@
...
+    def __init__(self, cluster_dict, inference_dict, benchmark_params_dict, model_name):
+        self.cluster_dict = cluster_dict
+        self.inference_dict = inference_dict
+        self.benchmark_params_dict = benchmark_params_dict
+        self.model_name = model_name
+        env_vars = cluster_dict.get("env_vars")
+        node_list = list(cluster_dict["node_dict"].keys())
+        user = cluster_dict["username"]
+        pkey = cluster_dict["priv_key_file"]
+        self.s_phdl = Pssh(log, node_list, user=user, pkey=pkey, env_vars=env_vars)
```
atnair-amd: use orch instead of pssh handles

cvs/tests/inference/inferencemax/conftest.py
```diff
@@ -0,0 +1,224 @@
...
+            timeout=60 * 20,
+        )
+
+        time.sleep(30)
+        out_dict = self.s_phdl.exec("docker ps")
+        for node in out_dict.keys():
+            if not re.search(re.escape(name), out_dict[node], re.I):
+                return False
+        update_test_result()
+        return True
```
atnair-amd: you should not invoke docker_lib directly; everything should be via orch handles. Look at vllm_single's test implementation

atnair-amd:
**`cvs/lib/inference/inferencemax_orch.py` line 397** — poll never exits early

`poll_client_completion` has no early-exit when all nodes complete. The inner `for node` loop skips the sleep when `"End-to-end Latency"` is found, but there is no `break`/`return` after the inner loop finishes, so the outer `for j` loop always runs all `default_client_poll_count` iterations. The function also returns `None` implicitly on success, so callers cannot distinguish "finished on iter 3" from "polled to cap without ever succeeding."

The sibling `VllmJob.wait_client_complete` in `vllm_orch.py` already does this correctly — it uses `orch.exec` (routes into the running container via `cvs/core`) and accumulates a `done[]` list, returning as soon as `all(done)` is true. `poll_client_completion` should follow the same pattern:

```python
def poll_client_completion(self):
    time.sleep(self.default_client_initial_wait)
    for it in range(self.default_client_poll_count):
        out = self.orch.exec(f"tail -30 {shlex.quote(self.client_log)}")
        done = []
        for host, log_tail in out.items():
            log_tail = log_tail or ""
            if _BENCH_CLIENT_TRACEBACK_RE.search(log_tail):
                raise Exception(...)
            m_fail = _BENCH_FAILED_REQUESTS_RE.search(log_tail)
            if m_fail and int(m_fail.group(1)) > 0:
                raise Exception(...)
            done.append(bool(re.search("End-to-end Latency", log_tail, re.I)))
        if done and all(done):
            return
        time.sleep(self.default_client_poll_wait_time)
    raise Exception("client did not complete before poll cap")
```

This also replaces the `c_phdl.exec_cmd_list` per-node fan-out with a single `orch.exec` call, consistent with how the rest of the vllm/inferencemax orchestrators talk to the container.

atnair-amd:
**`cvs/tests/inference/inferencemax/inferencemax_single.py` line 81** — AttributeError masks timeout reason

`poll_for_inference_completion` returns a status dict on the timeout/error paths without ever calling `get_inference_results_dict`, so `self.inference_results_dict` is never assigned. The attribute has no initialization in `__init__` either. Line 81 then does a bare `assert im_obj.inference_results_dict` without checking the return value, so a timeout crashes with `AttributeError: 'InferenceMaxJob' object has no attribute 'inference_results_dict'` instead of the descriptive assertion message.

Suggested fix — check the return value before asserting the attribute:

```python
poll_status = im_obj.poll_for_inference_completion(...)
assert poll_status.get("status") == "success", (
    f"Inference did not complete: {poll_status}"
)
assert im_obj.inference_results_dict, (
    "inference_results_dict empty after benchmark; log parsing likely failed"
)
```

Alternatively, initialize `self.inference_results_dict = {}` in `InferenceMaxJob.__init__` so the bare assert at least produces the intended message rather than an AttributeError.

atnair-amd:
**`cvs/lib/inference/inferencemax_orch.py` line 464** — Median TPOT regex truncates floats

The capture regex for `Median TPOT (ms)` uses `[0-9]+` (integers only) while every adjacent metric — Mean TPOT, P99 TPOT, Median TTFT, Mean TTFT — uses `[0-9\.]+`. If `bench_serving` emits a fractional value like `12.7`, the regex matches but captures only `12`, silently corrupting the metric used in threshold evaluation.

Suggested fix — one character change on line 464:

```python
# before
r'Median TPOT \(ms\):\s+([0-9]+)'
# after
r'Median TPOT \(ms\):\s+([0-9\.]+)'
```

atnair-amd:
**`cvs/lib/inference/base.py` line 160** — `setdefault` typo installs an unread key

`self.bp_dict.setdefault("random_range_ration", "1.0")` has a typo — `ration` instead of `ratio`. The intent is to provide a fallback for configs that omit `random_range_ratio`, but the misspelled key is never read. Both read sites (lines 260 and 435) use the correctly-spelled `bp_dict["random_range_ratio"]`, which has no default, so any config that does not explicitly set it crashes with `KeyError: "random_range_ratio"`. All current configs happen to set it explicitly, making this latent.

Suggested fix:

```python
# before
self.bp_dict.setdefault("random_range_ration", "1.0")
# after
self.bp_dict.setdefault("random_range_ratio", "1.0")
```

atnair-amd:
**`cvs/lib/inference/inferencemax_orch.py` line 441** — raw label string used as dict key

`get_inference_results_dict` stores the `total_generated_tokens` metric under the key `"Total generated tokens:"` (the raw benchmark output label, including the colon and capital letters). Every other metric in the same function uses a normalized snake_case key (`request_throughput_per_sec`, `mean_ttft_ms`, etc.). The same inconsistency exists in `base.py` and `sglang_disagg_lib.py` — this appears to be a copy-paste defect propagated across all three files.

No current threshold file requests `total_generated_tokens`, so it does not fail today. But if anyone adds it to a threshold config using the natural snake_case form, the membership check will silently miss the stored key and the threshold will never be enforced.

Suggested fix — normalize the key to match all others:

```python
# before
self.inference_results_dict[node]["Total generated tokens:"] = match.group(1)
# after
self.inference_results_dict[node]["total_generated_tokens"] = match.group(1)
```

Same fix applies to the identical line in `base.py` and `sglang_disagg_lib.py`.

atnair-amd (changes requested):

cvs/input/config_file/inference/inferencemax_single/mi300x_gpt_oss_120b_single/mi300x_gpt_oss_120b_single_config.json
```diff
@@ -1,15 +1,12 @@
...
+    "_comment": "Single-node InferenceMax. Set container_image to your lab's ROCm + vLLM image. ISL+OSL must fit max_model_length. CVS expects percentile_metrics (not percentiles_metrics). Shared server scripts: cvs.lib.dtni.vllm_benchmark_scripts (host_benchmark_scripts_relpath). benchmark_server_script_path auto stages from checkout or that package.",
     "config": {
-        "container_image": "rocm/utd-private:vllm-0.21.0-ubuntu24.04-py3.13-nightlies-device-all-cdna-rocm7.14.0a20260602-ad7125a43",
-        "container_name": "inference_max_rocm",
+        "container_image": "rocm/7.0:rocm7.0_ubuntu_22.04_vllm_0.10.1_instinct_20250927_rc1",
+        "container_name": "inferencemax_benchmark",
```
atnair-amd: keep this as \<changeme> .   i realized that even mine does not hide the image and name. Ill update my suite respectively

cvs/lib/dtni/vllm_benchmark_scripts/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+
+Keeping scripts here avoids drift between InferenceMax wrappers and vLLM single orchestration.
+'''
+
+from __future__ import annotations
+
+from pathlib import Path
+
+# Same fork used by legacy ``InferenceBaseJob`` / cluster configs (calibration-bearing client).
+BENCH_SERVING_GIT_URL = "https://github.com/kimbochen/bench_serving.git"
```
atnair-amd: i would say remove this entirely since this is a non amd supported script. In this case lets not maintain backwards compat

cvs/lib/dtni/vllm_benchmark_scripts/vllm_serve_mi300x.sh
atnair-amd: I rlly like the script change! 

Is there a reason why it is named gptoss_fp4?  Theoretically i can pass any model into this right? The only refernce to gptoss is line R:31

atnair-amd:
@hnimra-amd LGTM. Squash and Merge it in :)

PR #227, opened 2026-06-16
feat(dtni): move vLLM bench client to stock vllm bench serve (Spec 0)

atnair-amd:
Jira: [AIMVT-238](https://amd-hub.atlassian.net/browse/AIMVT-238)

## Summary

Move the vLLM benchmark client from the unpinned `kimbochen/bench_serving`
fork to the **stock `vllm bench serve` CLI** already baked into the run image
(`rocm/vllm-dev:nightly-sshd`). This de-risks the perf series before anything
calibrates against it: the client is now pinned to the image tag instead of a
third-party `main`. Source change only — no new metrics, no percentile
widening, no threshold calibration (`enforce_thresholds=false` unchanged).

This is the zero-cost moment to switch: thresholds are placeholders and
nothing is calibrated against the fork to lose.

## Changes

**Move to stock (`vllm_orch.py`):**
- `run_client` invokes `vllm bench serve ...` instead of
  `python3 bench_serving/benchmark_serving.py ...`. All `--flag` args unchanged
  (all valid stock flags).
- Deleted `_clone_bench_serving()` — no fork to clone in-image. Its docstring
  also carried a false "calibration-bearing fork" claim; removed.
- Dropped the now-dead `cd /app &&` from `client_cmd`, the
  `self.bench_serv_script` assignment, the `export RESULT_FILENAME=results`
  env line (stock takes the name via `--result-filename`), and renamed
  `client_log` `bench_serv_script.log` -> `client.log`.
- `config_loader.py`: removed `Params.bench_serv_script` field.
- `w1_..._config.json`: removed `bench_serv_script` key (required — `Params`
  is `extra="forbid"`, field and key move in lockstep).

**Robustness for the stock output shape (`vllm_orch.py`):**
- `COMPLETION_RE`: `"End-to-end Latency"` -> `"Serving Benchmark Result"`.
  The old anchor was a per-metric header stock prints only when `e2el` is in
  `--percentile-metrics`; a config omitting e2el would never be detected as
  complete and spin to the poll cap (~90 min). The summary banner is printed
  unconditionally.
- Added `CLIENT_LAUNCH_FAIL_RE` (narrow argparse / not-found markers) and wired
  it into the poll loop. A launch failure (renamed flag, missing `bench`
  subcommand, `vllm` not on PATH) exits via argparse with no traceback and no
  `Failed requests:` line, so the existing crash detector misses it — without
  this the loop hangs to its cap on any launch typo.

## Out of scope (later specs)
- `client.*` metric namespace, percentile widening, goodput, derived metrics — Spec 1.
- `/metrics` server source — Spec 2. QPS sweep — Spec 3.
- The `inferencemax` path (`base.py`) still uses the fork — separate workload, untouched.

## Test plan
- [x] `ruff check` + `ruff format --check` clean; `py_compile` OK; config JSON valid.
- [x] `grep -rn "kimbochen|bench_serv_script|_clone_bench_serving"` clean on the dtni/vllm path.
- [x] Hardware run validated (W1, MI300X).

### DIFF
--- a/cvs/input/config_file/inference/vllm_single/w1_llama31_70b_fp8kv/w1_llama31_70b_fp8kv_config.json
+++ b/cvs/input/config_file/inference/vllm_single/w1_llama31_70b_fp8kv/w1_llama31_70b_fp8kv_config.json
@@ -56,7 +56,6 @@
     "tokenizer_mode": "auto",
     "percentile_metrics": "ttft,tpot,itl,e2el",
     "metric_percentiles": "99",
-    "bench_serv_script": "benchmark_serving.py",
     "num_prompts": "3200",
     "client_poll_count": "90"
   },
--- a/cvs/lib/dtni/config_loader.py
+++ b/cvs/lib/dtni/config_loader.py
@@ -123,7 +123,6 @@ class Params(_Forbid):
     tokenizer_mode: str = "auto"
     percentile_metrics: str = "ttft,tpot,itl,e2el"
     metric_percentiles: str = "99"
-    bench_serv_script: str = "benchmark_serving.py"
     num_prompts: str = "3200"
     # Completion-poll budget for the bench client = client_poll_count * 60s
     # (plus a 120s initial wait). Large-output cells (high osl) need a bigger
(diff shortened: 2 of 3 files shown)


amd-droy (approved):
looks good to me. thanks @atnair-amd

PR #229, opened 2026-06-18
Hnimrama/inferencemax uplift restore

hnimra-amd:
## Summary

- Restores the InferenceMax / DTNI uplift from #225 after it was reverted by #228.
- Adds a follow-up fix so the vLLM benchmark client runs with the same Python that imports `vllm` (ROCm images where `python3` is not the vLLM interpreter), and hardens `docker exec` / `bash -c` quoting.

## Test plan

- [ ] `cvs run inferencemax_single` (or your usual InferenceMax path) on MI300x + current vLLM container.


cvs/tests/inference/inferencemax/inferencemax_single.py
```diff
@@ -0,0 +1,249 @@
...
+    lifecycle.record(request.node.nodeid, "container_launch", time.monotonic() - t)
+    if not ok:
+        lifecycle.failed = True
+        name = orch.get_container_name(orch.container_config, orch.container_config["image"])
+        pytest.fail(f"setup_containers() returned False for {name}")
+    name = orch.get_container_name(orch.container_config, orch.container_config["image"])
+    if not orch.verify_containers_running(name):
+        lifecycle.failed = True
+        pytest.fail(f"container {name} not running after setup_containers()")
+    time.sleep(30)
```
amd-droy, line 86: will you need this gap of 30 secs after the testcase?

cvs/tests/inference/inferencemax/inferencemax_single.py
```diff
@@ -0,0 +1,249 @@
...
+                for short, _unit in _METRICS:
+                    metric_cases.append((combo, c, short))
+                    metric_ids.append(cid + "-" + short)
+            metafunc.parametrize("seq_combo,concurrency,metric", metric_cases, ids=metric_ids)
+    elif "seq_combo" in metafunc.fixturenames and "concurrency" in metafunc.fixturenames and cases:
+        metafunc.parametrize("seq_combo,concurrency", cases, ids=ids)
+
+
+def _du_bytes(orch, path):
+    out = orch.exec(f"bash -c {shlex.quote(f'du -sb {shlex.quote(path)} 2>/dev/null | cut -f1')}")
```
amd-droy, line 65: The 2>/dev/null silently swallows errors, and the function returns 0 on any failure. Since test_model_fetch uses the return value of 0 to detect "no model present," a network failure, permission error, or du not being available would be indistinguishable from a genuinely missing model

cvs/tests/inference/inferencemax/inferencemax_single.py
```diff
@@ -0,0 +1,249 @@
...
+    try:
+        job.stop_server()
+        job.build_server_cmd()
+        t = time.monotonic()
+        job.start_server()
+        job.wait_ready()
+        lifecycle.record(request.node.nodeid, "server_ready", time.monotonic() - t)
+        t_client = time.monotonic()
+        job.run_client()
+        job.wait_client_complete()
```
amd-droy, line 192: better to have a timeout here. If the inference client hangs, this will block indefinitely with no recovery path.

cvs/tests/inference/inferencemax/conftest.py
```diff
@@ -0,0 +1,153 @@
...
+@pytest.fixture(scope="module")
+def hf_token(variant_config):
+    path = variant_config.paths.hf_token_file
+    if not os.path.isfile(path):
+        pytest.skip(f"hf_token file missing: {path}")
+    with open(path) as fp:
+        return fp.read().strip()
+
+
+@pytest.fixture(scope="session")
```
amd-droy, line 92: will the scope be session or module?

anujmittal-amd (changes requested):
Can we add your changed under new folder structure? Inferencemax has been depreciated, you should create folder under inference as atom or reuse the vllm/sglang structure and add atom in that location. Lets discuss this. 
Same comment for single/distributed; do we need a separate folder for single/multi? 
dtni folder should be renamed based on vLLM structure.

Once you are ready for final merge, please add your end to end test results to the PR along with original inferenceMax test run to identify there are no regressions. 

Thanks

hnimra-amd:
> Can we add your changed under new folder structure? Inferencemax has been depreciated, you should create folder under inference as atom or reuse the vllm/sglang structure and add atom in that location. Lets discuss this. Same comment for single/distributed; do we need a separate folder for single/multi? dtni folder should be renamed based on vLLM structure.
> 
> Once you are ready for final merge, please add your end to end test results to the PR along with original inferenceMax test run to identify there are no regressions.
> 
> Thanks

@anujmittal-amd I addressed the structural feedback on the IX-atom branch, this is the PR for it: [https://github.com/ROCm/cvs/pull/238](https://github.com/ROCm/cvs/pull/238 ) 
and it builds on the current PR

PR #232, opened 2026-06-19
changing the command to run TB schmoo to get better numbers

urtiwari:
changing the command to run TB schmoo to get better numbers, added flag to the existing command


anujmittal-amd (approved):
LGTM

PR #233, opened 2026-06-19
Restructure shared/inference libs + sweep selector + artifact-based vllm result path

atnair-amd:
## Summary

Splits the inference suite's library along a generic/inference seam, replaces
the sweep's cartesian explosion with an explicit cell selector, and rebuilds
the `vllm_single` result path to read the stock benchmark artifact and assert a
declared SLO contract. CVS is experimental and single-framework today; this is
a hard cut to the new shape (no back-compat path).

New to the package? Start with the authoring guide:
[`plans/building-a-cvs-test-suite.md`](https://github.com/ROCm/cvs/blob/atnair/vllm-client-metrics-from-artifact/plans/building-a-cvs-test-suite.md).
Each package also has an `AGENTS.md` describing its public surface.

## What changed

**Library boundary** — `cvs/lib/dtni` (a leftover codename) is renamed to
`cvs/lib/utils` for the truly framework-agnostic half: the
`paths`/`model`/`container` schema, `BaseVariantConfig`, placeholder
substitution, and `verdict.evaluate_all`. The inference-specific half moves into
a new `cvs/lib/inference/utils/` package: the `client.*` artifact parser
(`vllm_parsing.py`) and the inference config schema
(`inferencing_config_loader.py`: `Sweep`/`Run`/`Params`/`Roles`/`cell_key`).

**Sweep selector** — the old `sequence_combinations × concurrency_levels`
cartesian is replaced by named combos plus an explicit `runs[]` list that
cherry-picks exactly the `(combo, concurrency)` cells you want. Unknown or
duplicate combo names fail at load time (and at collection time, via a shared
validator) instead of silently dropping a run.

**Self-contained server launch** — the `vllm serve` command is now built in
Python (mirroring how the client command is built), so a run no longer depends
on an external `.sh` script that had to be hand-staged onto the node.
Per-model server quirks live in config as `roles.server.serve_args` (a
`{flag: value}` map) and `roles.server.env`.

**Artifact-based parsing** — `parse_results` reads the benchmark's stock
`results` JSON artifact via `to_client_metrics` instead of scraping the console
log with regex. The suite emits one HTML row per metric per cell with inline
Value/Unit columns.

**Single image source** — the image is declared once on `container.image`. The
duplicated top-level `image` block (which conftest silently copied over the
container block) is gone.

**Declared SLO contract** — `GATED_METRICS` in `vllm_parsing.py` names the
subset of metrics a calibrated run must *assert* (throughput, the full latency
distribution — mean/median/p90/p95/p99 per ttft/tpot/itl/e2el — and run
health). The loader's coverage check now enforces two axes: every sweep cell has
a threshold entry, **and** every present cell has a spec for every gated metric.
This closes the gap where a gated metric with no threshold would fall through to
a zero-assertion "record-only" row and report PASS. A new `max` verdict kind
supports unit-agnostic count ceilings (e.g. `failed`).

**Config cleanup** — the config/threshold pair is renamed to
`{model}_{precision}_{config|threshold}.json` and converted to the selector.
The bespoke `mi300x_vllm_single.json` cluster file is deleted: a cluster file
only needs the node IP, since the variant config supplies the container block.


cvs/lib/inference/unittests/test_inferencing_config_loader.py
```diff
@@ -0,0 +1,197 @@
...
+    """The gated-metric axis of _check_thresholds_cover_sweep."""
+
+    _CELL = "ISL=128,OSL=2048,TP=8,CONC=16"
+
+    def _variant_with(self, thresholds, enforce):
+        sw = Sweep(
+            sequence_combinations=[_combo("a")],
+            runs=[Run(combo="a", concurrency=16)],
+        )
+        vc = _variant(sw)  # remote guard ok; enforce/thresholds set below
```
amd-droy: is vc used anywhere? otherwise might give a linter issue.

cvs/lib/inference/unittests/test_inferencing_config_loader.py
```diff
@@ -0,0 +1,197 @@
...
+        del specs["client.p99_ttft_ms"]  # drop one gated metric
+        with self.assertRaises(ValidationError) as ctx:
+            self._variant_with({self._CELL: specs}, enforce=True)
+        self.assertIn("missing gated-metric specs", str(ctx.exception))
+        self.assertIn("client.p99_ttft_ms", str(ctx.exception))
+
+    def test_missing_gated_metric_warns_when_record_only(self):
+        specs = _full_gated_specs()
+        del specs["client.failed"]
+        import warnings as _w
```
amd-droy: can be moved to the import section at the top of the file

cvs/lib/inference/unittests/test_vllm_orch_parse.py
```diff
@@ -0,0 +1,527 @@
...
+
+    def test_real_value_passes_through(self):
+        cell = self._cell()
+        self.assertEqual(cell({"client.goodput": 4.76}, "client.goodput"), 4.76)
+
+    def test_zero_is_not_dashed(self):
+        # 0.0 is a real measurement, must NOT become '-'.
+        cell = self._cell()
+        self.assertEqual(cell({"client.request_throughput": 0.0}, "client.request_throughput"), 0.0)
+
```
amd-droy, line 261: _cell() method reads the file from disk, constructs a module spec, and executes the module body on every single test method invocation. The four test methods in this class therefore load _shared.py four times.

You can try this way - load the module once in setUpClass and cache _cell as a class attribute:
class TestTableCellRendering(unittest.TestCase):
    _cell_fn = None  # populated once by setUpClass

    @classmethod
    def setUpClass(cls):
        import importlib.util
        import sys as _sys
        from types import ModuleType

        _sys.modules.setdefault("tabulate", ModuleType("tabulate"))
        if not hasattr(_sys.modules["tabulate"], "tabulate"):
            _sys.modules["tabulate"].tabulate = lambda *a, **k: ""

        spec = importlib.util.spec_from_file_location("_shared_under_test", str(_SHARED))
        mod = importlib.util.module_from_spec(spec)
        spec.loader.exec_module(mod)
        cls._cell_fn = staticmethod(mod._cell)

    def test_present_but_none_renders_dash(self):
        self.assertEqual(self._cell_fn({"client.goodput": None}, "client.goodput"), "-")

cvs/lib/inference/unittests/test_vllm_orch_parse.py
```diff
@@ -0,0 +1,527 @@
...
+                "value": 0.5,
+            }
+        }
+        with self.assertRaises(ThresholdViolation) as ctx:
+            evaluate_all(actuals, thresholds)
+        self.assertIn("is None", str(ctx.exception))
+
+
+if __name__ == "__main__":
+    unittest.main()
```
amd-droy: wont the file end here?

cvs/lib/inference/utils/inferencing_config_loader.py
```diff
@@ -0,0 +1,219 @@
...
+    time) and `pytest_generate_tests` (collection time, which reads raw JSON
+    before the loader runs) so the two can never drift. Operates on plain lists
+    of strings so both call sites can feed it.
+
+    Without it a duplicate name silently shadows a combo and a typo'd
+    `run.combo` is a silently-dropped cell -- either way the sweep runs a
+    different matrix than the config reads.
+    """
+    names = list(combo_names)
+    dupes = sorted({n for n in names if names.count(n) > 1})
```
amd-droy: Time complexity will be O(n²) if the list grows, if it small then its ok. This might be faster - 

from collections import Counter

counts = Counter(combo_names)
dupes = sorted(name for name, count in counts.items() if count > 1)
if dupes:
    raise ValueError(f"duplicate sequence_combination names: {dupes}")

cvs/input/config_file/inference/vllm_single/w1_llama31_70b_fp8kv/llama31_70b_fp8_threshold.json
```diff
@@ -0,0 +1,97 @@
+{
+  "_comment": "PLACEHOLDER thresholds for W1 (Llama 3.1 70B FP8-KV, in=128/out=2048, TP=8, conc=16) -- the single cell the sweep enumerates. Every gated metric (vllm_parsing.GATED_METRICS: throughput, the full latency distribution mean/median/p90/p95/p99 per ttft/tpot/itl/e2el family, and success_rate/failed health) has a spec here so the loader gated-coverage check passes. Values are NOT calibrated (AMD publishes throughput-vs-latency curves, not per-cell scalars). config.json sets enforce_thresholds=false: the suite records metrics and skips pass/fail. Replace placeholders with real numbers and flip enforce_thresholds=true to assert.",
+  "ISL=128,OSL=2048,TP=8,CONC=16": {
```
anujmittal-amd: Is this correct? ISL of 128? this is the ISL/OSL matrix that we should be automating: 
ISL/OSL: 1K / 1K
ISL/OSL: 8K / 1K
ISL/OSL: 1K / 8K
ISL/OSL: 1K / 4K
ISL/OSL: 5000 / 1024

anujmittal-amd (changes requested):
Can you please add the test results to this PR?

Also, please run the original vLLM suites from the dtni/dev branch to confirm that the refactored code has not introduced any regressions.

As discussed during our live review, please add a reference to threshold_config.json in vllm_config.json.

Additional follow-ups:

Create a Jira for generating/adding threshold_config.json via cvs cli.
Add a README explaining the refactor so that others can understand the changes and usage.

cvs/lib/inference/unittests/test_inferencing_config_loader.py
atnair-amd: Resolved.

cvs/lib/inference/unittests/test_inferencing_config_loader.py
atnair-amd: Resolved.

cvs/lib/inference/unittests/test_vllm_orch_parse.py
atnair-amd: Resolved.

cvs/lib/inference/unittests/test_vllm_orch_parse.py
atnair-amd: Adopted your exact pattern. `setUpClass` now loads `_shared.py` once into `cls._cell_fn` (with the tabulate stub), and all four test methods call `self._cell_fn(...)` directly. Applied the same fix to `TestKeyConsistency` (`_producer_keys` → `setUpClass` caching `cls._produced`) and `TestMetricTests` (`setUp` → `setUpClass` for `cls.vs = _load_vllm_single()`) — they had the same per-invocation reload problem.

cvs/lib/inference/utils/inferencing_config_loader.py
atnair-amd: Adopted — `Counter` is imported at the top and `validate_sweep_selector` now uses it exactly as you suggested.

cvs/lib/inference/unittests/test_vllm_orch_parse.py
amd-droy: LGTM. And good catch extending it to TestKeyConsistency and TestMetricTests as well. It's cleaner to fix them all consistently rather than leaving partial coverage. No further changes needed from my side.

amd-droy (approved):
Looks good to me. Once you resolve Anuj's comment, it's good to merge. Thanks @atnair-amd.

cvs/input/config_file/inference/vllm_single/w1_llama31_70b_fp8kv/llama31_70b_fp8_config.json
```diff
@@ -58,9 +58,49 @@
   "sweep": {
     "sequence_combinations": [
       {
-        "name": "w1_isl=128_osl=2048",
-        "isl": "128",
-        "osl": "2048",
+        "name": "w1_isl=1024_osl=1024",
```
anujmittal-amd: Atul, this was not the ask. The comment was to change the ISL/OSL to the values mentioned in the DTNI Tracker sheet (please check the comments section of each workload). We dont need to run every workload for all ISL/OSL combination. Thanks

atnair-amd: With this Sweep you can define multiple sequence combinations and run them based on the run keyword. I have updated the config and threshold.json to reflect the seq combinations in the sheet

atnair-amd: [vllm_single_2026-06-23T212659.zip](https://github.com/user-attachments/files/29274779/vllm_single_2026-06-23T212659.zip)

solaiys:
No new blocking code issues found.
I reviewed the high-risk surfaces and did not find a new blocking defect:

**Summary:**
vllm_parsing.py — _safe_div guards every divisor (None/0/missing → None), derived metrics degrade cleanly. Sound.
verdict.py — all threshold kinds (min/max/max_ms/within/min_tok_s/min_ratio) handle None, missing reference, and divide-by-zero with explicit violations rather than TypeError. Sound.
config_loader.py / inferencing_config_loader.py — 3-pass substitution is bounded; sweep-selector validation (dup names / unknown run.combo) is shared between the typed validator and pytest_generate_tests so they can't drift; threshold-coverage gate is enforced. Sound.
vllm_single.py (driver) — readiness/completion/launch-failure regexes are deliberately narrow; commands are shlex.quoted; per-cell out_dir prevents stale-artifact reads. Sound.
Metric-gate wiring — the inf_res_dict key built in test_vllm_inference is byte-for-byte identical to the lookup key in test_metric, so the new per-metric gate does not silently skip assertions.
Rename safety — cvs.lib.dtni.* and vllm_orch/vllm.py have no remaining importers; the renames are complete, so no import-time break.

cvs/input/config_file/inference/vllm_single/w1_llama31_70b_fp8kv/llama31_70b_fp8_config.json
solaiys: by looking at this results, This is for just one inference test with W1 line item. 

1. Are we capturing all the logs during the inference? the logfile/html for inference seems to be very small.
2. So many tests for metrics validation without log . I guess its just comparison against the expected value from the threshold.json. 
3. Is this result for the complete test for W1 line item? or just one combo?
5. Just curious, If we enable all the seq. combo, how the html report will look like?

atnair-amd: @solaiys the above run is for only 1k/1k . 

1. Yes all the logs are being captured; i do have to do a followup PR so it shows progress in the html. Currently it gets cut off due to length
2. I can create html like print results table ; the thing is that a lot of the tests are one off metrics (like calculating percentiles, median, mean, etc).
3. i did  a run w/ 1k/1k to prove it works  end 2 end
4. the html report will be combined with both parametrizations. The setup, teardown, and results table will be shared; the run files themselves will be different

I reran with 1k/1k and 8k/1k  to illustrate what i mean. 
[vllm_single_2026-06-24T130335.zip](https://github.com/user-attachments/files/29302419/vllm_single_2026-06-24T130335.zip)

anujmittal-amd (approved):
LGTM: but can you please add the run results for the orginal vLLM run (original cvs vLLM code). Thanks

atnair-amd:
[vllm_single_2026-06-24T130335.zip](https://github.com/user-attachments/files/29313687/vllm_single_2026-06-24T130335.zip)

PR #234, opened 2026-06-19
Framework change implementation for sglang

amd-droy:
Implementation:
Merged all the different model test files into one - sglang_disagg_distributed.py.
conftest.py - loads and parses --config_file and the threshold-json for each model.
_shared.py - reporting for all the tests, smoke, accuracy and performance.
cvs\lib\utils\model_query_lib.py - for smoke tests with openai api endpoints and lm-eval tests.
cvs\lib\inference\sglang_disagg_lib.py - implemention for smoke tests and lm-eval gsm8k, hellaswag and mmlu tasks. 
sglang_disagg_distributed.py - single test file for all tests in sglang. All models will run the tests from this file.

Threshold jsons: file name comes from --config_file 
"threshold_file": "cvs/input/config_file/inference/sglang/mi30x_sglang_llama_70b_threshold.json",

cvs\input\config_file\inference\sglang\mi30x_sglang_llama_70b_threshold.json
cvs\input\config_file\inference\sglang\mi30x_sglang_deepseek_r1_0528_threshold.json
cvs\input\config_file\inference\sglang\mi30x_sglang_gpt_oss_120b_threshold.json

Reports - 
[reports_sglang.zip](https://github.com/user-attachments/files/29313413/reports_sglang.zip)


cvs/lib/sglang_disagg_lib.py
```diff
@@ -267,15 +287,16 @@ def exec_nic_setup_scripts(
...
             pout_dict = self.p_phdl.exec(cmd)
             dout_dict = self.d_phdl.exec(cmd)
             for node in pout_dict.keys():
-                if not re.search('hca_id:\s+bnxt_', pout_dict[node], re.I):
+                if not re.search('hca_id:\s+(bnxt_|rocep)', pout_dict[node], re.I):
                     log.info("%s", pout_dict[node])
                     fail_test(f'Broadcom libbnxt rdma driver is not properly copied on node {node}')
             for node in dout_dict.keys():
-                if not re.search('hca_id:\s+bnxt_', dout_dict[node], re.I):
+                if not re.search('hca_id:\s+(bnxt_|rocep)', dout_dict[node], re.I):
```
anujmittal-amd: Can we move this to the config.json where based on the nic type the search regex would be formulated? lets not hardcode it here.

amd-droy: done

cvs/lib/utils/model_query_lib.py
```diff
@@ -0,0 +1,543 @@
...
+        model_args_q = shlex.quote(
+            cls.build_model_args(
+                model_id=model_id,
+                base_url=base_url,
+                num_concurrent=num_concurrent,
+                extra_model_args=extra_model_args,
+            )
+        )
+        parts: list[str] = []
+        if pip_install:
```
anujmittal-amd, line 365: As discussed; please move the lm-eval install and instantiation to a common library under lib/inference

cvs/lib/inference/sglang_disagg_lib.py
```diff
@@ -267,15 +287,16 @@ def exec_nic_setup_scripts(
             self.nccl_ib_gid_index = 3
             cmd = f'docker exec {self.container_name} /bin/bash -c "sudo \
                     cp /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so.host \
-                    /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so;" '
+                    /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so; \
```
anujmittal-amd: This should come from the config.json and should not be hardcoded.

cvs/lib/sglang_disagg_lib.py
```diff
@@ -600,6 +608,9 @@ def launch_decode_servers(self, dtype='auto', kv_cache_dtype='auto'):
                               --dtype {dtype} \
                               --kv-cache-dtype {kv_cache_dtype} \
                               --tp {self.bp_dict['tensor_parallelism']} \
+                              --nnodes {self.decode_nnodes} \
+                              --node-rank {i} \
+                              --dist-init-addr {dist_init_addr} \
                               --disable-radix-cache --disable-cuda-graph \
                               --mem-fraction-static {self.bp_dict['memory_fraction']} \
                               --attention-backend aiter \
```
anujmittal-amd, line 655: This should also come via config.json; as this could be changed.

anujmittal-amd (commented):
Once you are ready to merge, can you please add the test results to this PR?

Also, please run the original SGLang suite from the dtni/dev branch to confirm that the refactored code has not introduced any regressions.

Please add a reference to threshold_config.json in sgLang_config.json.

Instead of commom-config; lets have it is following format:
platform_framework_model_single|distributed_config.json
And same with Threshold_config.json

amd-droy:
> Once you are ready to merge, can you please add the test results to this PR?
> 
> Also, please run the original SGLang suite from the dtni/dev branch to confirm that the refactored code has not introduced any regressions.
> 
> Please add a reference to threshold_config.json in sgLang_config.json.
> 
> Instead of commom-config; lets have it is following format: platform_framework_model_single|distributed_config.json And same with Threshold_config.json

there are no regressions coz, benchmark logic hasn't been changed.
threshold_config.json added.
common_config.json removed
reports attached

cvs/lib/utils/model_query_lib.py
amd-droy: moved here - cvs\lib\utils\model_query_lib.py
moved here - cvs\lib\inference\sglang_disagg_lib.py

model_query_lib.py - has all the openai api smoke tests and the lm-eval tests. All generalized. Can be used by all the other inference frameworks.

cvs/lib/inference/sglang_disagg_lib.py
amd-droy: this was always hardcoded here. I dont think this will need to be updated. Does it need to go in the config?

cvs/lib/sglang_disagg_lib.py
amd-droy: its coming from config only. here - dist_init_addr = f"{self.inf_dict['prefill_coordinator_addr']}:{self.inf_dict['prefill_coordinator_port']}"

solaiys (commented):
Re-reviewed the updated SGLang framework-change PR. The two earlier issues (lm-eval `KeyError` from missing `lm_eval_*` config blocks, and `active_benchmark: null` with multiple variants) are now resolved — thanks.

One **blocking** correctness issue remains on the accuracy threshold path (inline comment on `conftest.py`), plus a minor naming nit on the misspelled `theshold_file` alias.

Also re-raising @anujmittal-amd's earlier request: could you please **attach the actual run-results output** to this PR (the lm-eval accuracy tables + `bench_serv` perf tables / logs), and a run of the original SGLang suite from `dtni/dev` to confirm no regression? A real run is exactly what would surface the accuracy-gate issue above.

cvs/tests/inference/sglang/conftest.py
```diff
@@ -0,0 +1,295 @@
...
+    """Build an lm-eval threshold cell key, e.g. ``BENCH=lm_eval_hellaswag``."""
+    return f"BENCH={bench_name}"
+
+
+def flat_expected_from_specs(specs: Mapping[str, Any]) -> dict[str, float]:
+    """Convert ``{metric: {kind, value}}`` threshold specs to flat floats for legacy checks."""
+    out: dict[str, float] = {}
+    for metric, spec in specs.items():
+        if isinstance(spec, dict) and "value" in spec:
+            out[metric] = float(spec["value"])
```
solaiys, line 89: **Blocking (correctness): `kind` is discarded here, so the declared threshold direction is not enforced for accuracy.**

`flat_expected_from_specs` keeps only `value` and drops `kind`. The accuracy path (`cvs/lib/utils/model_query_lib.py` `LmEvalBenchmark.check_results`) then applies a **symmetric** window `abs(actual - expected) > tolerance_frac * expected` (default 5%), not the `min` floor the threshold files declare.

Effect: the threshold files set `acc_norm,none: {kind: min, value: 0.23}` (a floor), but a healthy Llama-3.1-70B (~0.85 HellaSwag) lands outside `0.23 ± 5%` and **fails a correct run**. Same for MMLU (floor 0.29 vs real ~0.82). Note `kind` is purely decorative today — this suite never calls the kind-aware `verdict.evaluate_all`.

Suggested fix: carry the spec (`kind`+`value`) through to the checker and evaluate by kind — `min`/`min_tok_s` => pass if `actual >= value`, `max`/`max_ms` => pass if `actual <= value`. If a symmetric +/-tolerance check is genuinely intended for accuracy, then the kind should not be `min` and the values must be the real measured accuracies rather than floors.

cvs/tests/inference/sglang/conftest.py
```diff
@@ -0,0 +1,295 @@
...
+)
+
+from cvs.tests.inference.sglang._shared import SGLANG_DISAGG_TEST_ORDER, resolve_benchmark_variant_key
+
+log = globals.log
+
+
+def _threshold_file_path(bp_dict: Mapping[str, Any]) -> str | None:
+    """Return the threshold file path from a benchmark_params variant (typo-tolerant)."""
+    for key in ("theshold_file", "threshold_file"):
```
solaiys: Minor (naming): the misspelled key `theshold_file` (missing `r`) is accepted as an alias here, and echoed in the docstrings at lines 10 and 178. Recommend accepting only the correctly-spelled `threshold_file` and dropping the misspelling, so a typo'd config key fails loudly instead of silently working. The shipped config already uses the correct `threshold_file`.

cvs/lib/inference/sglang_disagg_lib.py
solaiys: YES. 
In config file, its the mount volume for the docker.
basically, the source .so file is mounted as .so.host in to the docker. Then inside the docker this code will copy the .so.host file in to .so file.

This is done to avoid the mount failure if the file is existing inside the docker.

cvs/tests/inference/sglang/conftest.py
solaiys: @amd-droy 
I got this review comment from cursor. Please double check this and respond if its a real bug.
I am not sure about the statement " a healthy Llama-3.1-70B (~0.85 HellaSwag)"
So I have raised this comment.

cvs/lib/inference/sglang_disagg_lib.py
amd-droy: done. 
config now has - "mount_vol": "/usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so",

cvs/tests/inference/sglang/conftest.py
amd-droy: updated. good catch, did not notice it's still there.

cvs/tests/inference/sglang/conftest.py
amd-droy: yes, this part I will update according to Atuls's PR in my next PR. I just updated the config accordingly, but yet to implement verdict.evaluate_all function. As Atul's PR is now merged, I will update the config implementation in my next PR.

cvs/input/config_file/inference/sglang/mi30x_sglang_distributed.json
```diff
@@ -0,0 +1,408 @@
...
+	"proxy_router_serv_port": "8000",
+        "container_config":
+        {
+	  "device_list": [ "/dev/dri", "/dev/kfd", "/dev/infiniband/rdma_cm" ],
+	  "volume_dict":
+	  {
+	      "/home/{user-id}": "/home/{user-id}",
+	      "/mnt/dtni/models": "/root/models",
+	      "/dev/infiniband": "/dev/infiniband",
+		  "mount_vol": "/usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so",
```
hnimra-amd: Suggestion: move `mount_vol` out of `volume_dict` into a top-level config field (next to `nic_type` / `hca_id_prefix`). Less risk of accidentally passing it to Docker as a mount.

cvs/tests/inference/sglang/conftest.py
```diff
@@ -0,0 +1,300 @@
...
+    if isinstance(inference_config_root, dict) and "config" in inference_config_root:
+        cfg = inference_config_root["config"]
+    else:
+        cfg = inference_config_root
+    cfg = resolve_test_config_placeholders(cfg, cluster_dict)
+    volume_dict = cfg['container_config']['volume_dict']
+    cfg['mount_vol'] = volume_dict.pop(
+        'mount_vol',
+        '/usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so',
+    )
```
hnimra-amd: Suggestion: move `mount_vol` out of `volume_dict` into a top-level config field (next to `nic_type` / `hca_id_prefix`). Less risk of accidentally passing it to Docker as a mount.

cvs/tests/inference/sglang/sglang_disagg_distributed.py
```diff
@@ -0,0 +1,186 @@
...
+from cvs.lib import docker_lib, globals
+from cvs.lib.utils_lib import fail_test, update_test_result
+
+_spec = _ilu.spec_from_file_location(
+    "_sglang_disagg_shared", _pl.Path(__file__).with_name("_shared.py")
+)
+_mod = _ilu.module_from_spec(_spec)
+_spec.loader.exec_module(_mod)
+
+test_print_results_table = _mod.test_print_results_table
```
hnimra-amd: I doubt we need teh entire importlib to load `_shared.py`? A normal import should work: `from cvs.tests.inference.sglang._shared import test_print_results_table`

cvs/lib/inference/sglang_disagg_lib.py
```diff
@@ -1259,3 +1345,300 @@ def verify_inference_results(self, test_name, expected_result_dict):
...
+        if not ok:
+            summary = OpenAIProbe.summarize_results(results, ok, err)
+            fail_test(f"{err}")
+            return summary
+
+        summary = OpenAIProbe.summarize_results(results, ok, err)
+        return summary
+
+ 
+    def run_lm_eval_hellaswag_benchmark_test(self, _d_type="auto"):
```
hnimra-amd: Duplicate lm-eval test methods
Suggestion: factor `run_lm_eval_*` into one helper (e.g. `_run_lm_eval(task_name, default_metric, ...)`) to cut duplication and keep behavior in sync.

cvs/lib/inference/sglang_disagg_lib.py
```diff
@@ -837,18 +860,54 @@ def benchserv_test_random(self, d_type='auto'):
         cmd = f'''docker exec {self.container_name} /bin/bash -c  "
                       mkdir -p {self.log_dir}/benchmark_node; \
                       source /tmp/benchmark_env_script.sh && \
+                      pip install --upgrade --no-deps sglang[all] && \
```
hnimra-amd, line 838: `pip install` on every perf run
`pip install --upgrade --no-deps sglang[all]` runs on every `benchserv_test_random` call. Can this move to container setup (or be config-gated) to avoid extra runtime?

hnimra-amd (commented):
Nice refactor. Unified test file and shared `model_query_lib` will help other inference frameworks too.

**Minor changes needed:** simplify `_shared` import, consider deduping the three lm-eval runners, consider moving `pip install` out of the perf hot path.

cvs/tests/inference/sglang/conftest.py
amd-droy: done.

cvs/input/config_file/inference/sglang/mi30x_sglang_distributed.json
amd-droy: done.

cvs/tests/inference/sglang/sglang_disagg_distributed.py
amd-droy: true, removed it.

cvs/lib/inference/sglang_disagg_lib.py
amd-droy: great suggestion. Code updated. kept 1 helper function and calling the same function for all the 3 tasks. Code looks much cleaner now.

cvs/lib/inference/sglang_disagg_lib.py
amd-droy: this is just a stop-gap arrangement as none of container images has the latest sglang installed for which the perf benchmarking is failing. Once we have a working sglang image, this "pip install" will be removed. For now, it's just running once when the perf test runs. The perf test runs and collects all the data at once.

anujmittal-amd (approved):
LGTM, Please go ahead.

PR #240, opened 2026-06-24
Removed examples from TB and added a tag to build TB from that repo t…

urtiwari:
Removed the TransferBench example tests (1–6) from the test code, health config, and docs.

Pinned TransferBench installs to v1.67.00 via a new git_tag config field and post-clone checkout in install_transferbench.py.

Added GFX_TEMPORAL=3 and GFX_UNROLL=32 to the scaling test command.

### DIFF
--- a/cvs/tests/health/install/install_transferbench.py
+++ b/cvs/tests/health/install/install_transferbench.py
@@ -183,6 +183,7 @@ def test_install_transferbench(orch, config_dict):
     Steps:
       - Ensure git_install_path exists on all nodes.
       - Clone TransferBench repo under git_install_path on every node.
+      - Checkout the configured git tag.
       - Build on all nodes using detected ROCM_PATH and HIPCC.
       - Verify the build artifact is present on each node.
 
@@ -193,6 +194,7 @@ def test_install_transferbench(orch, config_dict):
       config_dict (dict): Includes:
         - git_install_path: directory to clone/build
         - git_url: repository URL
+        - git_tag: git tag to checkout after clone
     """
 
     globals.error_list = []
@@ -213,6 +215,13 @@ def test_install_transferbench(orch, config_dict):
         timeout=120,
     )
 
+    tb_src = f'{git_install_path}/TransferBench'
+    git_tag = config_dict['git_tag']
+    out_dict = orch.exec(
+        f"bash -c 'cd {tb_src} && git checkout {git_tag}'",
+        timeout=120,
+    )
+
     # Detect ROCm path and compiler
     rocm_path = detect_rocm_path(orch, config_dict.get('rocm_path', ''))
     hip_compiler = detect_hip_compiler(orch, rocm_path)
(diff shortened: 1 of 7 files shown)


cvs/tests/health/install/install_transferbench.py
```diff
@@ -213,6 +215,13 @@ def test_install_transferbench(orch, config_dict):
         timeout=120,
     )
 
+    tb_src = f'{git_install_path}/TransferBench'
+    git_tag = config_dict['git_tag']
+    out_dict = orch.exec(
+        f"bash -c 'cd {tb_src} && git checkout {git_tag}'",
+        timeout=120,
+    )
```
solaiys, line 227: **Blocking (correctness): the `git checkout` result isn't verified, so the version pin can silently fail open.**

- The `out_dict` returned by this `git checkout` is discarded — there's no per-node error check.
- The only post-install verification is `ls -l {git_install_path}/TransferBench` (line 239), which confirms the **clone** directory exists, not that the **tag** was actually checked out.
- So if `git checkout {git_tag}` fails (tag missing, dirty/detached tree, fetch issue), the build below proceeds on the **default branch**, the test still passes, and you get a false-green that looks like `v1.67.00` but isn't.
- Pinning to `v1.67.00` is the whole purpose of this PR, so an unverified checkout makes the pin a no-op exactly when it matters.

Suggested fix — scan the checkout output the same way the install step does:

```suggestion
    out_dict = orch.exec(
        f"bash -c 'cd {tb_src} && git checkout {git_tag}'",
        timeout=120,
    )
    # Fail loudly if the tag checkout did not succeed; otherwise the build below
    # silently falls back to the default branch and the version pin is a no-op.
    for node in out_dict.keys():
        if re.search(r'error:|fatal:|did not match any', out_dict[node] or '', re.I):
            fail_test(f'TransferBench checkout of tag {git_tag} failed on node {node}: {out_dict[node]}')
```

- Even stronger (optional): assert the resolved HEAD matches the tag — e.g. run `git describe --tags`/`git rev-parse HEAD` after checkout and compare against `{git_tag}` — so a wrong-but-non-erroring checkout is also caught.

cvs/tests/health/install/install_transferbench.py
```diff
@@ -213,6 +215,13 @@ def test_install_transferbench(orch, config_dict):
         timeout=120,
     )
 
+    tb_src = f'{git_install_path}/TransferBench'
+    git_tag = config_dict['git_tag']
```
solaiys: **Minor (robustness): `git_tag` is read with a bare subscript.**

- This is inconsistent with the nearby `config_dict.get('rocm_path', '')` and `config_dict.get(...)` usage.
- `git_tag` is now effectively required for every TransferBench health config; an older or copied config without it will crash here with an opaque `KeyError` instead of a clear, actionable message.
- Only `mi300_health_config.json` exists today (and it does define `git_tag`), so this isn't a current crash — it's future-proofing for other configs.

Suggested fix — fail cleanly with a clear message when it's absent:

```suggestion
    git_tag = config_dict.get('git_tag')
    if not git_tag:
        fail_test('TransferBench config is missing the required "git_tag" field')
        update_test_result()
        return
```

solaiys (commented):
Reviewed the TransferBench changes. The example-test removal is clean (no dangling references to `example_tests_path` / `example_results` / the removed functions), and the scaling-test env-var change looks fine.

One blocking correctness issue on the new version-pin checkout, plus a minor robustness nit on the `git_tag` lookup — both inline.

solaiys (approved):
Both comments are addressed.
LGTM.

PR #241, opened 2026-06-25
[CVS] GPU metrics polling integration for inference validation suites

atnair-amd:
## Jira

[AIMVT-245](https://amd-hub.atlassian.net/browse/AIMVT-245) — GPU metrics polling integration for inference validation suites (DTNI epic AIMVT-202)

---

## Background

CVS validation suites had no visibility into GPU-level resource utilisation during benchmark runs — no record of peak VRAM consumption, memory delta from model load, GPU compute activity, or memory bandwidth utilisation.

This PR adds a standalone GPU metrics polling library, `gpu.py`. It is suite-agnostic: it shells out to `amd-smi metric --json` via an `Orchestrator` and parses/aggregates the result, with no suite-specific logic and no import-time side-effects. Earlier iterations of this branch wired the library into `vllm_single` and `sglang_disagg` as reference integrations; that wiring has since been stripped back out (see Scope note below) so this PR now ships the library on its own, ready for any suite — inference or training — to adopt.

---

### New files

| File | Description |
|---|---|
| `cvs/lib/utils/gpu.py` | GPU metrics library: `capture_gpu_metrics`, `poll_gpu_metrics`, `agg_readings`, `GPU_METRICS`, `GPU_METRIC_UNITS`. Pure parsing/aggregation functions plus a small imperative shell (`amd-smi` exec, polling loop, log writing). Supports both single-node (`orch.exec_on_head`) and multi-node (`orch.exec(cmd, hosts=...)`) polling via an optional `nodes` param. |
| `cvs/lib/utils/unittests/test_gpu.py` | 73 implementation-blind unit tests covering zero-value guards, N/A degradation, partial entry exclusion, multi-GPU aggregation, multi-node/multi-host pooling, and failure/recovery cycling. |
| `cvs/lib/utils/docs/gpu-metrics.md` | Integration guide: 5-step walkthrough, sync/threaded wiring patterns, single-node vs multi-node usage, `gpu_poll.log` format, threshold JSON schema, failure/None handling table, and gotchas. |

### Modified files

| File | Change |
|---|---|
| `cvs/lib/utils/AGENTS.md` | Added `gpu.py` section: public API table, single-node/multi-node contract, `poll_gpu_metrics` parameter table, 5-metric derivation table, and gotchas. |

---

## Library placement

`gpu.py` lives in `cvs/lib/utils/` (suite-agnostic), not `cvs/lib/inference/utils/`, per the repo's own boundary rule: if every CVS suite could need it, it belongs in `cvs/lib/utils/`. The library contains zero inference-specific logic — training suites that need GPU metrics can import it the same way.

## The 5 derived metrics

| Key | Unit | Aggregation |
|---|---|---|
| `gpu.peak_gpu_memory_mb` | MB | max over polls, each poll summed across GPUs/nodes |
| `gpu.model_load_memory_mb` | MB | post-load minus pre-load VRAM snapshot (caller-measured) |
| `gpu.model_load_s` | s | wall-clock elapsed while server starts (caller-measured) |
| `gpu.gpu_bandwidth_util_pct` | % | mean UMC activity over polls, averaged across GPUs/nodes |
| `gpu.gpu_compute_util_pct` | % | mean GFX activity over polls, averaged across GPUs/nodes |

---

## Validation

- 73 unit tests: `python -m unittest cvs.lib.utils.unittests.test_gpu -v` — all pass
- Standalone smoke-tested against a live conductor node (`banff-cyxtera-s70-2.ctr.dcgpu`), confirming `capture_gpu_metrics`/`poll_gpu_metrics` work end-to-end against real `amd-smi` output without any CVS orchestrator/config scaffolding
- Earlier end-to-end run (prior iteration, with `vllm_single` wiring) on core42 node `10.245.135.11` (g21u31): 5 GPU metric rows visible in HTML report, `gpu_poll.log` written to run directory. Run artefacts: `AIMVT-245` attachment `vllm_single_2026-06-25T124216.zip`

## Scope

This PR ships the `gpu.py` library on its own. Suite integrations (`vllm_single`, `sglang_disagg`, etc.) are follow-on work in separate PRs, keeping this change reviewable as "one library, no suite coupling."

## Out of scope

- Suite-level wiring (vllm_single, sglang_disagg, InferenceX, or other suite integrations — follow-on PRs)
- Per-GPU metric breakdown (cluster-level aggregates only)
- Energy / power metrics surfaced as HTML rows (collected in raw snapshots, not yet exposed as derived metrics)


solaiys:
### Review: ✅ clean — no blocking issues

Reviewed `cvs/lib/utils/gpu.py` and the surrounding files. This is a well-built, additive change and I found **no blocking issues**:

- **Zero regression risk** — the library isn't imported by any suite yet (only its own tests reference it) and the PR is purely additive (+1979/-0, no existing code touched), so it can't break existing functionality.
- **Defensive** — safe nested-dict navigation (`_safe_get`, `N/A` → `None`), zero-value guards preserved (0 vs `None` distinguished), division guarded (`_mean`/`max` both handle empty), `amd-smi`/parse wrapped in try/except with a consecutive-failure cap, log-write wrapped in try/except.
- **No injection / secrets / concurrency issues** — static `amd-smi metric --json` command (no interpolation), single-threaded polling.
- **Real test coverage** — 73 unit tests, no `xfail`/`skip`/`TODO` markers.

---

Two **non-blocking** observations (not merge blockers — noting for future hardening, especially before a suite wires this in):

1. **Multi-node bypasses the failure cap.** `_capture_multi_node` catches exceptions *per node* and returns an all-`None` snapshot, so `consecutive_failures` never increments in `nodes`-mode. If every node's `amd-smi` fails persistently, `poll_gpu_metrics` records all-`None` readings indefinitely (until `is_done_fn()` returns True) instead of tripping `max_consecutive_failures` like single-node mode does. Consider counting an "all nodes failed" round toward the cap.

2. **`poll_gpu_metrics` docstring says "Never raises,"** but `is_done_fn()` is intentionally called outside the amd-smi try/except (line 318), so a raising `is_done_fn` would propagate and skip the summary/log write. The behavior looks deliberate (per the inline comment) — just a docstring wording mismatch worth tightening.

solaiys (approved):
LGTM. Check my non-blocking comments..

hnimra-amd:
LGTM

hnimra-amd (approved):

PR #244, opened 2026-06-29
Hnimrama/ix atom experimental run deck

hnimra-amd:
## Motivation

InferenceX ATOM lab runs produce rich sweep data (`inf_res_dict`, lifecycle timings, gate tiers) but results today live mainly in pytest logs and a flat results table. This PR adds a **shared, render-only report library** (`cvs.lib.report`) and wires **`inferencex_atom_single`** as the reference implementation so that `cvs run … --html` automatically produces an HTML/JSON dashboard, interactive viewer, and CI summary, bundled into the existing results zip, without changing pass/fail or threshold enforcement.

The branch also generalizes the pattern so **other inference suites** can opt in with a single preset file (`presets/<cvs_run_stem>.py`), with root pytest wiring handled centrally.

**Base branch:** `hnimrama/IX-atom`

### New package: `cvs/lib/report/`

Generic engines for inference and training suite reports:

- **Inference:** run card, lifecycle timeline, sweep analytics, concurrency charts, gate matrix + heatmap, per-cell cards, full results table, JSON sidecar (`schema_version: 1`)
- **Interactive viewer:** `{basename}_viewer.html` with filters, charts, CSV export; JSON embedded for `file://` / zip workflows
- **CI summary:** one-page `{basename}_summary.html` for quick status review
- **Parity:** cross-framework compare when sibling JSON exists (`inference_parity_report.*`); prev-run panel via `{basename}_prev.json` or env
- **Training pilot:** Megatron 8B single wired as a second consumer

Refactored for maintainability: shared `metrics`, `compare`, `json_io`, `artifacts`, `panel_shell`, payload/HTML split for inference reports.

### IX-atom integration

- Preset: `cvs/lib/report/presets/inferencex_atom.py` (W1 parity metrics, cell cards, viewer)
- Auto-load shim: `presets/inferencex_atom_single.py` (matches `cvs run inferencex_atom_single` stem)
- Suite collects data only in tests; **no report hooks in suite conftest**
- Root `cvs/conftest.py`: auto-registers preset, binds session fixtures, attaches lifecycle/cell-card HTML extras, triggers `generate_suite_reports()` at session end

### Plug-and-play for other suites

- `make_inference_report_config()`: sensible defaults for title, basename, charts
- `_inference_suite_template.py`: copy to `presets/<stem>.py`
- `auto_register.py`: loads preset by `cvs run` module stem
- Docs: `cvs/lib/report/README.md`, `ADDING_A_SUITE.md` (IX-atom as reference)

### Session / zip plumbing

- `HtmlReportManager.generate_suite_reports()` writes artifacts beside pytest HTML and links them in the **Reports** section
- Reports copied into `{suite}_{timestamp}.zip` with pytest HTML and logs

### Bug fixes

- Parity publish: `UnboundLocalError` when sibling comparator JSON triggered parity without `htmlpath` set
- Interactive viewer: embed full JSON in HTML so viewer works when opened from zip / `file://` (no failed `fetch` of sibling JSON)

## Test Plan

- [x] `python -m pytest cvs/lib/report/unittests/ -q`
- [x] Lab smoke on MI300X:
  ```bash
  cvs run inferencex_atom_single \
    --cluster_file ~/input/cluster_file/mi300x_atom_single.json \
    --config_file <smoke_config.json> \
    --html=~/cvs_results/smoke.html --self-contained-html -vvv -s
  ```
- [x] Confirm beside pytest HTML: `inferencex_atom_report.html`, `.json`, `_viewer.html`, `_summary.html`
- [x] Confirm zip contains reports and pytest HTML **Reports** links work after extract
- [x] Open `_viewer.html` from extracted zip (embedded JSON loads)
- [x] Optional: place `inferencex_atom_report_prev.json` or comparator JSON beside HTML; verify prev-run / parity artifacts

## Test Result

- Report library unit tests pass locally (`cvs/lib/report/unittests/`)
- IX-atom lab validation pending on target hardware (MI300X / MI355X); see `cvs/input/config_file/inference/inferencex_atom_single/README.md` for smoke and W1 perf commands


cvs/conftest.py
```diff
@@ -24,7 +24,60 @@ def pytest_configure(config):
...
+    return try_auto_register_inference_suite_report(config)
+
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_configure(config):
+    _ensure_html_report_manager(config)
+
+
+@pytest.hookimpl(trylast=True)
+def pytest_configure(config):
```
atnair-amd, line 51: **P0: duplicate `pytest_configure` — `tryfirst` body is dead code.**

Python resolves two `def` statements with the same name in the same module by keeping the last one. The `@pytest.hookimpl(tryfirst=True)` definition at line 47 is silently discarded — `_ensure_html_report_manager` never runs via this hook.

Merge both bodies into a single `pytest_configure`:

```python
@pytest.hookimpl(tryfirst=True)
def pytest_configure(config):
    _ensure_html_report_manager(config)
    _auto_register_inference_suite_report(config)
```

Remove the `trylast` definition entirely. `tryfirst` is correct here — the manager must exist before other plugins reference it.

cvs/conftest.py
```diff
@@ -24,7 +24,60 @@ def pytest_configure(config):
...
+    _auto_register_inference_suite_report(config)
+
+
+@pytest.fixture(scope="module", autouse=True)
```
atnair-amd: **P1: fixture scope should be `"session"`, not `"module"`.**

`scope="module"` means pytest creates a new fixture instance per test module. In a multi-module run the session store gets rebound on each module boundary, so the final report contains only the last module's data.

```python
@pytest.fixture(scope="session", autouse=True)
```

cvs/conftest.py
```diff
@@ -93,15 +146,26 @@ def pytest_metadata(metadata):
...
     outcome = yield
     report = outcome.get_result()
-    report.extras = item.config._html_report_manager.write_test_log(report, item.originalname)
+    report.extras = _ensure_html_report_manager(item.config).write_test_log(report, item.originalname)
+
+    from cvs.lib.report.registry import get_suite_report_config
+    from cvs.lib.report.types import InferenceReportConfig
+
+    if isinstance(get_suite_report_config(item.config), InferenceReportConfig):
+        from cvs.lib.inference.inference_suite_lifecycle import attach_lifecycle_html_table
```
atnair-amd: **P1: layering violation — import `attach_lifecycle_html_table` through `inference_wiring.py`, not directly from `cvs.lib.inference`.**

`inference_wiring.py` exists as the stable public boundary for this kind of hook wiring but doesn't expose this function yet. Add it there:

```python
# in inference_wiring.py
def attach_inference_suite_lifecycle_extra(item, report) -> None:
    from cvs.lib.inference.inference_suite_lifecycle import attach_lifecycle_html_table
    attach_lifecycle_html_table(item, report)
```

Then replace this direct import with a call to `attach_inference_suite_lifecycle_extra`.

cvs/lib/report_plugins.py
```diff
@@ -356,17 +356,51 @@ def generate_reports_section(self):
         if not self._custom_test_reports:
             return ""
 
-        # Simple Reports section - only test reports, no config files
         html = '<div><h2>Reports</h2><ul>'
-
-        # Add test reports only
         for report in self._custom_test_reports:
             html += f'<li><a href="{report["path"]}" target="_blank">{report["name"]}</a></li>'
```
atnair-amd: **P0: XSS — `report["path"]` and `report["name"]` interpolated raw into HTML anchor.**

A report name or path containing `"` or `<` breaks the anchor or injects markup.

```python
html += (
    f'<li><a href="{html.escape(report["path"], quote=True)}"'
    f' target="_blank">{html.escape(report["name"])}</a></li>'
)
```

Also fix lines 95 and 100 in the same file (pre-existing, not in this diff):

```python
# line 95
log_content.append(
    f"<h3>{html.escape(section_name)}</h3><pre>{html.escape(section_content)}</pre>"
)
# line 100
f"<html><body><h1>{html.escape(report.nodeid)}</h1>{''.join(log_content)}</body></html>",
```

`import html` at the top of this file — stdlib, no new dependency.

cvs/lib/report/inference_payload.py
```diff
@@ -0,0 +1,287 @@
...
+            }
+        )
+    return summaries
+
+
+def build_results_table(config: InferenceReportConfig, inf_res_dict: Mapping[tuple, Any]) -> dict:
+    headers = [label for label, _key in config.results_columns]
+    metric_keys = [key for _label, key in config.results_columns]
+    rows: List[List[Any]] = []
+    for key, host_dict in sorted(inf_res_dict.items(), key=lambda kv: (kv[0][4], kv[0][5])):
```
atnair-amd: **P2: `IndexError` on short-tuple keys.**

`kv[0][4]` and `kv[0][5]` raise `IndexError` if a key tuple has fewer than 6 elements. The unpacking on the next line would also fail, but the sort runs first.

```python
key=lambda kv: (kv[0][4] if len(kv[0]) > 4 else "", kv[0][5] if len(kv[0]) > 5 else "")
```

cvs/lib/report/inference_payload.py
```diff
@@ -0,0 +1,287 @@
...
+        model, gpu, isl, osl, policy, conc = key
+        if not isinstance(host_dict, dict):
+            continue
+        fixed = [model, gpu, isl, osl, policy, conc]
+        for host, metrics in host_dict.items():
+            row = list(fixed)
+            row.append(host)
+            for mk in metric_keys[7:]:
```
atnair-amd: **P1: hardcoded column offset `[7:]` is fragile.**

`metric_keys[7:]` assumes exactly 7 fixed columns before the metric columns begin. Any preset with a different fixed-column count silently drops or misidentifies metric columns.

Compute the offset from the column definition itself — fixed columns are already identified by `key is None`:

```python
n_fixed = sum(1 for _label, key in config.results_columns if key is None)
for mk in metric_keys[n_fixed:]:
    ...
```

cvs/lib/report/ci_summary.py
```diff
@@ -0,0 +1,189 @@
...
+    cells = list(payload.get("cells") or [])
+    gated = config.gated_tiers
+    headline_metric = config.headline_metric
+    scored = sorted(
+        ((_cell_worst_score(c, gated, headline_metric), c) for c in cells),
+        key=lambda item: item[0],
+    )
+    worst: List[dict] = []
+    for score, cell in scored:
+        if score[0] > 1 and len(worst) >= limit:
```
atnair-amd: **P2: unreachable guard — dead code.**

`score[0] > 1 and len(worst) >= limit` is unreachable. `scored` is sorted ascending, so once `len(worst) >= limit` the `if len(worst) >= limit: break` at line 53 always fires first.

Remove lines 49-51 entirely:

```python
for score, cell in scored:
    worst.append(cell)
    if len(worst) >= limit:
        break
```

atnair-amd (changes requested):
### P0 — Must fix before merge

**Duplicate `pytest_configure` (`cvs/conftest.py:47-53`)** — see inline.

**XSS in per-test HTML log (`cvs/lib/report_plugins.py:95, 100, 361`)**

Lines 95 and 100 are pre-existing but in the file this PR expands — fix them here:

```python
# line 95
log_content.append(
    f"<h3>{html.escape(section_name)}</h3><pre>{html.escape(section_content)}</pre>"
)

# line 100
f"<html><body><h1>{html.escape(report.nodeid)}</h1>{''.join(log_content)}</body></html>",
```

Line 361 is in the diff — see inline. `import html` is stdlib, no new dependency.

---

### P1 — Required before merge

**Fixture scope (`cvs/conftest.py:57`)** — see inline.

**Direct `cvs.lib.inference` import in root conftest (`cvs/conftest.py:164`)** — see inline.

**Hardcoded `metric_keys[7:]` (`cvs/lib/report/inference_payload.py:147`)** — see inline.

**Test gaps:**

- `lifecycle_report={}` in every test case — `aggregate_lifecycle` never executes. Add at least one fixture with populated lifecycle data and assert on `cell_lifecycle`.

- `enforce_thresholds=True` with actuals below threshold never flows through `build_cell_record`. The fixture at `_fixtures.py:12` sets the flag but no test passes failing actuals through the normal build path. Add a variant; assert `tier_status == "fail"`.

- `bar_pct()` has 5 branches (`min`/`max`/`within`/default/`target<=0`) and zero direct tests. Add unit tests per branch.

---

### P2 — Should fix

- `inference_payload.py:139` — sort key `lambda kv: (kv[0][4], kv[0][5])` raises `IndexError` if a key tuple has fewer than 6 elements — see inline.

- `inference_payload.py:86-88` — `saturated` is True only when the last concurrency point is also the best. A curve that peaks mid-sweep (degradation) shows as not saturated, which is correct, but saturation and peak-at-last are conflated into one flag. Consider a separate `curve_shape` field or a comment documenting the intended semantics.

- `ci_summary.py:49` — dead unreachable guard — see inline.

- `conftest.py` fixture block (~line 74) — single `try` over three `getfixturevalue` calls means a missing `lifecycle` fixture silently drops all three bindings. Split into three separate `try/except` blocks with a warning log on each.

cvs/lib/report/formatting.py
```diff
@@ -0,0 +1,81 @@
...
+
+def link_or_text_html(url: str, label: str) -> str:
+    if not url:
+        return "\u2014"
+    safe_label = html.escape(label)
+    if re.match(r"^https?://", url):
+        safe_url = html.escape(url)
+        return f'<a href="{safe_url}" target="_blank" rel="noopener">{safe_label}</a>'
+    normalized = url.replace("\\", "/")
+    if normalized.startswith("../") or "/" in PurePosixPath(normalized).parts:
```
atnair-amd: **P1: this branch matches absolute paths, not relative bundle-links — the opposite of the intent.**

`PurePosixPath(...).parts` only contains `"/"` as an element for an **absolute** path (it's the root anchor); a relative path like `"../ix_atom_run.html"` has `parts == ('..', 'ix_atom_run.html')` with no `"/"`. So `"/" in PurePosixPath(normalized).parts` is true exactly for absolute filesystem paths, not for the `"../foo.html"` bundle-relative links this was meant to preserve. The `startswith("../")` check already handles the relative case correctly on its own.

Net effect: an absolute path (e.g. `/home/user/cvs_results/run.html`) now renders as `href="/home/user/cvs_results/run.html"` instead of being reduced to `href="run.html"` as before this change — leaking a local filesystem path into the HTML report. `presets/builder.py:25` still has a fallback (`pytest_html_href or pytest_html_path`) that assumes this function sanitizes an absolute path down to a basename if needed; that safety net is broken by this.

Suggested fix — drop the redundant/incorrect clause:
```python
if normalized.startswith("../"):
    return f'<a href="{html.escape(normalized)}">{safe_label}</a>'
```

This also fixes the two currently-failing tests: `test_link_or_text_html_local_path_uses_basename` and `test_render_report_html_from_payload`.

atnair-amd (commented):
Verified all 7 items from the previous review round are fixed correctly in `065cb5c` — duplicate `pytest_configure` merged, XSS escaped everywhere (including the two spots outside the original diff), layering routed through `inference_wiring.py`, sort-key/slice hardening, and the dead-code guard removed. The fixture-scope fix took a different (and valid) approach: instead of switching to `scope="session"`, the bind now happens in a `request.addfinalizer` at module teardown so it fires after the module's tests populate the fixtures, rather than at setup. That addresses the original data-loss concern.

While re-running the suite on `0c2b973` + `065cb5c`, I found one new regression introduced by `0c2b973` (unrelated to the review items above — a separate bundle-relative-links fix):

**P1: `link_or_text_html` mis-detects absolute paths as bundle-relative links.**

See inline comment on `cvs/lib/report/formatting.py`. Two tests currently fail because of this: `test_formatting.py::test_link_or_text_html_local_path_uses_basename` and `test_inference_report.py::test_render_report_html_from_payload`.

(A third failure, `test_provenance.py::test_payload_includes_provenance_and_run_card_fields` asserting `"CVS version" in labels`, is pre-existing at `2243943` — unrelated to either of these two new commits, calling it out only so it isn't conflated with the regression above.)

amd-droy (approved):
looks great. Thanks.

PR #247, opened 2026-07-06
Hnimrama/ix atom multinode

hnimra-amd:
Add multinode ATOM inference support for InferenceX atom suites, including distributed serve, scaling-efficiency gates, and 2-node W1 perf sweep variants for MI300X and MI355X.

## Motivation

Single-node `inferencex_atom` coverage exists on `dev/dtni`, but lab validation for multi-GPU / multi-node DeepSeek R1 runs requires a DTNI path that can launch ATOM across cluster ranks, assert scaling efficiency, and run a realistic multinode sweep matrix.

This PR adds that multinode layer: distributed ATOM serve, multinode config/schema fields, scaling metrics in gates and reports, 2-node cluster examples for **MI300X and MI355X**, and matching `perf_multi` variants with expanded sweep thresholds.

**Merge order:** Intended to land **after** `hnimrama/refactor-ix-atom` (moves shared suite helpers under `inference/utils/`). Expect a short rebase/conflict-resolution pass once that refactor merges.

### Suite rename

- `inferencex_atom_single` → **`inferencex_atom`** (test module, config directory, docs, CLI references)
- Config moved: `cvs/input/config_file/inference/inferencex_atom_single/` → `inferencex_atom/`
- Report preset shim `inferencex_atom_single.py` removed (auto-register uses `inferencex_atom` stem)

### Multinode orchestration (`InferenceXAtomJob`)

- Distributed **ATOM-native** multinode serve via SPMD env vars (`ATOM_DP_*`, `-dp`), not vLLM `--node-rank` / PP flags
- `_atom_multinode_argv()` isolates ATOM CLI tokens; vLLM distributed flags stripped from `atom_args` when present
- `parse_results` accepts optional `scaling_baseline_output_throughput` + `nnodes` for efficiency derivation

### Config & metrics

- New params: `nnodes`, `node_rank`, `master_addr`, `master_port`, `pipeline_parallel_size`, `scaling_baseline_output_throughput`
- New metric tier **`scaling`** with `scaling.efficiency_pct` (= actual / (single-node baseline × nnodes))
- Loader validates multinode variants define `scaling.efficiency_pct` thresholds when baseline is set
- Cell keys include `PP=` / `NN=` suffixes when multinode dimensions apply

### Cluster & variants

| GPU | Cluster file | Variant |
|-----|--------------|---------|
| MI300X | `cvs/input/cluster_file/mi300x_atom_multi.json` | `mi300x_inferencex-atom-single_deepseek-r1_fp8_perf_multi` |
| MI355X | `cvs/input/cluster_file/mi355x_atom_multi.json` | `mi355x_inferencex-atom-single_deepseek-r1_fp8_perf_multi` |

Both variants use the same sweep matrix: 5 sequence shapes × concurrencies 16/64/128 (15 cells), `nnodes=2`, `PP=2`.

**MI300X:** `rocm/atom-dev:latest`, `scaling_baseline_output_throughput=1500`, `enforce_thresholds: true`.

**MI355X:** `rocm/atom-dev:nightly_202606211542`, `scaling_baseline_output_throughput=4000` (seeded from single-node CI), `enforce_thresholds: false` until 2-node lab confirms.

### Infrastructure

- **`container.py`**: install `openssh-server` when image lacks `sshd` (multinode SSH setup)
- Lifecycle/report fixes: `CollectReport` tolerance in HTML hooks, `test_launch_container` syntax repair

### Key commits

1. Multinode params + scaling efficiency metric
2. Distributed ATOM serve across cluster ranks
3. MI300X 2-node cluster + `perf_multi` variant
4. Unit tests + docs for multinode routing/config
5. Orchestrator openssh install
6. ATOM-native multinode serve fixes + argv isolation
7. Expanded multinode sweep matrix + scaling threshold enforcement
8. Lifecycle/report fixes for multinode runs
9. Suite rename `inferencex_atom_single` → `inferencex_atom`
10. MI355X multinode cluster + `perf_multi` variant + config-loader test

## Test Plan

- [x] Unit tests: multinode job routing, config loading, and orch parse (`test_inferencex_atom_config_loader`, `test_inferencex_atom_orch_parse`, `test_inferencex_atom_parsing`)
- [x] Unit test: `test_load_w1_mi355x_multinode_variant`
- [x] Lab smoke + perf: `cvs run inferencex_atom` with multinode cluster + `perf_multi` (2-node MI325X, W1)

## Test Result

| Check | Result |
|-------|--------|
| Multinode unit tests (`test_inferencex_atom_config_loader`, `test_inferencex_atom_orch_parse`) | Added |
| `test_load_w1_mi355x_multinode_variant` | Added |
| Multinode lab run (2-node MI325X) | **Pass** |


cvs/lib/inference/atom/atom_parsing.py
amd-droy, line 1: For driver=atom with the configs (TP=8, nnodes=2), seems like the setup is not a coordinated 2-node distributed model. It is two independent full replicas, and the benchmark only uses the head. _atom_spmd_dp_cli() returns [] because tp * nnodes > 8, so neither -dp nor ATOM_DP_* env vars are applied. _atom_server_argv(rank) also ignores rank on the ATOM path.

Both nodes therefore start the same -tp 8 command with no rank-specific coordination, while configs/thresholds label cells as PP=2,NNODES=2. 

After containers are up, the inference test runs the job:
test_inferencex_atom_inference() → InferenceXAtomJob.start_server() 
For driver=atom, the serve command is built from:
roles.server.atom_args in config → ["-tp", "8", "--kv_cache_dtype", "fp8", ...]
_atom_server_argv() , that becomes roughly:
python -m atom.entrypoints.openai_server --model ... --server-port 8000 -tp 8 --kv_cache_dtype fp8 ...

cvs/core/orchestrators/container.py
```diff
@@ -477,11 +477,16 @@ def setup_sshd(self):
             "chown -R root:root /root/.ssh",
             "bash -c 'chmod 700 /root/.ssh && chmod 600 /root/.ssh/*'",
             "mkdir -p /run/sshd",  # Create privilege separation directory for sshd
+            "bash -c 'which sshd > /dev/null 2>&1 || (apt-get update -qq && apt-get install -y -q openssh-server iproute2)'",
             "/usr/sbin/sshd -p2224",
         ]
 
+        timeouts = {
+            "bash -c 'which sshd > /dev/null 2>&1 || (apt-get update -qq && apt-get install -y -q openssh-server iproute2)'": 120,
+        }
```
amd-droy: Installing openssh-server at test time via apt-get update && apt-get install adds network dependency, latency (120s timeout), and failure modes on images.

Rather fail fast with a clear message (“image must include sshd”) or document openssh-server as a required base image property.

cvs/lib/inference/atom/atom_orch.py
```diff
@@ -262,27 +483,49 @@ def _atom_warmup_ok(self):
...
         return bool(out) and all(r["exit_code"] == 0 for r in out.values())
 
+    def _tail_server_logs(self, lines=30):
+        if self.distributed:
+            out = {}
+            for rank in range(self.nnodes):
+                chunk = self._exec_all(f"tail -{lines} {shlex.quote(self._rank_server_log(rank))}")
+                out.update(chunk or {})
+            return out
+        return self._exec_all(f"tail -{lines} {shlex.quote(self.server_log)}")
```
atnair-amd, line 600: **`_tail_server_logs` returns wrong log content per host in multinode runs**

```python
def _tail_server_logs(self, lines=30):
    if self.distributed:
        out = {}
        for rank in range(self.nnodes):
            chunk = self._exec_all(f"tail -{lines} {shlex.quote(self._rank_server_log(rank))}")
            out.update(chunk or {})
        return out
```
`_exec_all` runs the command against **all** hosts (no `hosts=` filter), unlike `start_server`'s equivalent per-rank call which does pass `hosts=[host]`. Each loop iteration tails *rank N's* log path on *every* host, and `out.update(...)` means the final iteration's paths win for every host key. With 2 nodes: after the loop, both host entries end up holding the result of tailing rank1's log path, evaluated on each host -- not each host's own server log.

This is called from `wait_ready()` for early-failure detection, so a crash on node0 can be masked by whatever rank1's log path resolves to on node0 -- multinode early-failure detection is effectively broken.

**Proposed fix:** filter by host, matching the pattern already used in `start_server`:
```python
def _tail_server_logs(self, lines=30):
    if self.distributed:
        hosts = list(getattr(self.orch, "hosts", []) or [])
        out = {}
        for rank, host in enumerate(hosts):
            chunk = self._exec_all(
                f"tail -{lines} {shlex.quote(self._rank_server_log(rank))}", hosts=[host]
            )
            out.update(chunk or {})
        return out
    return self._exec_all(f"tail -{lines} {shlex.quote(self.server_log)}")
```

atnair-amd (commented):
**`_tail_server_logs` returns wrong log content per host in multinode runs** — see inline comment below.

cvs/lib/inference/unittests/test_inferencex_atom_config_loader.py
```diff
@@ -29,23 +29,53 @@ class TestInferenceXAtomConfigLoader(unittest.TestCase):
...
+            "input/config_file/inference/inferencex_atom/mi300x_inferencex-atom_gpt-oss-120b_bf16_config.json"
         )
         variant = load_variant(config, _cluster_dict())
-        self.assertEqual(variant.framework, "inferencex_atom_single")
+        self.assertEqual(variant.framework, "inferencex_atom")
         self.assertEqual(variant.params.driver, "vllm")
         self.assertEqual(variant.expected_cells(), ["ISL=7168,OSL=1024,TP=8,CONC=64"])
         self.assertIn("enforce-eager", variant.roles.server.serve_args)
 
+    def test_deprecated_framework_alias_normalizes(self):
```
atnair-amd: **`test_deprecated_framework_alias_normalizes` fails on current HEAD**

This test builds a `driver="atom"` variant without `roles.server.atom_args`, but a validator (`_atom_driver_requires_inline_server_args`) now raises `ValueError` whenever `driver=="atom"` and `atom_args` is empty. The test wasn't updated alongside the new validator, so `python -m unittest cvs.lib.inference.unittests.test_inferencex_atom_config_loader` currently fails at this HEAD.

**Proposed fix:** add `roles={"server": {"atom_args": ["-tp", "8"]}}` to this test's `InferenceXAtomVariantConfig(...)` call (or switch the fixture to `driver="vllm"` if inline args aren't the point of the test).

atnair-amd (commented):
**`test_deprecated_framework_alias_normalizes` fails on current HEAD** — see inline comment below.

atnair-amd:
Two blockers remain before merge:
1. `python -m unittest discover -s cvs/lib/inference/unittests -p "test_inferencex_atom*.py"` currently fails with 4 errors (see inline comments plus the note below) — none are exotic, all are stale test fixtures not updated alongside recent validator/key changes.
2. `_tail_server_logs` has a real host-clobbering bug that breaks early-failure detection on multinode (see inline comment) — this is a correctness issue in shipped code, not just tests.

Recommend fixing both before merge; happy to send a small patch for all 4 if useful.

---

**`server_session_key` tests use stale `SimpleNamespace` fixtures missing `nnodes`**

`server_session_key()` reads `p.nnodes`, `p.pipeline_parallel_size`, `p.master_addr`, `p.master_port` (added for multinode reuse-key disambiguation). Three existing tests build `params=SimpleNamespace(driver="atom", tensor_parallelism="8")` without those fields, so they now fail with `AttributeError: 'SimpleNamespace' object has no attribute 'nnodes'` at this HEAD:
- `test_reuse_server_flag_and_session_key_helpers` in `cvs/lib/inference/unittests/test_inferencex_atom_config_loader.py:350` (predates this PR, unchanged — noting here since GitHub won't let me anchor a comment on an untouched line)
- `test_server_session_key_differs_for_model` and `test_server_session_key_differs_for_shape` in `cvs/lib/inference/unittests/test_inferencex_atom_server_reuse.py` (also not part of this PR's diff)

This is purely a test-fixture staleness issue, not a production bug — the real `InferenceXAtomParams.nnodes` field has a pydantic default (`"1"`), so actual callers never hit this.

**Proposed fix:** add the missing fields to each fixture's `params=SimpleNamespace(...)`, e.g. `SimpleNamespace(driver="atom", tensor_parallelism="8", nnodes="1", pipeline_parallel_size="1", master_addr="", master_port="29501")`.

PR #250, opened 2026-07-09
Integrating nodesmoke tier 1 tests using Primus cli

urtiwari:
## Motivation

Integrate Primus node_smoke Tier 1 into CVS preflight checks so clusters can run per-node GPU, host, and RDMA health screening through the existing cvs run preflight_checks workflow 
Goals:

Add an opt-in test_node_smoke preflight step that launches primus-cli direct --single -- node_smoke on each reachable node via parallel SSH
Automate Primus clone/venv setup on cluster nodes (with NFS-safe shared install)
Surface node_smoke results in the preflight HTML report and summary alongside existing checks (reachability, ROCm, interfaces, GID, IFoE, RDMA)

## Technical Details

cvs/tests/preflight/preflight_checks.py — adds test_node_smoke wired into the preflight pytest suite; opt-in via node_smoke.connectivity_mode ("run" / "skip", default "skip"); failing nodes are reported but not pruned from phdl
cvs/lib/preflight/report.py — adds executive summary, recommendations, and HTML section for node_smoke failures (per-node fail reasons, dump path)
cvs/parsers/schemas.py — adds PreflightNodeSmokeConfig with validation for all node_smoke settings
cvs/input/config_file/preflight/preflight_config.json — adds full node_smoke config block with documented defaults

## Test Plan

Unit tests: python3 -m unittest cvs.lib.preflight.unittests.test_node_smoke cvs.lib.preflight.unittests.test_primus_setup (23 tests, all pass)

 Config validation: preflight_config.json parses and schema accepts PreflightNodeSmokeConfig

 Manual cluster run on 2-node test cluster

Verified auto-setup (clone, branch sync, shared NFS install, torch venv) on shared home

 Verified node_smoke PASS on both nodes after setting node_smoke.connectivity_mode: "run" and cluster-specific gid_index, expected_rocm_version, rdma_interfaces

 Verified preflight HTML report includes node_smoke summary section

 Full preflight suite run with default config (connectivity_mode: "skip") — node_smoke correctly reported as SKIPPED

## Test Result

Unit tests (test_node_smoke, test_primus_setup) | PASS — 23/23
test_node_smoke on 2-node cluster (with connectivity_mode: "run") | PASS — 2/2 nodes
Primus auto_setup (shared NFS install) | PASS
Preflight HTML report (node_smoke section) | PASS
Default config (connectivity_mode: "skip") | SKIPPED (expected)

Test reports are uploaded,

http://10.7.54.167/cvs/preflight_checks_html/

### DIFF
--- a/cvs/input/config_file/preflight/README_preflight_config.md
+++ b/cvs/input/config_file/preflight/README_preflight_config.md
@@ -12,6 +12,7 @@ The preflight checks system validates essential cluster health before running pe
 4. **Interface Name Consistency** - Validates RDMA interface naming patterns
 5. **IFoE L2 Connectivity (AIMVT-180; opt-in)** - Runs `afmctl test ping`
    on each node and enforces per-port and Summary pass/fail accounting
+6. **Primus Node Smoke (opt-in)** - Per-node host / GPU / RDMA roll-call via `primus-cli direct -- node_smoke`
 
 ## Configuration File Structure
 
@@ -37,6 +38,13 @@ The preflight configuration file follows this structure:
         "inter_full_mesh_group_pairs_per_wave": "auto"
       }
     },
+    "node_smoke": {
+      "connectivity_mode": "skip",
+      "auto_setup": true,
+      "primus_dir": "/home/{user-id}/INSTALL/Primus",
+      "venv_activate": "/home/{user-id}/envs/preflight/.venv/bin/activate",
+      "gpus_per_node": 8
+    },
     "reporting": {
       "generate_html_report": "true",
       "artifacts_root_dir": "/tmp/{user-id}/preflight",
@@ -59,6 +67,7 @@ preflight/
 ├── connectivity_check/   # Inter-node connectivity tests
 │   ├── rdma/             # RDMA-specific parameters (including nodes_per_full_mesh_group)
 │   └── ifoe/             # IFoE L2 ping parameters (AIMVT-180; opt-in)
+├── node_smoke/           # Primus node_smoke per-node health screening (opt-in)
 └── reporting/           # Output and report generation
 ```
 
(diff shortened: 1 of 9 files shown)


solaiys:
**Non-blocking suggestion (not a merge blocker):**

`NodeSmokeCheck.run()` builds its `exec_cmd_list` command list only over the filtered `hosts` subset, whereas `PrimusSetup.run()` builds one command per `self.phdl.reachable_hosts` (padding non-target hosts with `"true"`). `exec_cmd_list` maps `cmd_list[i]` positionally to `reachable_hosts[i]` and requires equal length, so the two are only guaranteed to line up because the sole caller passes `node_list = list(phdl.reachable_hosts)` and `auto_setup` aborts early if a node is pruned mid-run.

To make `NodeSmokeCheck` robust against a future caller that passes a subset (or a different ordering) of `reachable_hosts`, consider mirroring the `PrimusSetup` pattern — iterate `self.phdl.reachable_hosts`, emit the real command for target hosts and `"true"` for the rest, then map results back by host. This keeps the command/host alignment correct regardless of how `node_list` relates to `reachable_hosts`.

Purely a robustness improvement; the current code is correct for the existing `test_node_smoke` caller.

solaiys:
LGTM. 
After rebase on top of the latest dev/dtni branch, it can be ready to be merged.


Previous non-blocking suggestion — ✅ addressed.

NodeSmokeCheck.run() now correctly mirrors the PrimusSetup pattern. The updated code iterates self.phdl.reachable_hosts, pads non-target hosts with "true", and uses a host_ranks dict to assign each host its correct node_rank:

```
commands: List[str] = []
for h in self.phdl.reachable_hosts:
    if h not in hosts_set:
        commands.append("true")
    else:
        commands.append(build_remote_node_smoke_command(..., node_rank=host_ranks[h], ...))
```
Results filtering (for host, output in out_dict.items(): if host not in hosts_set: continue) is also correctly in place. The command/host alignment is now robust regardless of caller.

PR #252, opened 2026-07-09
Adding tests for Kimi-K2.6 and GLM-5.2-FP8

amd-droy:
cvs/lib/inference/sglang/sglang_disagg_orchestrator.py - Added this file for orchestrating the runs.
Added the threshold files for llama-70b, deepseek-ri-0528, Kimi-2.6-w4A8, GLM-5.2-FP8 and gpt-odd-120b for the concurrency and ISL/OSL sweeps.
Added the long Context accuracy tests for ISL - 96K 128K 200K

Removed the Run deck reports viewer for now. Will integrate it once its fully updated.
Attaching the reports generated for reference.

[reports.zip](https://github.com/user-attachments/files/29891004/reports.zip)


cvs/lib/docker_lib.py
```diff
@@ -84,7 +97,7 @@ def old_install_docker_on_ubuntu(phdl):
 
 
 def install_docker_on_ubuntu(phdl):
-    phdl.exec('sudo rm /etc/apt/keyrings/docker.gpg')
+    phdl.exec('sudo rm /petc/apt/keyrings/docker.gpg')
```
solaiys: 🔴 **Blocking — typo breaks docker keyring removal.** This path is `/petc/apt/keyrings/docker.gpg`; the base branch has `/etc/apt/keyrings/docker.gpg`. `/petc/...` doesn't exist, so the stale keyring is never removed and the `rm` errors out. Looks like an accidental edit — please restore `/etc/apt/keyrings/docker.gpg`.

cvs/lib/inference/sglang/sglang_disagg_lib.py
```diff
@@ -1183,10 +1197,10 @@ def timed_out() -> bool:
             # (e.g., OOM, RDMA failures, backend crashes).
             # --------------------------------------------------------------
             # Early abort on inference errors
-            if not self.scan_for_inference_errors():
-                msg = 'Failures seen in inference logs, Aborting!!!'
-                fail_test(msg)
-                return {"status": "error", "reason": msg}
+            # if not self.scan_for_inference_errors():
```
solaiys, line 1235: 🔴 **Blocking — early-abort on inference errors disabled.** This was the only early-abort on inference-log failures (OOM, RDMA failures, backend crashes) during polling. With it commented out, a crashed/errored run no longer fails fast and can run out the poll window and be scored as PASS on stale/garbage output — a false-positive regression. If this is intentional (e.g. it was aborting on false positives), please remove it with justification or gate it behind a flag rather than leaving commented-out dead code.

cvs/input/config_file/inference/sglang/mi30x_sglang_kimi_k2.6_threshold.json
```diff
@@ -0,0 +1,47 @@
+{
+    "_comment": "Kimi-K2.6 thresholds for MI30X SGLang disaggregated. Performance from bench_serv_random expected_results.auto; accuracy from lm_eval_* expected_results in mi30x_sglang_distributed.json.",
+    "ISL=1024,OSL=1024,TP=16,CONC=64": {
```
solaiys: 🟠 Non-blocking heads-up: this cell is keyed `TP=16`, but `mi30x_sglang_distributed.json` sets `tensor_parallelism: "8"` for kimi-k2.6. Gating still works (specs are injected per-cell), but `_thresholds_for_cell()` in `_shared.py` rebuilds the cell id with `TP=8`, so it won't match this `TP=16` key and the Expected/Result columns in the detailed perf table will render as `-`. Suggest aligning the key to `TP=8`.

cvs/input/config_file/inference/sglang/mi30x_sglang_glm_52_fp8_threshold.json
```diff
@@ -0,0 +1,76 @@
+{
+  "_comment": "GLM-5.2-FP8 thresholds for MI30X SGLang disaggregated. ISL=1024 OSL=1024 concurrency sweep: 8,16,32,64,128,256. Other ISL/OSL at CONC=64.",
+  "ISL=1024,OSL=1024,TP=16,CONC=8": {
```
solaiys: 🟠 Non-blocking heads-up: same TP mismatch as the kimi threshold file — these cells are keyed `TP=16` but `mi30x_sglang_distributed.json` sets `tensor_parallelism: "8"` for glm-52-fp8, so the perf table won't associate expected values (Expected/Result show `-`). Suggest aligning the keys to `TP=8`.

solaiys (commented):
Review focused on blocking issues (per my review checklist). Found 2 blocking issues and 1 non-blocking heads-up; details are left as inline comments.

- 🔴 `docker_lib.py`: `/petc/` typo breaks docker keyring removal.
- 🔴 `sglang_disagg_lib.py`: early-abort on inference errors is commented out (false-PASS regression).
- 🟠 (non-blocking) Kimi/GLM threshold files are keyed `TP=16` while the model config uses `tensor_parallelism: 8`, so the perf table won't associate expected values.

No other reviews were present when I looked, so nothing to deduplicate.

amd-droy:
Redundant. Kimi is not working for sglang yet.

PR #255, opened 2026-07-13
refactor(inference): move shared suite helpers under inference/utils

hnimra-amd:
Colocate lifecycle, cache probe, and results table modules with other inference utilities, update import paths across IX atom suites and report wiring, and fix ruff/pylint gates for `make test`.

## Motivation

Shared inference suite helpers (`inference_suite_lifecycle`, `cache_probe`, `inference_suite_results_table`) lived at the top level of `cvs/lib/inference/` while related config and parsing code already lives under `cvs/lib/inference/utils/`. This split made the layout harder to navigate and inconsistent as IX atom support grew.

This PR is the first step of the IX atom refactor: move those shared helpers into `utils/` and fix the resulting import paths. It also unblocks `make test` on `dev/dtni` by fixing ruff formatting (26 files) and ruff lint issues (12 pre-existing violations) so both `fmt-check` and `lint` pass.

### Module moves (`cvs/lib/inference/` → `cvs/lib/inference/utils/`)

| Module | Purpose |
|--------|---------|
| `inference_suite_lifecycle.py` | Lifecycle-as-tests stage helpers, `InferenceLifecycle`, pytest HTML hooks |
| `cache_probe.py` | Model-cache size probing (`du_bytes`) |
| `inference_suite_results_table.py` | End-of-session results table (`make_print_results_table`) |

`InferenceXAtomJob` remains at `cvs/lib/inference/inferencex_atom_orch.py` (unchanged).

### Import path updates

Consumers updated to `cvs.lib.inference.utils.*`:

- `cvs/tests/inference/inferencex_atom/conftest.py`
- `cvs/tests/inference/inferencex_atom/inferencex_atom_single.py`
- `cvs/tests/inference/inferencex_atom/_shared.py`
- `cvs/lib/report/inference_wiring.py`
- `cvs/lib/report/pytest_extras.py`
- `cvs/lib/report/presets/inferencex_atom.py`
- `cvs/lib/report/presets/_inference_suite_template.py`
- `cvs/lib/inference/unittests/test_inference_suite_lifecycle.py`

Doc/path references updated in `cvs/input/config_file/inference/inferencex_atom_single/README.md`.

### Formatting & lint (`make test` gates)

- **`ruff format`** (0.14.8) — 26 files reformatted so `ruff format --check .` passes.
- **`ruff check`** — resolved 12 issues:
  - moved `cache_probe` import to top of `inference_suite_lifecycle.py` (`E402`)
[...]

PR #257, opened 2026-07-16
feat(vllm): unify vllm_single + vllm_distributed into one topology-parametrized suite

atnair-amd:
## Jira

[AIMVT-247](https://amd-hub.atlassian.net/browse/AIMVT-247)

## Motivation

Replaces the standalone `vllm_single` and `vllm_distributed` suites with a single
unified `vllm` suite parametrized on topology (`nnodes`/`pipeline_parallel_size`
in config), so single-node and 2-node PP-distributed vLLM runs share one code
path instead of two divergent ones. This branch also carries the rebase of the
prior `atnair/vllm-distributed` work (PR #242) onto `dev/dtni`, plus fixes found
during live multi-node validation.

## Technical Details

**New unified suite (replaces `vllm_single.py` + `tests/inference/vllm_distributed/`):**
- `cvs/lib/inference/vllm_job.py` — unified `VllmJob`: distributed flags (topology,
  IB HCAs, master addr/port) are only added when `nnodes > 1`; `run_client` /
  `wait_client_complete` / `parse_results` always execute head-only via
  `exec_on_head`.
- `cvs/lib/inference/utils/vllm_config_loader.py` — unified `VariantConfig` for
  both single-node (`nnodes=1`, `pp=1`) and distributed (`nnodes>1`, `pp>1`);
  `ib_netdev` required when `nnodes>1`; `cell_key` only emits a `PP=` segment
  when `pp>1`.
- `cvs/lib/utils/ib_discovery.py` — `discover_ib_hca_names()` via `ibv_devinfo -l`
  with an `/sys/class/infiniband` sysfs fallback for images that ship without
  `libibverbs-dev`; fails loudly on empty or asymmetric HCA lists across nodes.
- `cvs/tests/inference/vllm/vllm.py` (renamed from `vllm_single.py`) — adds a
  `test_discover_topology` lifecycle stage that runs once per sweep for
  distributed configs (skipped on single-node) and validates any explicit
  `ib_hca_devices` list against what's discovered.
- Ray added as an alternate `distributed-executor-backend` (alongside `mp`) for
  `nnodes>1, pp=1`: bootstraps a Ray cluster (head + workers), runs `vllm serve`
  on the head only, tears down via `ray stop`.
- Server reuse across sweep cells: `VllmJob.server_signature()` identifies the
  rank-agnostic server identity (argv minus `--node-rank`, plus env), so
  consecutive cells that only change client-side concurrency skip a full
  stop/start/weight-reload (~9 min/cell saved on multi-concurrency sweeps).
- Legacy `vllm_single.py` + `tests/inference/vllm_distributed/` suites deleted.

## Test Plan

- `python -m py_compile` across the full repo tree — clean.
- `python run_all_unittests.py` (`.test_venv`) — 650 tests, 2 pre-existing
  failures in `test_inferencing_config_loader.py` unrelated to this branch
  (confirmed via side-by-side comparison against the pre-rebase commit).
- Live 2-node MI300X validation runs (linked in ticket)

## Test Result

650/652 unit tests passing (2 pre-existing, unrelated failures). Compiles
clean. Lint clean on all files touched by this branch.


cvs/lib/inference/vllm_job.py
```diff
@@ -0,0 +1,573 @@
...
+                        log.info("[%s rank%d server.log] %s", h, rank, line)
+                if self.EARLY_FAILURE_RE.search(output or ""):
+                    raise RuntimeError(f"vllm server early failure on {h} (rank {rank}): {(output or '')[-500:]}")
+            out = self.orch.exec(
+                f"grep -m1 -iE {shlex.quote(self.FATAL_LOG_RE.pattern)} {shlex.quote(rank_log)}",
+                detailed=True,
+                hosts=[host],
+            )
+            for h, r in (out or {}).items():
+                if r.get("exit_code") == 0 and r.get("stdout", "").strip():
```
solaiys: 🔴 **Blocking — wrong dict key disables this fatal-error check.** The orchestrator's `detailed=True` exec returns `{"output", "exit_code"}` (every other call site uses `output`/`exit_code` — e.g. lines 308, 317, 367 here, and `container.py`). `r.get("stdout", "")` is therefore always empty, so this `FATAL_LOG_RE` branch can never raise. A fatal `RuntimeError: … engine` (matched by `FATAL_LOG_RE` but not by `EARLY_FAILURE_RE`) won't fast-fail; the run instead hangs until the full readiness poll cap — exactly the multinode failure path this PR targets. Fix: use `r.get("output", "")`.

solaiys (commented):
Review focused on blocking issues (per my review checklist). The new unified vLLM logic itself looks solid (topology gating on `nnodes>1`, head-only client stages, `server_signature()` reuse across concurrency-only cells, duplicate `--max-model-len` resolved, IB discovery fails loudly on empty/asymmetric HCA lists). Two blocking issues, one left inline and one below.

🔴 **Blocking — reintroduces modules that the merged PR #255 deleted, landing duplicate/stale copies on `dev/dtni`.**

I checked each path on both refs. `dev/dtni` has these modules **only** at their #255 locations, but this PR head carries them at **both** the new and the pre-#255 old paths:

| Module | on `dev/dtni` | on this PR head |
|---|---|---|
| `cvs/lib/inference/inferencex_atom_orch.py` (old) | absent | present |
| `cvs/lib/inference/cache_probe.py` (old) | absent | present (byte-identical dup) |
| `cvs/lib/inference/inference_suite_results_table.py` (old) | absent | present (content-divergent) |
| `cvs/lib/inference/utils/inferencex_atom_config_loader.py` (old) | absent | present (content-divergent) |
| `cvs/lib/inference/utils/inferencex_atom_parsing.py` (old) | absent | present (content-divergent) |

The authoritative #255 copies (`inference/inferencex_atom/*`, `inference/utils/*`) also exist on the head, so merging silently adds divergent duplicate source modules — a regression of #255's cleanup and a correctness hazard (edits to one copy won't touch the other). This looks like an unclean rebase (the branch carries both #255's moves and its own pre-move copies). Fix: rebase onto current `dev/dtni` and drop the reintroduced old-path files so only the #255 locations remain.

cvs/lib/inference/unittests/test_vllm_job_ray_backend.py
```diff
@@ -0,0 +1,1350 @@
+'''
```
solaiys, line 1: 🔴 **Blocking (fmt gate) — this new file fails `make fmt-check`.** `ruff format --check .` reports `Would reformat: cvs/lib/inference/unittests/test_vllm_job_ray_backend.py`, so the PR does not pass the formatting gate as-is (this contradicts the PR description's "fmt-check pass" claim — likely edited after the last format run). Fix: run `make fmt` and commit the result.

solaiys (commented):
Re-review status — both earlier blocking findings are ✅ **verified resolved** on this head (`26e88c0b`):

- ✅ Duplicate/stale reintroduced modules — all 5 pre-#255 old-path files are now absent; only the authoritative #255 locations remain (file count 19→14, additions 3754→2779).
- ✅ `vllm_job.py` `_check_early_failure` now reads `r.get("output", "")` (was `stdout`), so the `FATAL_LOG_RE` fast-fail branch is live again.

**`make fmt` / `make lint` check (run against this head):**

🔴 **`make fmt-check` fails on a file this PR adds** — `cvs/lib/inference/unittests/test_vllm_job_ray_backend.py` (see inline comment). `ruff format --check` reports it would reformat. Please run `make fmt` and commit.

⚪ **Not caused by this PR (pre-existing on `dev/dtni`):** `cvs/lib/utils/gpu.py` (fmt) and `cvs/lib/utils/unittests/test_gpu.py` (fmt + all 5 `ruff` E401 lint errors). I confirmed via `git diff origin/dev/dtni HEAD` that both files are byte-identical to `dev/dtni` and are not in this PR's diff — so the `make lint` failure and 2 of the 3 `fmt-check` files are independent of #257 and should not block it (worth a separate cleanup on `dev/dtni`).

Net: the only gate issue owned by this PR is the `ruff format` violation in `test_vllm_job_ray_backend.py` — a one-command fix.

solaiys (approved):
please check with make fmt-check and lint check.. and address the files in your PR.
i think a minor one is present now..
otherwise its LGTM...

please fix the ruff fmt issue before merge.

PR #258, opened 2026-07-16
feat(vllm): GPU metrics polling + fix per_gpu_throughput pp-undercounting

atnair-amd:
## Summary
- `per_gpu_throughput` in `to_client_metrics` divided total token throughput by `tp` alone, undercounting GPU count on any pipeline-parallel run (actual GPUs = `tp * pp`). `pp` is now a required keyword-only argument.
- Widened the vLLM Ray-backend regression test to assert `pp` explicitly (subTest over `pp=1,2`) so a broken `pp` passthrough can't silently regress again — the prior version only asserted `tp`/`isl`.
- Scope is vLLM only. InferenceX ATOM's `to_client_metrics` call sites are not updated for the new required `pp` kwarg and are left broken by this change (out of scope here).
- **Added GPU metrics polling to the vLLM suite**: `test_vllm_inference` now takes pre/post-model-load VRAM snapshots and polls `amd-smi` in the background while the client runs; a new `test_gpu_metric` test reports one HTML row per GPU metric per sweep cell (`peak_gpu_memory_mb`, `model_load_memory_mb`, `model_load_s`, `gpu_bandwidth_util_pct`, `gpu_compute_util_pct`). Threshold-coverage validation (`vllm_config_loader.py`) now gates `gpu.*` metrics the same way it already gated `client.*` metrics. amd-smi is invoked via the container-routed `orch.exec`/`orch.exec_on_head` — GPU device files (`/dev/kfd`, `/dev/dri`) are passed through into the container, so no host-vs-container bypass routing is needed.

- **Fixed HW-observed SSH race in the GPU-metrics poller**: replaced the thread-based `poll_gpu_metrics()` with `start_gpu_poller()`/`stop_and_collect_gpu_poller()` — a detached remote script per node instead of a second thread sharing the orchestrator's SSH transport, eliminating a `SessionError(OutOfBoundaryError())` race seen on real hardware, with round-aligned multi-node collection and a never-raises collect contract.
- **Poller `/tmp` file hygiene**: the poller's remote script/log filenames are now scoped by local SSH user (not just run_id), the log is truncated at launch instead of only appended to, and both files are `rm -f`'d after collection — avoiding the shared-node `/tmp` file-collision/permission-error failure mode reported against Fremont's conda-based CVS runs.

## Jira
- [AIMVT-263](https://amd-hub.atlassian.net/browse/AIMVT-263) — Wire GPU metrics polling into the vLLM inference suite
- [AIMVT-264](https://amd-hub.atlassian.net/browse/AIMVT-264) — Fix per_gpu_throughput undercounting for pipeline-parallel vLLM runs
- [AIMVT-268](https://amd-hub.atlassian.net/browse/AIMVT-268) — Fix HW-observed SSH race in vLLM GPU-metrics poller
- All are subtasks of [AIMVT-245](https://amd-hub.atlassian.net/browse/AIMVT-245) — [CVS] GPU metrics polling integration for inference validation suites

## Test plan
- [x] `python3 -m unittest discover -s cvs/lib/inference/unittests -p "test_*.py"` — 188 tests, only known-out-of-scope failures: `test_inferencing_config_loader.py` (2 pre-existing, unrelated failures confirmed via `git stash` against commit `5134be1`) and `test_inferencex_atom_orch_parse.py` (3 failures caused by this PR's `pp` requirement — ATOM is not being updated here, tracked separately)
- [x] `python3 -m unittest discover -s cvs/lib/utils/unittests -p "test_*.py"` — 126/126 pass (gpu.py polling/aggregation logic)
- [x] `python3 -m unittest discover -s cvs/core/orchestrators/unittests -p "test_*.py"` — 57/57 pass
- [x] Mutation-tested: reverting the `pp` fix in `vllm_job.py` causes the regression test to fail, confirming it actually guards the fix
- [x] Hardware validation on demo cluster: full vLLM suite run against DeepSeek-R1-0528 in a real `tp8pp2` config (8-way tensor-parallel, 2-way pipeline-parallel, 2-node Ray cluster) — **43 passed, 1 skipped, rc=0**. Confirms `per_gpu_throughput` and the rest of the metrics pipeline compute and report correctly with `pp=2` on live hardware (total throughput 823.47 tok/s, TTFT/TPOT/ITL all populated in the results table).
- [x] `pytest --collect-only` against `cvs/tests/inference/vllm/vllm.py` with a real vLLM config confirms `test_gpu_metric` collects correctly (5 GPU metric rows per cell) alongside the existing 44 tests, with no collection errors.

- [x] GPU-poller unit tests: `python3 -m unittest discover -s cvs/lib/utils/unittests -p "test_gpu.py"` — 90/90 pass (includes 5 new tests for user-scoped markers, stale-log truncation, and post-collection remote cleanup)
- [x] Full repo gate (`make test` / `run_all_unittests.py`): 715 tests, same 2 known-out-of-scope pre-existing failures as above, no regressions from the `/tmp` hygiene fixup
- [x] Hardware validation of the poller round trip: 2-node DeepSeek-R1 tp8pp2 run, 800 prompts, 1000/1000 ISL/OSL, concurrency 16 — **48 passed, 1 skipped, rc=0**, 0 failed requests, total throughput 801.47 tok/s. GPU poller collected 133 readings across both nodes with 0 failures (peak_vram=3,962,984 MB, compute_util=96.0%, bw_util=7.2%), confirming the SSH-race fix holds under real multi-node concurrent load. Run bundle attached to [AIMVT-268](https://amd-hub.atlassian.net/browse/AIMVT-268).

## Known out-of-scope breakage
InferenceX ATOM's `to_client_metrics` call sites do not pass `pp` and will fail after this change (ATOM has no pipeline-parallel concept). This is expected and intentionally left unfixed here — scope is vLLM only.


solaiys (commented):
Review focused on blocking issues. The core fix is correct (`per_gpu_throughput` now divides by `tp * pp` via `_gpu_count()` with safe `None` degradation, and the Ray-backend regression test was widened to assert `pp` over `{1,2}`). One blocking issue + two non-blocking notes.

🔴 **Blocking — this knowingly breaks the InferenceX ATOM suite at runtime, not just in tests.** Making `pp` a required kwarg on the shared `vllm_parsing.to_client_metrics` (see inline comment) breaks ATOM's callers, which are not updated:
- `cvs/lib/inference/inferencex_atom/inferencex_atom_parsing.py:72` delegates via `_vllm_to_client_metrics(raw, tp=tp, isl=isl)` — no `pp`.
- `cvs/lib/inference/inferencex_atom/inferencex_atom_orch.py:505` calls that wrapper.

So `InferenceXAtomJob.parse_results()` will raise `TypeError: to_client_metrics() missing 1 required keyword-only argument: 'pp'` on **every** ATOM run — the 3 `test_inferencex_atom_orch_parse.py` failures are the unit-test symptom of a real runtime break. Merging to `dev/dtni` leaves a shipping suite broken; "out of scope" doesn't help a shared branch. Fix (either): update ATOM's two call sites to pass `pp="1"` (ATOM has no pipeline-parallel concept), or give `pp` a default of `"1"` in `vllm_parsing.to_client_metrics`. The first keeps the explicit-topology design and is ~2 lines.

⚪ Non-blocking notes (not gating):
- `vllm_config_loader.py` now also requires `gpu.*` gated-metric specs in every threshold cell. Only affects configs with `enforce_thresholds=true` (shipped examples use `false`, so they only warn) and fails loudly — an intentional extension of the existing `client.*` gating; worth a heads-up for anyone with an enforcing config.
- The GPU poll runs `orch.exec_on_head` from a daemon thread while `wait_client_complete` polls the client log on the main thread (shared `Pssh`). It degrades gracefully (per-iteration exceptions → `None`) and you validated live on 2 nodes, so not blocking — just worth keeping in mind if intermittent poll failures appear.

cvs/lib/inference/utils/vllm_parsing.py
```diff
@@ -41,7 +41,20 @@ def _safe_div(num, den):
...
+    (missing, empty string, non-numeric string, wrong type) so callers can
+    feed the result straight into `_safe_div`'s existing None/0 guard.
+    """
+    try:
+        return int(tp) * int(pp)
+    except (TypeError, ValueError):
+        return None
+
+
+def to_client_metrics(raw, *, tp, isl, pp):
```
solaiys: 🔴 **Blocking — making `pp` required (no default) breaks the InferenceX ATOM suite at runtime.** ATOM's `inferencex_atom_parsing.to_client_metrics` (line 72) delegates here via `_vllm_to_client_metrics(raw, tp=tp, isl=isl)` — without `pp` — and `inferencex_atom_orch.py:505` drives it, so `InferenceXAtomJob.parse_results()` will raise `TypeError: ... missing 1 required keyword-only argument: 'pp'` on every ATOM run once this merges to `dev/dtni`. Fix: either default `pp` to `"1"` here, or update ATOM's wrapper + call site to pass `pp="1"` (ATOM has no pipeline-parallel concept).

cvs/tests/inference/vllm/conftest.py
```diff
@@ -126,6 +126,11 @@ def inf_res_dict():
     return {}
 
 
+@pytest.fixture(scope="module")
+def gpu_metrics_snap():
+    return {}
```
amd-droy: is this needed? Otherwise, you can remove it.

cvs/lib/inference/utils/vllm_parsing.py
atnair-amd: Fixed in ce79246c: `pp` now defaults to `"1"` in `vllm_parsing.to_client_metrics` rather than requiring ATOM's call sites to be updated. ATOM's wrapper (`inferencex_atom_parsing.to_client_metrics`) never passes `pp`, so it now silently gets `pp="1"` -> `_gpu_count(tp, 1) == tp`, reproducing the pre-fix `ttot/tp` formula exactly for ATOM. vLLM call sites still pass `pp` explicitly per the docstring.

Reran the full gate: `test_inferencex_atom_orch_parse.py`'s 3 failures are gone (188/126/57, only the 2 known pre-existing `test_inferencing_config_loader.py` failures remain, confirmed unrelated to this branch).

Replaced `test_pp_is_required_keyword_arg` with `test_pp_defaults_to_one` to cover the new default behavior.

cvs/tests/inference/vllm/conftest.py
atnair-amd: Removed in 0980af20.

solaiys (approved):
LGTM — my earlier blocking issue is resolved on this head.

✅ **ATOM runtime break fixed** (`ce79246c`): `to_client_metrics(raw, *, tp, isl, pp="1")` now defaults `pp`, so ATOM's callers (`inferencex_atom_parsing.py:72` → `_vllm_to_client_metrics(raw, tp=tp, isl=isl)`) no longer hit the missing-kwarg `TypeError`. The vLLM fix is preserved: `vllm_job.parse_results()` still passes `pp=self.pp` explicitly so `per_gpu_throughput = ttot / (tp*pp)` stays correct on PP runs, and the `test_vllm_job_ray_backend` `pp∈{1,2}` regression test still guards the passthrough. The unit test was updated to `test_pp_defaults_to_one` accordingly.

✅ Unused `gpu_metrics_snap` fixture removed (`0980af20`) per @amd-droy — confirmed no remaining references in `conftest.py`/`vllm.py`.

The two non-blocking notes from my earlier review still stand but don't gate merge: the `gpu.*` threshold gating only affects `enforce_thresholds=true` configs (and fails loudly), and the threaded GPU poll degrades gracefully. Nice work.

atnair-amd:
Added a fix for a real HW-observed SSH race: the GPU-metrics poll thread and the main thread's client-log-tail polling shared the orchestrator's gevent-based SSH transport, causing `SessionError(OutOfBoundaryError())` under load.

Replaced `poll_gpu_metrics()` (thread-based) with `start_gpu_poller()`/`stop_and_collect_gpu_poller()`: a detached remote background script per node snapshots `amd-smi` to a file, read back via ordinary sequential `exec`/`exec_on_head` calls — no second OS thread, no shared-SSH race. Also extends to real multi-node round-aligned collection (previous single-node-only limitation).

Unit-tested (85 tests, all green) — **HW validation of the actual remote-poller round trip (real amd-smi, real SSH, real multi-node) is still needed** before this is considered fully validated, same caveat as the original GPU-metrics polling work.

solaiys (commented):
Re-review of the race-condition fix (`c144cc80`) — ✅ resolved.

The earlier design ran the GPU poller in a daemon thread calling `orch.exec_on_head` while the main thread simultaneously polled the client log on the **same shared `Pssh`/SSH transport** — a data race on the shared connection. The fix eliminates it cleanly:

- `threading` (and the `threading.Event`/poll-thread/`join`) is removed from `vllm.py`.
- New `start_gpu_poller()` writes a small bash script and launches it **detached on each node** (`nohup … &`); `stop_and_collect_gpu_poller()` later `pkill`s it and reads back the per-node log file. All orchestrator SSH calls now happen **sequentially on the main thread** — the shared transport is never used from two threads at once.
- Orphan-safety: the remote script self-terminates via `hard_cap_s` (default 4h) if collection never runs.
- `run_id` is sanitized (`[^A-Za-z0-9_.-]` → `_`) and the script is `shlex.quote`d into the write command, so a pytest nodeid can't inject shell; `stop_and_collect_gpu_poller` is explicitly `finally`-safe (never raises on transport failure).

Both of my prior blocking findings (ATOM `pp` break, unused fixture) remain resolved, and I don't see any new blocking issues in the poller rewrite.

LGTM from my side — no blocking issues.

solaiys:
@atnair-amd ,
One General Note on how we handle files in /tmp during tests.
When your tests create files in /tmp (**e.g., script_path = f"/tmp/{marker}.sh"**), please follow one of these practices:

- Use mktemp to generate a unique temporary directory, so there's no risk of path collisions.
- Use user-specific paths like /tmp/{userid}/ to isolate your files.
- Clean up after yourself — delete any files your tests create in /tmp when they're done.

**Why this matters:** 
In containerized environments, this isn't an issue since each run gets its own isolated /tmp. However, in shared environments — such as conda setups or Docker volume mounts where /tmp:/tmp is mapped to the host — files written to /tmp persist on the node, owned by the user who created them. When the next person runs the same test on that node, they'll hit a permission error trying to write to the same path.
We ran into exactly this problem with the Fremont CVS internal use cases running on conda environments, and it caused some frustrating test failures.
Let's keep this in mind going forward.

amd-droy (approved):
Looks good to me. Thanks @atnair-amd

PR #260, opened 2026-07-16
Replace manual orch_sudo config with automatic per-command sudo fallback

atnair-amd:
## Summary
- Removes the manual `orch_sudo` cluster-file config key and replaces it with automatic, probe-once sudo detection: `BaremetalOrchestrator.sudo_prefix()` probes `sudo -n true` on all hosts a single time per orchestrator instance, caches the head node's result, and returns `'sudo -n '` or `''` for every subsequent privileged command. No config needed on sudo-enabled or sudoless clusters.
- Fixes `build_mpi_cmd()`'s hostfile handling: the stale-hostfile removal and the new hostfile write now share one `sudo_prefix()` value (previously only the removal used it) and both raise `RuntimeError` on non-zero exit code, instead of silently leaving a leftover/wrong-owner `/tmp/mpi_hosts.txt` in place.
- Fixes `sudo_prefix()`'s disagreement handling: when hosts disagree on passwordless-sudo availability, the cached answer is now deterministically the head node's result (still logging a warning) instead of an arbitrary dict-iteration-order pick.
- Adds optional Docker registry login (`runtime.args.registry: {username, password_file, server}`) so `setup_containers()` can log in before pulling a private image (e.g. `rocm/ufb-private`). Password is piped via `--password-stdin` from a file path, never embedded in cluster-file JSON. Skipped when `image_tar` is set.
- Updates cluster file templates, README, and docs to drop `orch_sudo`, describe the actual probe-once/cache behavior, and document `registry`.

Jira: AIMVT-261

## Test plan
- [x] `make fmt-check && make lint && make test` — all clean
- [x] `python -m unittest cvs.core.orchestrators.unittests.test_baremetal cvs.core.runtimes.unittests.test_docker` — includes regression tests for the hostfile race, head-node-cache determinism, and registry login
- [x] Live `docker pull` of a 47.3GB private image through `registry_login()` + sudo-fallback path, verified via `docker images` digest


atnair-amd:
Follow-up fix: `with_sudo_fallback` rendered `cmd || sudo -n cmd`, so any legitimate command failure (not just permission-denied) re-ran the exact same command a second time under sudo. For non-idempotent caller payloads (`DockerRuntime.exec`/`exec_cmd_list`/`exec_on_head`), this could silently double-run arbitrary commands whenever they failed legitimately and passwordless sudo happened to be configured.

Replaced with `BaremetalOrchestrator.sudo_prefix()`: probes `get_passwordless_sudo_status()` once per orchestrator instance, caches the result, and builds each command deterministically (plain OR sudo-prefixed, never both via `||`). Migrated all 10 call sites in `docker.py`/`baremetal.py` and deleted `with_sudo_fallback`. Added regression tests proving the probe fires at most once across multiple exec-family calls.

26/26 targeted tests pass; full suite (432 tests) unaffected aside from 3 pre-existing, unrelated test-discovery path errors.

solaiys (approved):
LGTM — no blocking issues found. Focused on the security-sensitive parts (registry credential handling, sudo command construction, and the `sudo docker` → `sudo_prefix()` migration).

**Verified:**
- **Credential handling (`registry_login`)** — password is piped via `cat <file> | docker login … --password-stdin`, never on the command line/process list; `password_file`, `username`, and `server` are all `shlex.quote`d; the exec uses `print_console=False`; and the docs/example correctly put the token in a host-side file, not in cluster JSON.
- **Empty `server` (Docker Hub) path** — `docker login '' --username …` is safe: docker's CLI treats an empty server address as the default index server, so the documented `"server": ""` case works.
- **`sudo_prefix()` probe** — `get_passwordless_sudo_status()` returns real booleans, cached via `self._needs_sudo is None`, so the probe fires at most once per orchestrator; host-disagreement warns and deterministically uses the first host's answer (consistent with the passwordless-or-none model). The original `cmd || sudo -n cmd` double-run form is fully gone.
- **Migration completeness** — zero remaining `with_sudo_fallback` references and no hardcoded `sudo docker`/`sudo rm` left in `docker.py`/`baremetal.py`; all call sites go through `sudo_prefix()`, and `registry` never leaks into `docker run`.

Nice work on the `--password-stdin` + host-side file approach and the probe-once rework.

PR #261, opened 2026-07-16
feat(vllm): wire unified vLLM suite into inference report engine

atnair-amd:
## Summary
- PR #257 already brings in the generic inference report engine (`cvs/lib/report/`) via the `dev/dtni` merge, but no suite preset exists for `vllm`, so `cvs run vllm` never rendered a run deck (HTML/JSON dashboard, viewer, CI summary).
- Adds the missing preset (`cvs/lib/report/presets/vllm.py`) plus vLLM-specific tier/column glue in `vllm_parsing.py`: `VLLM_RESULTS_COLUMNS` (mirrors the existing console table in `_shared.py`), `METRIC_TIERS`/`METRIC_TIER_ORDER` seeded from the existing `GATED_METRICS` set (pinned by a unit test as an exact bijection), and `tier_metric_specs()`.
- The preset overrides the report engine's default lifecycle labels (`sshd_setup`/`client_complete`), since the unified vLLM suite instead records `topology_discovery` and has no `client_complete` stage.
- Removes the suite-local `pytest_runtest_makereport` lifecycle-table hookwrapper in `cvs/tests/inference/vllm/conftest.py` — confirmed byte-for-byte equivalent to the shared `attach_inference_suite_lifecycle_table` that root `conftest.py` now invokes automatically once the preset registers, so keeping both would have rendered a duplicate table per test row.
- Render-only: none of this changes pass/fail or threshold enforcement.

## Test plan
- [x] New unit tests (`cvs/lib/inference/unittests/test_vllm_report_preset.py`, 9 cases) — pin the tier partition against `GATED_METRICS`, lifecycle label correctness, and end-to-end auto-register resolution for stem `vllm`.
- [x] Offline dry render: synthesized a `VariantConfig` + 2-cell `inf_res_dict`, called `write_report()` directly — confirmed `vllm_run_deck.html`/`.json`/`_viewer.html`/`_summary.html` all generate, gate matrix tiers compute correctly, and the lifecycle timeline shows only the stages this suite actually records (no empty legacy `sshd_setup`/`client_complete` slots).
- [x] `make ut` equivalent (`run_all_unittests.py`): 659 tests, only the same 2 pre-existing/unrelated failures present on the unmodified branch tip (`test_inferencing_config_loader.py`).
- [x] `make test` CLI suite (`test_cli.sh`): all 46 checks pass, including `cvs list vllm`.
- [x] `ruff check`/`ruff format --check` on all touched files: clean. (Repo-wide `make fmt-check`/`make lint` currently fail on 3 unrelated pre-existing files (`gpu.py`, ray-backend files) already broken on the unmodified branch tip — out of scope here.)

## Jira
- [AIMVT-260](https://amd-hub.atlassian.net/browse/AIMVT-260): tracks this vLLM run-deck report generation work (parent epic AIMVT-202).

## Known follow-up
- Live TensorWave smoke test of this preset surfaced 4 caught-but-logged `shutil.SameFileError`s (one per run-deck artifact) in the shared inference report engine's `add_html_to_report()` — pre-existing bug from PR #257, non-fatal, out of scope for this PR. Not yet filed separately.

### DIFF
--- a/cvs/lib/inference/unittests/test_vllm_report_preset.py
+++ b/cvs/lib/inference/unittests/test_vllm_report_preset.py
@@ -1,0 +1,102 @@
+'''
+Copyright 2025 Advanced Micro Devices, Inc.
+All rights reserved.
+'''
+
+import unittest
+
+from cvs.lib.inference.utils.vllm_parsing import (
+    CLIENT_METRICS,
+    GATED_METRICS,
+    METRIC_TIER_ORDER,
+    METRIC_TIERS,
+    VLLM_RESULTS_COLUMNS,
+    tier_metric_specs,
+)
+from cvs.lib.report.presets.vllm import VLLM_REPORT_CONFIG
+
+
+class TestVllmReportPreset(unittest.TestCase):
+    def test_results_columns_fixed_positional_prefix(self):
+        fixed = VLLM_RESULTS_COLUMNS[:7]
+        self.assertEqual(
+            fixed,
+            (
+                ("Model", None),
+                ("GPU", None),
+                ("ISL", None),
+                ("OSL", None),
+                ("Policy", None),
+                ("Conc", None),
+                ("Host", None),
+            ),
+        )
+
+    def test_metric_tiers_subset_of_tier_order(self):
+        self.assertTrue(set(METRIC_TIERS) <= set(METRIC_TIER_ORDER))
+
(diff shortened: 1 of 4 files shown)


hnimra-amd:
LGTM, The results look good too!

hnimra-amd (approved):

PR #265, opened 2026-07-21
Added exec_cmd_list to both docker runtime and container orchestrator

sukesh-amd:
## Submission Checklist

### DIFF
--- a/cvs/core/orchestrators/container.py
+++ b/cvs/core/orchestrators/container.py
@@ -103,8 +103,7 @@ class ContainerOrchestrator(BaremetalOrchestrator):
 
         # Start with dynamic user-specific mounts (from original command)
         volumes = [
-            f"/home/{user}:/workspace",  # User home directory
-            f"/home/{user}/.ssh:/host_ssh",  # SSH keys for multinode
+            f"/home/{user}/.ssh:/host_ssh"  # SSH keys for multinode
         ]
 
         # Merge with config.json runtime args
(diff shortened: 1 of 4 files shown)


solaiys:
## Review (head `3b92dbc3`)

Reviewed the diff and specifically checked whether removing the `/home/{user}:/workspace` mount affects existing tests. **Blocking issues only** below.

### Blocking issues

- **New `exec_cmd_list` is sound.** `shlex` is imported in `docker.py`, and each command is wrapped exactly like the existing `exec` (`docker exec … bash -c {shlex.quote(cmd)}`). The underlying `MultiProcessPssh.exec_cmd_list` validates `cmd_list` length against `host_list`/`reachable_hosts` (raises `ValueError` otherwise) and preserves ordering, so positional host↔command correspondence holds. `container.py.exec_cmd_list` correctly guards on `self.container_id`.

### Does removing `/home/{user}:/workspace` break any existing tests?

**No existing test breaks.** Checked every test that could touch it:

- `cvs/core/runtimes/unittests/test_docker.py` references `/home/u:/workspace` in 3 places, but each passes it as an **explicit input** to `rt.setup_containers(..., volumes=[...])` (a hand-built fixture). None call `ContainerOrchestrator.get_volumes()` or assert its output, so they keep passing unchanged.
- `cvs/core/orchestrators/unittests/test_container.py` has **zero** references to `volumes`/`get_volumes`/`workspace` — it doesn't exercise the mount.
- No test asserts `get_volumes()` output, and no CVS code pins the container `working_dir`/`WORKDIR` to `/workspace` (`aorta.py` uses the configurable `container_mount_path`).

### Non-blocking context (not a merge blocker)

- **Runtime behavior change:** the container no longer bind-mounts the user's home at `/workspace` (the `.ssh:/host_ssh` mount is retained). The training frameworks referenced as `/workspace/Megatron-LM`, `/workspace/torchtitan`, `/workspace/maxtext` are **image-baked paths**; the old `/home/$user:/workspace` mount would have *shadowed* them, so this removal looks like an intended fix rather than a regression. I found no flow that requires user-home visible at `/workspace`, and per-config volume mounts (e.g. `/tmp/TRAINING_LOGS:/workspace/maxtext/output`) are independent of the home mount.
- **Docs now stale (doc-only):** `cvs/input/cluster_file/README.md` and `docs/reference/configuration-files/cluster-file.rst` still state the container "always also receives `/home/$user:/workspace`". Worth updating in this PR.

_Reviewed 1 PR: 0 blocking issues — clean._

solaiys (approved):
LGTM

PR #266, opened 2026-07-22
Add optional Docker registry login for private image pulls

atnair-amd:
## Summary
- Adds optional Docker registry login (`runtime.args.registry: {username, password_file, server}`) so `setup_containers()` can log in before pulling a private image (e.g. `rocm/ufb-private`). Password is piped via `--password-stdin` from a file path, never embedded in cluster-file JSON. Skipped when `image_tar` is set.
- Updates cluster file templates and README to document the new `registry` key.

Split out of #260, where it was reviewed as unrelated scope creep on top of that branch's sudo-fallback fix.

Jira: AIMVT-265

## Test plan
- [x] `make fmt-check && make lint && make test` — all clean

### DIFF
--- a/cvs/core/orchestrators/factory.py
+++ b/cvs/core/orchestrators/factory.py
@@ -84,7 +84,8 @@ class OrchestratorConfig:
                 "network": "host",
                 "ipc": "host",
                 "ulimit": ["memlock=-1"],
-                "privileged": true
+                "privileged": true,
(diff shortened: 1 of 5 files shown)


atnair-amd:
Reverting the split — registry login has been validated by solayis alongside the sudo-fallback work and is being kept in #260 instead. Closing this PR; branch will be left as-is for reference / can be deleted.

PR #267, opened 2026-07-22
Add OpenAI-compatible smoke test to unified vllm suite

atnair-amd:
## Summary
- Adds `VllmJob.probe_openai_endpoints()`, reusing the shared `OpenAIProbe` helper already used by the sglang suite, driven through `orch.exec_on_head` instead of `docker exec`/Pssh.
- Wires it in as a new lifecycle stage (`test_openai_compatible_smoke`, rank 4) that brings up a short-lived server at a small fixed cell (isl=128/osl=32) and probes GET `/v1/models`, POST `/v1/chat/completions`, POST `/v1/completions`, and structured-JSON output before the full sweep runs.
- Always stops its server in a `finally`, so a smoke failure can't leave a stray process for `test_vllm_inference`'s first cell to inherit.

Ticket: AIMVT-267

## Test plan
- [x] `python -m unittest cvs.lib.inference.unittests.test_vllm_job_server_reuse -v` — 23/23 pass (16 existing + 7 new `TestProbeOpenAIEndpoints` cases)
- [x] `python run_all_unittests.py` — 690 tests, same 2 pre-existing failures in `test_inferencing_config_loader.py` (unrelated to this change, present on `dev/dtni` HEAD)

### DIFF
--- a/cvs/lib/inference/unittests/test_vllm_job_server_reuse.py
+++ b/cvs/lib/inference/unittests/test_vllm_job_server_reuse.py
@@ -8,8 +8,10 @@ Unit tests for cvs.lib.inference.vllm_job.VllmJob server-command construction:
   - _flatten_serve_args boolean handling and log-level pass-through
   - _check_early_failure tail emission and CLI parse error detection
   - RoleServer.serve_args log-level validator
+  - probe_openai_endpoints(), the OpenAI-compatible HTTP smoke probe
 '''
 
+import json
 import unittest
 import unittest.mock as mock
 from types import SimpleNamespace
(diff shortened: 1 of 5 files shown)


solaiys (approved):
LGTM — no blocking issues found. Small, focused change that reuses the shared `OpenAIProbe` helper for a pre-sweep OpenAI-compatible smoke stage.

**Verified:**
- **`OpenAIProbe` API usage is correct** — all four call sites match the real signatures in `model_query_lib.py`: `probe_script(port, model)`, `log_results(results, logger)`, `check_results(results, *, port, logger) -> (ok, err)`, and `summarize_results(results, ok, err) -> list[str]`. `results` is built as `{step: (int(code), body)}`, matching what those expect.
- **No concurrency risk** — the smoke stage is fully sequential (`stop → build → start → wait_ready → probe`, `finally: stop`); no threads/async/background jobs or shared-transport reuse introduced.
- **Fails loud, no silent green** — missing/malformed output, non-JSON last line, bad shape, or a failed check all raise `RuntimeError`, and `lifecycle.failed` is set so downstream stages skip.
- **No stray server** — `job.stop_server()` in `finally` guarantees the short-lived smoke server is stopped on success or failure; it doesn't set `lifecycle.live_server_sig`, so `test_vllm_inference`'s first cell still does a correct fresh bringup.
- **Injection surface** — probe script is base64-encoded + `base64 -d`'d (no heredoc), and `out_dir`/`probe_path`/`b64` are `shlex.quote`d; `model_id`/`port` come from validated config.
- **Ordering** — `test_openai_compatible_smoke` pinned at rank 4 with the rest shifted down consistently.

Minor non-blocking note: the smoke reuses the full `wait_ready` budget, so a broken server fails in the same time as a sweep cell rather than truly "fast" — fine as-is.

PR #268, opened 2026-07-22
feat(inference): lm-eval-harness based accuracy evaluation + vllm suite integration 

atnair-amd:
Jira: AIMVT-269

## Summary
Adds an lm-eval-harness based accuracy evaluation capability + vLLM suite integration, opt-in via config.json.

- **Config schema** (`accuracy_config.py`): `AccuracyTask`/`AccuracyConfig` (pydantic, `extra=forbid`) hold task selection only — id/task/num_fewshot/metadata/include_path/num_concurrent/apply_chat_template/gen_kwargs. Rejects duplicate task ids.
- **Result parsing** (`lm_eval_parsing.py`): `project()` auto-discovers real numeric metrics from an lm-eval `results.json` payload and flattens them into a single dict, keyed `task.metric__filter`.
- **Command construction + orchestration** (`lm_eval_job.py`): builds the `lm_eval` CLI invocation (local-completions vs local-chat-completions, model_args, gen_kwargs, metadata) and runs it via `orch.exec_on_head` against an already-live inference server, then locates and parses the resulting `results*.json`.
- **Lifecycle stage** (`inference_suite_lifecycle.py::test_accuracy_eval`): opt-in pytest stage, parametrized one node per configured accuracy task (`test_accuracy_eval[<task_id>]`) so each task gets its own report row instead of one collapsed row for all tasks. Skips cleanly if a prior stage failed or no accuracy tasks are configured. Records per-task results via `lifecycle.record` (detail visible in the per-test HTML). Gates each task independently against `threshold.json`'s `accuracy.<task_id>` block when `enforce_thresholds` is set — a threshold miss or run failure on one task does not skip or fail sibling tasks.
- **Suite wiring**: `accuracy: AccuracyConfig` field added to `VariantConfig` (vllm), defaulting to empty so existing configs are unaffected.

Not included in this PR: the `long_isl_osl_stress` custom lm-eval task (its gibberish/repetition scoring logic needs its own design pass — deferred).

Hardware-validated on the TensorWave 2N DeepSeek-R1 tp8pp2 cluster (10.32.80.122/10.32.80.123):
- musr task completed successfully end-to-end (config -> server launch -> lm_eval run -> results parsed -> report bundle), both before and after the per-task parametrization change.
- Latest validation (2026-07-28) confirms the parametrized report: `test_accuracy_eval[musr]` renders as its own row (Passed, 2m43s) with per-subtask musr metrics in the linked detail HTML, rather than the earlier single collapsed `test_accuracy_eval` row.

## Test plan
- [x] `./.cvs_venv/bin/python -m pytest cvs/lib/inference/unittests/ --cluster_file /dev/null --config_file /dev/null` — 249 passed (2 pre-existing, unrelated failures in `test_inferencing_config_loader.py` persist with or without this branch)
- [x] Verified vllm suite module and conftest.py collection ordering import cleanly with `test_accuracy_eval` inserted
- [x] Hardware-validated on TensorWave 2N DeepSeek-R1 tp8pp2 (10.32.80.122/10.32.80.123), musr accuracy task, record-only mode


solaiys:
## Review (head `66e57af0`, base `dev/dtni`)

Regular review — blocking issues only.

### Blocking issues

What I checked:

- **Command construction / injection (`lm_eval_job.py`)** — every `lm_eval` arg is `shlex.quote`d, and `model_args`/`gen_kwargs` are joined into a single quoted token; the `find`/`cat` follow-ups quote `task_out_dir` and `result_path`. No shell-injection or interpolation gaps. `exec_on_head` single-result unpacking is guarded (`(x,) = out.values()` wrapped with a clear `RuntimeError`), and a missing `results*.json` is treated as a run failure.
- **Result parsing (`lm_eval_parsing.py`)** — pure function; skips `alias`/non-real numbers (bool/NaN excluded), flattens numeric metrics only. No crash paths.
- **Config schema (`accuracy_config.py`)** — Pydantic v2 `_Forbid`; mutable defaults (`{}`/`[]`) are per-instance-copied by Pydantic (not a shared-mutable-state bug), and duplicate task ids are rejected. The new `accuracy` field defaults to empty via `default_factory`, so existing configs are unaffected.
- **Threshold refactor (`vllm_config_loader.py` → shared `validate_thresholds_cover_sweep`)** — no regression: `import warnings` was removed from `vllm_config_loader.py` and it has no remaining `warnings.` use, while the shared function imports/uses `warnings`. New `gated_gpu_metrics` param defaults to `None` (backward-compatible for other callers); vllm passes `GATED_GPU_METRICS` to preserve prior gpu gating. `accuracy` is excluded from sweep-cell coverage via `NON_SWEEP_THRESHOLD_KEYS`.
- **Lifecycle stage (`test_accuracy_eval`)** — skips on prior failure / missing task, records per-task independently (doesn't set the shared `lifecycle.failed`), and only gates via `evaluate_all` when `enforce_thresholds` is set (record-only otherwise). vllm collection order/parametrization (`pytest_generate_tests` empty-list auto-skip, conftest rank) is consistent.
- **Concurrency** — no new threads/async; per-task output dirs are isolated (`output_dir/<task.id>`), so no shared-file races in normal (non-xdist) collection.

### Non-blocking note (not a merge blocker, but worth confirming)

The PR description states the `accuracy` field was added to `InferenceXAtomVariantConfig` and `test_accuracy_eval` was "imported into **both** suites' test modules." The diff only wires the **vllm** suite — the `inferencex_atom` suite (`inferencex_atom_single.py` / `_shared.py` / `conftest.py`) has no accuracy references, and no `inferencex_atom` config/suite file is changed. This isn't a bug in the code that's present (nothing breaks; accuracy is simply unavailable for `inferencex_atom`), but the described scope for `inferencex_atom` isn't delivered here — worth confirming whether that was intended for this PR or a follow-up.

_Reviewed 1 PR: 0 blocking issues — clean (with one scope-accuracy note)._

cvs/lib/inference/utils/vllm_config_loader.py
amd-droy, line 1: Please place all vLLM-related files under cvs\lib\inference\vllm. This will ensure that any files specific to the vLLM suite are organized in a single location, making them easier to identify and maintain.

cvs/lib/inference/utils/inference_suite_lifecycle.py
```diff
@@ -169,6 +172,60 @@ def test_model_fetch(orch, variant_config, lifecycle, request):
...
+    output_dir = f"{variant_config.paths.log_dir}/accuracy"
+
+    t = time.monotonic()
+    try:
+        actuals_by_id = run_accuracy_tasks(
+            orch=orch,
+            tasks=[task],
+            base_url=f"{params.base_url}:{params.port_no}",
+            model_id=variant_config.model.id,
+            model_path=variant_config.model.id,
```
amd-droy, line 211: Both model_id and model_path are set to variant_config.model.id. vLLM’s serve/fetch path sets HF_HUB_CACHE={paths.models_dir}; the lm-eval command does not export that env var. This worked on TensorWave, but on a fresh head node lm-eval’s tokenizer_backend=huggingface may fail to resolve the tokenizer.

cvs/lib/inference/unittests/test_accuracy_config.py
```diff
@@ -0,0 +1,513 @@
...
+    if j != -1:
+        tail = tail[:j]
+    return tail
+
+
+class TestAccuracyTaskDefaults(unittest.TestCase):
+    """AC12, AC23, AC24: defaults, empty-string id, include_path (no I/O)."""
+
+    def test_defaults_on_minimal_task(self):
+        t = AccuracyTask(id="a", task="mmlu")
```
amd-droy: mmlu is not needed. And it consists of 57 subjects, so takes a long time to run. But its ok if you want to keep it for now.

cvs/lib/inference/utils/lm_eval_job.py
```diff
@@ -0,0 +1,136 @@
...
+    endpoint_path = "/v1/chat/completions" if task.apply_chat_template else "/v1/completions"
+
+    model_args = ",".join(
+        [
+            f"base_url={ctx.base_url}{endpoint_path}",
+            f"model={ctx.model_id}",
+            f"tokenizer={ctx.model_path}",
+            "tokenizer_backend=huggingface",
+            f"num_concurrent={task.num_concurrent}",
+            "max_retries=3",
```
amd-droy, line 45: Many models with custom tokenizer code will break without this - trust_remote_code=True.

cvs/lib/inference/utils/lm_eval_job.py
```diff
@@ -0,0 +1,136 @@
...
+    model_id: str,
+    model_path: str,
+    output_dir: str,
+) -> Dict[str, Dict[str, float]]:
+    ctx = LmEvalCtx(base_url=base_url, model_id=model_id, model_path=model_path, output_dir=output_dir)
+    results: Dict[str, Dict[str, float]] = {}
+
+    for task in tasks:
+        cmd = build_lm_eval_cmd(task, ctx)
+        out = orch.exec_on_head(cmd, timeout=PER_TASK_TIMEOUT_S)
```
amd-droy: run_accuracy_tasks treats lm-eval as successful if any results*.json exists under {output_dir}/{task.id}; it does not verify lm-eval’s exit status - exit code == 0.

amd-droy (approved):
Looks good to me. thank you. good to merge.

PR #269, opened 2026-07-23
Ashutom/anc interlock

ashutom:
Motivation

  CVS had no first-class support for ANC (AMD Node Check) as a cluster validation suite. This PR adds an end-to-end ANC interlock: install ANC on every node, run its CPU/GPU check groups over parallel-SSH, collect per-node logs, and
  surface results in the HTML report — with the install and execution phases cleanly separated so either can run on its own.

  It also hardens the surrounding execution path so ANC behaves correctly on real clusters: slow package downloads and long-but-healthy check groups no longer trip fixed timeouts, unreachable nodes fail loudly instead of silently
  passing, and multi-node log artifacts no longer collide.

  Technical Details

  - Suite structure (generated): ANC validation is split into anc_test_cpu / anc_test_gpu suites, each holding one test_<group> function per ANC group. Both files are generated from CPU_GROUPS / GPU_GROUPS in cvs/lib/anc_lib.py via
  build_tools/gen_anc_suites.py (make gen-anc-suites), so adding/removing a group is a one-line edit plus a regenerate. Shared install/verify/ldconfig/collect logic lives in anc_lib.py; shared fixtures in cvs/tests/anc/conftest.py.
  - Install / execution split: anc_installation runs install + ROCm ldconfig fix as a standalone pre-task; group suites call ensure_anc_ready (session-cached). Install dispatches by release-archive flavour (deb/rpm/tar),
  auto-detected from the URL, with an anc_version precheck (skip if already present) and post-verify. All flavours install to /opt/amdtools/anc.
  - Inactivity timeouts (replaces fixed caps): Group runs now use a per-line inactivity timeout (inactivity_timeout, default 900s) instead of a total wall-clock cap, so a group that keeps printing progress runs as long as it needs.
  Implemented in Pssh/MultiProcessPssh via a gevent per-line timer that re-arms on every output line; a heartbeat is emitted when ANC console output is suppressed so the channel never looks idle. Package download+install similarly
  uses a config-driven inactivity timeout (install_timeout, default 1800s) with a wget partial-size progress heartbeat, so a slow ~550MB content package no longer trips the timer while a genuine stall still fails fast.
  - Unreachable-node correctness: phdl.exec only returns reachable hosts, so install / ldconfig / verification previously judged pass/fail from exec output alone and silently passed on down nodes. Added _fail_unreachable_nodes
  (cross-checks the cluster-file node_dict), gated the version precheck on full node coverage, and made install/verify fail any expected-but-missing node.
  - Log collection & HTML report: Per-node log layout is anc_logs/<node>/<test_name>/<timestamp>; root-owned ANC logs are archived via sudo tar and chowned back to the runner. A unified file-derived <ip>_<hostkey> node label is now
  used for both the HTML report path and the collected-log folder, fixing a multi-node arcname collision that overwrote per-node evidence inside the report bundle. Adds COLLECT_HTML_REPORTS / ADD_ANC_LOGS_TO_HTML_REPORTS config
  flags.
  - Docs: cvs/tests/anc/README.md documents the full config-key reference; every key is also documented inline in the shipped anc_config.json.

  12 commits; ~+2,700 lines across anc_lib.py, the ANC suites/fixtures, the parallel-SSH layer, the suite generator, and docs. The anc_release_url in the shipped sample config is intentionally left as REPLACE_ME.


cvs/lib/anc_lib.py
```diff
@@ -0,0 +1,1315 @@
...
+        # real hang. The heartbeat reports the growing output file size, mirroring
+        # the install download-progress snippet.
+        remote_stdout = "/tmp/anc_run_$$.out"
+        cmd = (
+            f"cd '{ANC_DIR}' && "
+            f"( sudo ./anc.py -g {groups_arg} > '{remote_stdout}' 2>&1 ) & "
+            f"anc_pid=$! && "
+            f"while kill -0 $anc_pid 2>/dev/null; do "
+            f"sz=$(wc -c < '{remote_stdout}' 2>/dev/null || echo 0); "
+            f"echo \"ANC {test_name} running: ${{sz}} bytes of output\"; "
```
solaiys: ⚪ Non-blocking robustness: in `print_all_to_console=false` mode the heartbeat emits every interval as long as `anc_pid` is **alive**, independent of whether output is growing — and because `inactivity_timeout` disables the total `read_timeout`, a live-but-**hung** ANC won't trip the inactivity timer and can hang until an external harness timeout. (The default `print_all=true` path is fine: ANC's own silence trips the timer.)

Suggest gating the heartbeat on output-file **growth** so a stalled ANC actually goes silent and the inactivity timer fires as intended, e.g. only echo when `sz` increased since the last tick (track a `last` size and `echo` only when `sz -gt last`). That preserves "keeps printing while healthy" while restoring "a genuine stall fails fast" in suppressed mode too.

solaiys (commented):
Reviewed for blocking issues + design-flow consistency.

✅ **No blocking issues found.** Highlights I verified:
- **Concurrency (gevent inactivity timer) is correct** — `_iter_lines` wraps each `next()` in a fresh `gevent.Timeout(exception=Timeout)` re-armed per line, caught by the existing handler; no new OS threads/shared state (heartbeats are remote shell loops). `inactivity_timeout` is threaded consistently through `Pssh.exec` and both `MultiProcessPssh` paths and is backward-compatible (`None` default), so existing callers are unaffected.
- **Unreachable-node correctness** — `_fail_unreachable_nodes` cross-checks `node_dict` so a down node fails loudly instead of silently passing.
- **Log collection is safe** — `tarfile.extractall(…, filter="data")` blocks path traversal; the unified file-derived `_node_label_from_file` is used for both the HTML path and the log folder (multi-node arcname collision fix is genuine and consistent).
- **Injection surface** — remote values are single-quoted and validated via `_assert_shell_safe`; config is operator-controlled.

✅ **Design-flow consistency: consistent.** The suite follows the established preflight family (same `cluster_file`/`config_file`/`cluster_dict`/`config_dict`/`phdl` fixtures over `Pssh`+`node_dict`, same `fail_test`/`update_test_result`/`print_test_output` and placeholder resolvers, same `pytest_runtest_makereport` HTML-extras hookwrapper). The `list_plugin.py` pruning (drop `conftest.py`/`resources`/`__pycache__`) is a uniform net improvement, and the root-conftest ANC auto-HTML is cleanly gated on `"anc" in config_dict` (best-effort try/except) so it can't affect other suites.

⚪ Non-blocking notes:
1. Suppressed-output-mode hang gap — see inline comment on `run_anc_groups`.
2. Minor: empty placeholder test classes (`TestAncInstallCoreTasks`/`TestAncInstallPostTasks`); `skip_dirs={"resources", …}` assumes no existing suite uses a `resources/` dir for real tests.

solaiys:
Comments from ignatious:
1]
**In Html Report:** 
add link to  error.json file along with the anc_logs.tar.gz file.  
   **Example:**  test_cpu_all_errors.json  |  test_cpu_all_20260727-232438_anc_logs.tar.gz
                         test_cpu_content_check_errors.json |  test_cpu_content_check_20260727-233426_anc_logs.tar.gz

 we do no need the zip file link , next to Full log link   "Full Log   ANC logs: test_cpu_all (FAILED)" in each test row in html summary.
 

 **In Config file:**
 2]
"log_folder_path": "{home}/cvs_logs/anc_logs/<node>/<test_name>/<timestamp>",
Lets get the base path from the config file.  appending the path with runtime values should be handled inside the code. 

3]
"html_report_path": "{home}/cvs_logs/html_reports/<node>/<test_name>/<timestamp>"
do we need this path?  --html command line arg can be used for the html report path right?

cvs/lib/anc_lib.py
ashutom: Fixed in commit 71758977ef3a300a2ed3531e85577b9203b4dd56.

Root cause confirmed valid: 
     in print_all_to_console=false, the heartbeat echoed every interval while anc.py was alive, re-arming the per-line inactivity timer regardless of output growth; since inactivity_timeout disables the total read cap, a live-but-hung ANC never tripped it. 

Fix :  Now the heartbeat emits only when the output file has grown since the last tick (seeded last=-1 so the
  first tick still arms the timer through ANC startup). Healthy runs keep the channel active; a genuine stall goes silent and the inactivity timer fires — matching print_all=true behavior.

ashutom:
> Comments from ignatious: 1] **In Html Report:** add link to error.json file along with the anc_logs.tar.gz file. **Example:** test_cpu_all_errors.json | test_cpu_all_20260727-232438_anc_logs.tar.gz test_cpu_content_check_errors.json | test_cpu_content_check_20260727-233426_anc_logs.tar.gz
> 
> we do no need the zip file link , next to Full log link "Full Log ANC logs: test_cpu_all (FAILED)" in each test row in html summary.
> 
> **In Config file:** 2] "log_folder_path": "{home}/cvs_logs/anc_logs//<test_name>/", Lets get the base path from the config file. appending the path with runtime values should be handled inside the code.
> 
> 3] "html_report_path": "{home}/cvs_logs/html_reports//<test_name>/" do we need this path? --html command line arg can be used for the html report path right?



The latest HTML reports appears like this : 
image

The summary reports has the failed test with the list of notes it failed on. Below that would be ANC logs for each tests.
If all pass then summary report shall say all pass and nothing more
The links below has these :  Full logs , error.json for each node, then anc logs for each node. 


The cofig file is cleaned. CVS_HOME and HTML report path is done away with. The log path prefix is shortened.

solaiys:
## Re-review (head `71758977`) — prior finding resolved, no blocking issues

✅ **Heartbeat fix confirmed** (my earlier inline comment): seeding `last=-1` and echoing only when the output file has **grown** means a healthy run keeps the SSH channel active, while a live-but-hung ANC stops growing the file → the channel goes silent → the inactivity timer fires, matching `print_all_to_console=true` behavior. Thanks for the quick turnaround.

Also verified across the new commits:
- **Shared `report_plugins.py` is backward-compatible** — the new `add_html_to_report` kwargs (`dest_name`/`track_in_reports`/`test_name`/`passed`) all default, and `generate_reports_section` routes non-ANC artifacts (no `test_name`, no verdicts) through the original flat list. rccl/preflight/inference reports are unaffected.
- **Fail-fast config guard is wired** in the `config_dict` fixture (`find_replace_me_placeholders` for all ANC suites; `validate_anc_path_prefixes` for `anc_test*`), so an unset / `REPLACE_ME` `log_folder_path` fails in seconds instead of raising mid-run.
- **Per-node reporting** — each node gets its own verdict line, `errors.json` link (`errors_occured:`/`no error:`), and `<label>_<test>_<ts>_anc_logs.tar.gz`; `NodeResult` is threaded consistently, including a synthesized entry for unreachable nodes so their label still shows.
- **Concurrency clean** — no new threads/shared state; the per-node loop is sequential on the main thread; the heartbeat is a remote shell loop. Install uses `mktemp -d` staging with `trap 'rm -rf "$STAGE"' EXIT` cleanup.

### Non-blocking hardening note
In `_install_anc_tar`, the top-level-folder cleanup

```
for name in $tops; do sudo rm -rf "/opt/amdtools/$name"; done
```

derives `$name` from the archive listing and filters only `.`/empty (`grep -vE '^[.]?$'`), not `..` or absolute components. A corrupt/malicious release archive could traverse (`..` → `sudo rm -rf /opt`). It's low-risk given the release is already fully trusted (sudo-extracted and executed), but filtering `$name` to a safe basename (drop `..`, absolute, empty) would harden the `sudo rm -rf`.

### Suggestion — group the ANC library code under `cvs/lib/anc/`
`anc_lib.py` (~1,600 lines) currently sits directly in `cvs/lib/`. Consider giving ANC its own package, `cvs/lib/anc/`, the way inference suites already do (`cvs/lib/inference/sglang/`, `cvs/lib/inference/inferencex_atom/`). That keeps suite-specific code discoverable and would let the big `anc_lib.py` be split later (install / execution / log-collection / reporting) without cluttering the shared `cvs/lib/` namespace.

One clarification on `run_config_paths.py`: it reads ANC-adjacent here, but its actual contents are **generic** run-config/baseline-path helpers (`resolve_runner_results_base`, `resolve_baseline_csv_folder`, `parallel_tree_path`, `{use_internal_baseline_data_path}`) — not ANC-specific. So I'd suggest: move the genuinely ANC-only modules (`anc_lib.py`, and the ANC resource/suite helpers) under `cvs/lib/anc/`, but keep `run_config_paths.py` in shared `cvs/lib/` (it's reusable by any suite that resolves baseline/runner paths). If it's intended to be ANC-only, renaming it to reflect that (or moving it under `anc/`) would avoid the shared-vs-specific ambiguity.

All non-blocking / organizational — nothing here blocks merge.

solaiys (approved):
LGTM — approving. My earlier blocking-adjacent finding (the suppressed-mode heartbeat hang) is resolved in `71758977`: the heartbeat now echoes only when the output file grows, so a live-but-hung ANC goes silent and the inactivity timer fires as intended.

Also verified: the shared `report_plugins.py` changes are backward-compatible (non-ANC suites keep the original flat Reports list; new `add_html_to_report` kwargs all default), the fail-fast config guard is wired in the `config_dict` fixture, per-node verdicts/`errors.json`/log-tarball links are threaded consistently via `NodeResult`, concurrency is clean (no new threads; sequential per-node loop; remote-shell heartbeat), and install uses `mktemp -d` staging with `trap ... EXIT` cleanup.

No blocking issues. The two items I raised are **optional, non-blocking follow-ups** for a later PR if you wish:
1. Harden the `_install_anc_tar` cleanup (`sudo rm -rf "/opt/amdtools/$name"`) by filtering `$name` to a safe basename (drop `..`/absolute) — low-risk given the release is already trusted.
2. Consider grouping ANC-only libs under `cvs/lib/anc/` (mirroring `cvs/lib/inference/sglang/`), keeping the generic `run_config_paths.py` shared in `cvs/lib/` unless it's intended ANC-only.

Nice work on the interlock.

PR #270, opened 2026-07-23
Capture vllm server log into test_vllm_inference output

atnair-amd:
## Summary
`_check_early_failure()` only tails the vLLM server log during the
startup/readiness wait; nothing re-checks it afterward, so a
mid-benchmark server crash (e.g. `EngineDeadError`) never appears in
`test_vllm_inference`'s captured output.

## Change
- Added `VllmJob.dump_server_log()`, mirroring the existing
  `dump_client_log()` pattern but per-rank, skipping ray-backend
  headless workers (rank > 0), which never run `vllm serve`
- Wired into `test_vllm_inference`'s failure path, via
  `lifecycle.live_server_job` so the dump targets whichever job
  actually owns the running server (on the server-reuse path, an
  earlier cell's job -- not the current cell's, which never issued
  `build_server_cmd()`/`start_server()`)
- Out of scope: no success-path dump -- the server log is per-server,
  not per-cell, so dumping on every passing cell sharing a reused
  server would re-emit the same growing log repeatedly

## Tests
- New `TestDumpServerLog` (3 cases: full content, mp-backend dumps
  every rank, ray-backend skips worker ranks) in
  `test_vllm_job_server_reuse.py`
- Gate: `.test_venv/bin/python -m unittest discover -s cvs/lib/inference/unittests -p "test_vllm_job*.py"` (92 passed)
- Full `cvs/lib/inference/unittests` discovery: 205/207 passed (2
  pre-existing failures in `test_inferencing_config_loader.py`,
  unrelated, reproduced identically on unmodified `dev/dtni`)

## Hardware validation (2026-07-27, jumphost 66.51.103.62, nodes 10.32.80.122/123)
1. **DeepSeek R1 TP16/PP1, 1k1k-conc16**, image
   `rocm/ufb-private:vllm-0.25.1-ubuntu24.04-py3.14-nightlies-device-all-cdna-rocm7.15.0a20260721-752a3a504`
   -- `test_vllm_inference[1k1k-conc16]` FAILED for real (375/400
   requests failed, `EngineDeadError` cascade). Captured pytest output
   contained 432 `EngineDeadError`-related lines, including the full
   `EngineCore`/`APIServer` tracebacks and shutdown sequence --
   confirms `dump_server_log()` captures the crash end-to-end, not
   just startup output. (1 failed, 6 passed, 43 skipped, 2996.52s)
2. **DeepSeek R1 TP8/PP2, 1k1k-conc16**, image
   `rocm/ufb-private:vllm-0.23.0-ubuntu24.04-py3.14-prereleases-device-all-cdna-rocm7.14.0rc3-0fc695fc6`
   -- full pass, no crash (49 passed, 1 skipped, 3456.52s). Confirms
   TP8/PP2 is healthy on this image/hardware; this run doesn't
   exercise the crash-capture path since nothing failed, but rules
   out this config/image combo as also affected.

Jira: AIMVT-270

### DIFF
--- a/cvs/lib/inference/unittests/test_vllm_job_server_reuse.py
+++ b/cvs/lib/inference/unittests/test_vllm_job_server_reuse.py
@@ -60,6 +60,23 @@ class FakeOrchWithOutput:
         return {}
 
 
+class FakeOrchMultiHost:
+    """Two-host fake orch that records which hosts each exec() call targeted."""
+
+    hosts = ["10.0.0.1", "10.0.0.2"]
+
+    def __init__(self):
+        self.exec_calls = []  # list of (cmd, hosts) actually issued
+
+    def exec(self, cmd, hosts=None, detailed=False):
+        self.exec_calls.append((cmd, hosts))
+        host = hosts[0]
+        return {host: f"content for {host}"}
+
+    def exec_on_head(self, cmd, *a, **k):
+        return {}
+
+
 def _make_job_for_check(tail_output="", grep_exit=1):
     """Construct a VllmJob suitable for testing _check_early_failure."""
     variant = mock.MagicMock()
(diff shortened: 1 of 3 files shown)


atnair-amd:
JIRA AIMVT-270: https://amd-hub.atlassian.net/browse/AIMVT-270

solaiys:
## Review (head `f2c7ab62`, base `dev/dtni`)

Regular review — blocking issues only.

### Blocking issues

What I checked:

- **Attributes/methods are all real and consistent:** `shlex` is imported (L39); `orch.hosts`, `_is_ray_backend` (property), `nnodes`, and `_rank_log()` all exist and are used the same way elsewhere. `self.orch.exec(cmd, hosts=[host])` is the established per-host call (`_wait_ready`, `_check_early_failure`), so no `NameError`/signature mismatch.
- **Correctness of rank handling:** the skip guard `self._is_ray_backend and int(self.nnodes) > 1 and rank > 0` is identical to `_check_early_failure` (L382) — mp backend dumps every rank, ray multinode dumps only rank 0 (workers have no server log), single-node dumps rank 0. The 3 new `TestDumpServerLog` cases cover exactly these.
- **Reuse-path targeting is right:** `lifecycle.live_server_job` is set right after `build_server_cmd()` (before `start_server()`), so a start/readiness failure still dumps the correct owning job's log; on the reuse path it stays pointed at the earlier cell's job (the one that actually owns the running server). `getattr(lifecycle, "live_server_job", job)` safely falls back to the current job if never set.
- **Security:** `rank_log` is internal and `shlex.quote`d — no injection.
- **Concurrency:** no new threads/async; per-host `cat` is read-only; `live_server_job` is set/read within the sequential pytest node flow — no shared-state race.
- **Scope:** additive method + failure-path wiring only; no API break, and success-path is intentionally not dumped (documented rationale: per-server, not per-cell log).

One tiny non-blocking observation (not worth gating): in `vllm.py` the `dump_server_log()` call sits just before `raise` in the `except` block, so a hypothetical exception from the dump would surface instead of the original — but `orch.exec` on a `cat` doesn't raise on a missing/empty file, and the original is preserved via exception chaining, so it's negligible.

_Reviewed 1 PR: 0 blocking issues — clean._

PR #272, opened 2026-07-24
AIMVT-139: migrate remaining dmesg/journal scans to node-scraper

speriaswamy-amd:
## Summary
- Migrate `verify_dmesg_for_errors`, `full_journalctl_scan`, and `verify_driver_errors` to the node-scraper adapter behind `CVS_DMESG_PARSER` (builds on merged #221).
- Consolidate shared parsing into `_node_scraper_scan()` plus helpers: `cvs_dmesg_error_regex()` (legacy pattern extensions), `_parse_cvs_time()` (tz-aware time windows).
- Refactor `full_dmesg_scan` to reuse the shared helper and CVS pattern extensions.
- Add unit tests for time-range filtering, journalctl path, driver-only filtering, and helper shape.

## Test plan
- [ ] `python -m unittest cvs.lib.unittests.test_verify_lib.TestDmesgMigrations`
- [ ] `python -m unittest cvs.lib.unittests.test_verify_lib.TestFullDmesgScan`
- [ ] Confirm `CVS_DMESG_PARSER=legacy` still uses regex paths for all four scanners
- [ ] Confirm default node-scraper path uses ISO dmesg / `journalctl -k -o short-iso` collection commands

### DIFF
--- a/cvs/lib/verify_lib.py
+++ b/cvs/lib/verify_lib.py
@@ -83,16 +83,19 @@ def cvs_dmesg_error_regex():
 
 
 def _parse_cvs_time(time_str):
-    """Convert a `date +"%a %b %e %H:%M"` string to a tz-aware datetime.
+    """Convert a `date +"%a %b %e %H:%M"` or `date +"%a %b %e %H:%M:%S"` string
+    to a tz-aware datetime.
 
     Returns None if the string cannot be parsed. The year is assumed to be the
     current year and the timezone the local timezone (CVS already assumes the
-    cluster is NTP-synced with the head node).
+    cluster is NTP-synced with the head node). Seconds default to 0 when the
+    input string is minute-precision only, for backward compatibility with
+    callers that haven't been updated to capture seconds.
     """
     if not time_str:
         return None
     match = re.search(
-        r'([A-Za-z]{3})\s+([A-Za-z]{3})\s+(\d+)\s+(\d{1,2}):(\d{2})',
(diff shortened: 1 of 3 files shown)


cvs/lib/verify_lib.py
```diff
@@ -444,6 +549,17 @@ def full_journalctl_scan(phdl):
         not containing those keywords if broader scanning is desired.
     """
 
+    if use_node_scraper_dmesg():
+        # node-scraper path: scan the full kernel journal (ISO timestamps) with
+        # the node-scraper analyzer plus CVS's historical patterns, instead of
+        # the lossy egrep prefilter + per-line regex.
+        output_dict = phdl.exec('sudo journalctl -k -o short-iso')
```
cijohnson: can we use 
journalctl -k -o short-iso --since="$start"

instead of collecting full logs journalctl -k -o short-iso'?

cijohnson (changes requested):

cvs/lib/verify_lib.py
speriaswamy-amd: Good catch, thanks — added `full_journalctl_scan(phdl, start_time_dict=None)` so callers can supply a start time and it's passed to journalctl via `--since` (server-side filtering), instead of always collecting the entire kernel journal. Also updated `check_cluster_health.py` (the only caller) to pass its own start time through, and to capture that timestamp with an explicit `date` format so it parses reliably (bare `date` output is locale-dependent).

Pushed in 6db1698.

cijohnson (approved):

PR #273, opened 2026-07-24
AIMVT-257: Containerize CVS

speriaswamy-amd:
## Summary
- add a multi-stage Docker image for the CVS head-node CLI
- document secure configuration, SSH-key, and results bind mounts
- add a pull-request container build and CLI smoke check

## Validation
- `git diff --check`
- clean-environment `pip install .`
- `cvs --version`
- `cvs list`
- `cvs copy-config --list`
- `cvs run host_configs_cvs --collect-only` (13 tests collected)

The local Docker daemon is not accessible in this workspace, so the new GitHub Actions `container-smoke` job verifies the actual image build and runtime commands.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,11 @@
+.git
+.github
+.cvs_venv
+.doc_venv
+.ruff_venv
+.test_venv
+dist
+docs/_build
+*.egg-info
+**/__pycache__
+**/*.py[cod]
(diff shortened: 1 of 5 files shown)


mikegreen-amd:
Hi - is the goal for these to be published as public packages that can be pulled down via ghcr.io?

cijohnson (approved):

speriaswamy-amd:
> Hi - is the goal for these to be published as public packages that can be pulled down via ghcr.io?

Hi @mikegreen-amd, this is mostly an internal ask for + some customers find it easier to run containers over python executables. But this PR does offer us the foundation for publishing on ghcr eventually.

PR #274, opened 2026-07-24
feat(vllm): add Prometheus /metrics-derived latency metrics (queue/prefill p50/p95)

atnair-amd:
Jira: AIMVT-271

## Summary
- Scrapes vLLM's own `/metrics` Prometheus endpoint before/after each client run and diffs the queue-time/prefill-time histograms to isolate one sweep cell's observations from a reused server's cumulative counters.
- Interpolates `p50`/`p95` quantiles matching PromQL's `histogram_quantile()`, exposed as gated metrics: `prom.queue_time_p50_ms`, `prom.queue_time_p95_ms`, `prom.prefill_time_p50_ms`, `prom.prefill_time_p95_ms`.
- Gating follows the existing `GATED_GPU_METRICS` precedent — a separate `GATED_PROM_METRICS` set, kept out of `vllm_parsing.py`'s `client.*` tiering machinery (that partition is asserted exactly by a locked invariant test).
- One-shot, synchronous, main-thread scrape bracket (mirrors `capture_gpu_metrics`), avoiding the SSH-session race the GPU poller previously hit.

## Test plan
- [x] New pure-parser unit suite `test_vllm_server_metrics.py` (30 tests): bucket/sum/count parsing, before/after diffing, hand-computed interpolation math, end-to-end realistic scrape pairs, I/O-boundary scrape failure modes.
- [x] Extended `test_vllm_config_loader.py` with `GATED_PROM_METRICS` coverage tests (missing-metric-raises / warns / full-set assertions).
- [x] Adjacent-suite regression: `test_vllm_report_preset.py`, `test_vllm_job_server_reuse.py`, `test_vllm_job_ray_backend.py` all pass unchanged.
- [x] Full-repo `run_all_unittests.py`: 750 tests (34 new), same 2 failures + 3 errors as the unmodified base branch (pre-existing, in unrelated `test_inferencing_config_loader.py`/collection issues — confirmed via A/B stash comparison, not touched by this change).

### DIFF
--- a/cvs/lib/inference/unittests/test_vllm_config_loader.py
+++ b/cvs/lib/inference/unittests/test_vllm_config_loader.py
@@ -13,6 +13,7 @@ from pydantic import ValidationError
 
 from cvs.lib.inference.utils.vllm_config_loader import (
     GATED_GPU_METRICS,
+    GATED_PROM_METRICS,
     Run,
     SeqCombo,
     Sweep,
@@ -26,9 +27,10 @@ def _combo(name, isl="128", osl="2048"):
 
 
 def _full_gated_specs():
-    """A spec for every gated client.* and gpu.* metric -- the minimum that
-    satisfies coverage. Values are inert so the set passes without asserting
-    anything; these tests pin the coverage gate, not the numbers."""
+    """A spec for every gated client.*, gpu.*, and prom.* metric -- the
+    minimum that satisfies coverage. Values are inert so the set passes
+    without asserting anything; these tests pin the coverage gate, not the
+    numbers."""
     out = {}
     for m in GATED_METRICS:
         kind = "max_ms" if m.endswith("_ms") else "max" if m == "failed" else "min"
@@ -36,6 +38,8 @@ def _full_gated_specs():
     for m in GATED_GPU_METRICS:
         kind = "max" if m in ("peak_gpu_memory_mb", "model_load_memory_mb", "model_load_s") else "min"
         out[f"gpu.{m}"] = {"kind": kind, "value": 0 if kind == "min" else 1e12}
+    for m in GATED_PROM_METRICS:
+        out[f"prom.{m}"] = {"kind": "max_ms", "value": 1e12}
     return out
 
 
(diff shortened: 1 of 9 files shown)


solaiys:
## Review (head `14dbb8b5`, base `dev/dtni`)

Regular review — blocking issues only, plus two non-blocking notes worth confirming.

### Blocking issues

What I checked:

- **Parsing/math (`vllm_server_metrics.py`)** — pure, I/O-free; `parse_prometheus_text`/`to_prom_metrics` degrade to `{}`/all-`None` on bad input and never raise. `diff_histogram` clamps negatives (counter resets). Histogram counters are correctly diffed before/after per cell (server is reused across concurrency-only cells).
- **Scrape (`scrape_vllm_metrics`)** — single, synchronous, main-thread `exec_on_head` (`curl -sf`), wrapped in try/except → returns `None` on failure. No background thread/poller, so no SSH-session race.
- **No `UnboundLocalError` on the failure path** — `prom_before`/`prom_after` are both assigned inside the outer `try`; the success-only code that consumes them (`to_prom_metrics`) is only reached when the `try` completes, so a failing cell `raise`s before touching them.
- **Suite wiring** — `test_prom_metric` mirrors `test_gpu_metric` (skip on prior failure / missing cell / `None` value; gate only when `enforce_thresholds` and a spec exists). `pytest_generate_tests` prom branch mirrors the gpu branch; conftest rank and preset `row_card_test_names` updated consistently.
- **Concurrency** — no new threads/shared mutable state; scrapes are read-only point-in-time `curl`s.

### Non-blocking notes (worth confirming, not merge blockers)

1. **Backward-compat of the new gated family (most important):** adding `prom.*` to `GATED_PROM_METRICS` makes all four `prom.*` metrics **mandatory per gated cell** — `test_missing_prom_metric_raises_when_enforced` confirms that with `enforce_thresholds: true`, a threshold file missing any `prom.*` spec now raises `ValidationError` at config load. The two committed vllm configs are `enforce_thresholds: false` (warn-only), so nothing in-repo breaks, and this follows the exact precedent of `GATED_GPU_METRICS`. But any **downstream/user threshold.json with `enforce_thresholds: true`** will fail to load until it adds the 4 `prom.*` specs to every cell — worth calling out in migration notes so those files are updated in lockstep.

2. **`histogram_quantile` diverges from PromQL in the `+Inf` bucket** (`vllm_server_metrics.py`): when a quantile lands in the top (`+Inf`) bucket, the interpolation returns `inf` (→ `inf` ms), whereas PromQL returns the highest *finite* bucket bound. The docstring claims parity with PromQL's algorithm. Under the overload/`EngineDeadError` scenarios these metrics target, a p95 above the largest finite bucket would surface as `inf` ms (and, if gated, fail the cell / risk `Infinity` in strict-JSON serialization). Consider clamping to the highest finite bound like PromQL.

_Reviewed 1 PR: 0 blocking issues — clean (with 2 non-blocking notes)._

solaiys:
## Re-review update (now head `9ea82910`; previous review was at `14dbb8b5`)

Both non-blocking notes from my [earlier review](https://github.com/ROCm/cvs/pull/274#issuecomment-5101289448) are now **✅ resolved**:

1. **`histogram_quantile` `+Inf` divergence → fixed in `b61ef5bf`.** When a quantile lands in the top `+Inf` bucket, the code now clamps to the highest *finite* bound (`return bound if len(parsed) == 1 else prev_bound`), matching PromQL, instead of returning `inf` ms. The `+Inf`-only histogram still returns `+Inf` (also PromQL-correct). No more risk of `inf` p95 failing a cell or producing non-strict JSON.

2. **Mandatory `prom.*` gated coverage → fixed in `9926639c`.** `_check_thresholds_cover_sweep` now passes `gated_metrics=set()` (and no gpu/prom sets), so no `client.*`/`gpu.*`/`prom.*` metric is mandatory per cell — a threshold.json only needs an entry per cell and may gate just the metrics the operator cares about. This removes the migration burden I flagged: downstream `enforce_thresholds: true` files no longer fail to load for missing `prom.*` specs. The unit tests were updated in lockstep (`test_missing_*_metric_does_not_raise_when_enforced`), and evaluation still gates per-metric where a spec exists.

### Still no blocking issues
Re-verified on the current head: the scrape remains single/synchronous/main-thread (no race), the interpolation math matches PromQL, and everything degrades to `None`/all-`None` without raising.

### One minor new note (non-blocking, optional)
The code/comments reference `VLLM_PROMETHEUS_METRICS_SPEC.md` ~10× (Sec 3.1, 3.2, 5, 6, 8.3, …), but that file isn't in the repo/PR. Either add the spec doc or trim the references so a reader isn't pointed at a missing file.

_Updated: previously 0 blocking + 2 non-blocking → now 0 blocking, both prior notes resolved (1 minor doc-reference note remaining)._

solaiys (approved):
LGTM — approving. Both non-blocking notes from my earlier review are resolved on this head:

- ✅ `histogram_quantile` now clamps to the highest finite bound in the `+Inf` case (`b61ef5bf`), matching PromQL.
- ✅ Threshold coverage no longer forces every gated `client.*`/`gpu.*`/`prom.*` metric per cell (`9926639c`), so downstream `enforce_thresholds: true` configs won't break; tests were updated in lockstep.

Re-verified on the current head: the `/metrics` scrape is single/synchronous/main-thread (no SSH-session race), the interpolation math matches PromQL, and everything degrades to `None`/all-`None` without raising. No blocking issues.

One small suggestion, ideally **before merge**: please remove (or add) the `VLLM_PROMETHEUS_METRICS_SPEC.md` references — the code/comments cite it ~10× (Sec 3.1, 3.2, 5, 6, 8.3, …) but the file isn't in the repo/PR, so readers get pointed at a missing doc. Non-blocking, but would be great to clean up so the references don't dangle. Nice work on the Prometheus metrics.

PR #275, opened 2026-07-27
SGLANG - orchestration and reporting changes

amd-droy:
Tests added:
sglang\sglang_single.py - Runs for single node, n=1.
sglang\sglang_distributed.py - Runs for multinode, n=1,2,3....
sglang\sglang_disagg_distributed.py - Runs in prefill/decode disaggregated mode, n=2,4,6...(n/2 nodes for prefill and decode)

Configs:
Main config - config_file\inference\sglang\mi30x_sglang_distributed.json - same for all models.
Threshold files, one for each model -
config_file\inference\sglang\mi30x_sglang_llama_70b_threshold.json
config_file\inference\sglang\mi30x_sglang_deepseek_r1_0528_threshold.json
config_file\inference\sglang\mi30x_sglang_glm_52_fp8_threshold.json
config_file\inference\sglang\mi30x_sglang_gpt_oss_120b_threshold.json
config_file\inference\sglang\mi30x_sglang_kimi_k26_threshold.json

Reporting: Runs for all three, single, distributed and disaggregated. All reports attached.
[sglang_disagg.zip](https://github.com/user-attachments/files/30470225/sglang_disagg.zip)
[sglang_distr_TP8_PP2.zip](https://github.com/user-attachments/files/30470226/sglang_distr_TP8_PP2.zip)
[sglang_distr_TP16_PP1.zip](https://github.com/user-attachments/files/30470228/sglang_distr_TP16_PP1.zip)
[sglang_single.zip](https://github.com/user-attachments/files/30470233/sglang_single.zip)


atnair-amd (commented):
## Code-flow review notes

I traced the entry points, the three lib modules (`sglang_single_lib.py`/`sglang_distributed_lib.py`/`sglang_disagg_lib.py`), the shared config/parsing helpers, and the `cvs/core` dependency (new `ContainerOrchestrator.cleanup_log_dir()`). Six items below, verified independently against the diff on `dr/sgl_reports` vs `dev/dtni`.

---

### 1. Bug: two of three report presets define a custom run-card builder but never wire it in

`cvs/lib/report/presets/sglang_single.py:41` defines `_sglang_single_run_card` and `cvs/lib/report/presets/sglang_disagg_distributed.py:51` defines `_sglang_disagg_run_card`, but neither is ever passed as `run_card_display_builder=` into their respective `make_inference_report_config(...)` calls. Since `make_inference_report_config` (`builder.py:126`) defaults to the generic `_default_run_card` when the kwarg is omitted, both presets silently render with generic Model/GPU/Framework/TP/Thresholds fields instead of the SGLang-specific fields (benchmark node, prefill/decode/proxy node lists) their custom builders were written for — most glaring for the disagg case, where prefill/decode node visibility is arguably the most important thing to show in a report.

For comparison, `sglang_distributed.py:97` does this correctly (`run_card_display_builder=_sglang_distributed_run_card`), and `inferencex_atom.py:28,68` follows the same correct pattern elsewhere in the codebase. No exception is raised — this fails silently.

**Fix:** add `run_card_display_builder=_sglang_single_run_card` / `run_card_display_builder=_sglang_disagg_run_card` to the two config calls.

---

### 2. Bug: `kimi-k2.6` threshold_file path doesn't match the file on disk

In `cvs/input/config_file/inference/sglang/mi30x_sglang_distributed.json` (~line 136), the `kimi-k2.6` variant references:
```
"threshold_file": "cvs/input/config_file/inference/sglang/mi30x_sglang_kimi_k2.6_threshold.json"
```
but the file actually present is `mi30x_sglang_kimi_k26_threshold.json` (no dot). `sglang_config_loader.py`'s threshold resolution (`_threshold_file_path` → `_resolve_threshold_path` → `_load_thresholds_file`) does a literal path join/open with no fuzzy matching, so selecting this variant will raise `FileNotFoundError`. All 4 other variants (llama-70b, deepseek-r1, glm-52-fp8, gpt-oss-120b) correctly match their files — this is an isolated typo, not a pattern. Currently latent since nothing else in the repo references `kimi-k2.6` yet, but will break the first time someone runs it.

**Fix:** correct the filename to `mi30x_sglang_kimi_k26_threshold.json`.

---

### 3. Dead code: commented-out test in `sglang_single.py` calls a method that doesn't exist on `SglangSingle`

`cvs/tests/inference/sglang/sglang_single.py:92-114` has a commented-out `test_run_long_context_accuracy` calling `im_obj.run_long_context_niah_accuracy(...)`, but that method is only implemented on `SglangDisaggPD` (`sglang_disagg_lib.py:1358`) — not on `SglangSingle`. If uncommented as-is, this would `AttributeError` immediately. (The equivalent commented-out test in `sglang_disagg_distributed.py:107-129` is fine, since `SglangDisaggPD` does implement it.)

**Fix:** either remove this dead scaffolding from `sglang_single.py`, or leave a clear `# TODO: not implemented for single-node` note so it's not accidentally uncommented expecting it to work.

---

### 4. Design: `cleanup_log_dir()` shouldn't be a new `cvs/core` orchestrator method

`ContainerOrchestrator.cleanup_log_dir()` (`cvs/core/orchestrators/container.py:711-746`) is new shared-orchestrator surface for what is really just "run `rm -rf && mkdir -p && chown` over SSH." The orchestrator already exposes exactly what's needed for this — `orch.head`/`orch.all` (public `Pssh` handles) and `orch.hosts`/`orch.user` — and this same PR already uses `orch.head.exec(...)`/`orch.all.exec(...)` directly at multiple call sites (`sglang_disagg_lib.py:214,219,221`, `sglang_single_lib.py:154`, `sglang_distributed_lib.py:204,210,212`, `conftest.py:418,422,428`) rather than going through a dedicated core method.

Adding a one-off convenience method to the shared `ContainerOrchestrator` class for SGLang's log-cleanup needs expands core's API surface unnecessarily and sets a precedent for other suites to do the same for their own one-off needs. It also shipped with no unit test coverage, unlike sibling methods on the same class (`setup_containers`, `teardown_containers`, `_setup_sshd` all have dedicated tests in `test_container.py`).

**Suggestion:** drop `cleanup_log_dir()` from `cvs/core/orchestrators/container.py` and inline the `rm -rf`/`mkdir -p`/`chown` command construction as a small local helper in `sglang_common.py`, built on `orch.head.exec`/`orch.all.exec` — consistent with how the rest of this PR already talks to the orchestrator.

---

### 5. Cleanup: MMLU is dead scaffolding left over from the single-file → 3-lib split

The old monolithic `sglang_disagg_lib.py` had a working `lm_eval_mmlu` spec and a `run_lm_eval_mmlu_benchmark_test` method; neither survived the split. `LM_EVAL_SPECS` in `sglang_common.py:164-177` now only has `lm_eval_hellaswag`/`lm_eval_gsm8k`, but `sglang_config_loader.py:175` and `_shared.py:184-189` still reference `lm_eval_mmlu`/`accuracy_mmlu` as if it were live. These are safe `.get()`-guarded lookups today (no threshold file defines an mmlu key, so it's inert), but it's confusing leftover scaffolding that implies MMLU support exists when it doesn't.

**Suggestion:** either remove the mmlu references from both files, or restore `lm_eval_mmlu` support in `LM_EVAL_SPECS` + one of the three new libs if it was dropped unintentionally — worth confirming with the author which was intended.

---

### 6. Design guidance (non-blocking): significant duplication across the three SGLang libs — worth a shared abstraction

This isn't a request to rework this PR before merge — it's already HW-validated across all three modes and is a large, working change. Flagging concretely now so it's tracked as a deliberate follow-up rather than silently compounding as more SGLang modes get added.

**a. `sglang_distributed_gpu_counts()` and `sglang_disagg_gpu_counts()` share a byte-for-byte identical inner function.** Compare `sglang_distributed_lib.py:580-600` and `sglang_disagg_lib.py:1153-1173` — the nested `_count_per_node(out_dict)` closure (parse `amd-smi --json` output, unwrap `gpu_data`, threshold by `used_vram`) is copy-pasted verbatim, 21 lines, same `try/except (json.JSONDecodeError, AttributeError)` fallback and same `mem_threshold_mb=5000` default. The only real difference is what the outer method does with the per-node dict afterward: distributed sums one group, disagg sums two groups and adds them (`sglang_disagg_lib.py:1191`: `"total_occupied_gpus": occupied_prefill + occupied_decode`). A shared helper taking `{group_name: node_list}` and returning per-group + total counts would cover both cases — and would make it nearly free to add the same topology check to `SglangSingle`, which currently has none.

**b. `_SERVER_READY_RE` is a copy-pasted module-level regex literal in all three files** (`sglang_single_lib.py:42`, `sglang_distributed_lib.py:43`, `sglang_disagg_lib.py:44`) — same pattern defined independently three times. Belongs in `sglang_common.py` as one shared constant.

**c. Five near-identical `setup_*_container_env` functions** (`sglang_single_lib.py:183`, `sglang_disagg_lib.py:365/393/421/443`) all follow the identical shape: build an `export ...` env body, wrap in a heredoc via `shlex.quote`, chmod, exec, sleep. Only the output path and the specific exported variables differ — a single "write and source an env script with these key=value pairs at this path" primitive would replace all five.

**d. `launch_server`/`launch_prefill_servers`/`launch_decode_servers`** (`sglang_single_lib.py:207-232`, `sglang_distributed_lib.py:284`, `sglang_disagg_lib.py:508,583`) all build the same heredoc-write → chmod → mkdir-log-dir → nohup → redirect-to-log launch-script pattern, with only the CLI flags differing (disagg's `--disaggregation-mode`/`--disaggregation-ib-device`, distributed's `--nnodes`/`--node-rank`/`--dist-init-addr`).

**e. `poll_for_server_ready`/`poll_and_check_server_ready`** also duplicate the same grep-log-for-pattern-with-timeout loop across all three libs (e.g. `sglang_single_lib.py:234-256` vs `sglang_disagg_lib.py:828-859`).

**Suggested direction:** a `ServerRole`/`ServerGroup` abstraction — `{role_name, node_list, extra_cli_flags, env_vars, log_path}` — with one shared implementation for write-env-script, build-and-launch-script, poll-for-ready, and count-occupied-gpus. Under that model: `single` = one role, `distributed` = one role spanning N nodes, `disagg` = two roles (prefill, decode) + a router. This would also have structurally prevented finding #1 above, since run-card/report wiring driven by the role list would come from one code path instead of three hand-copied ones.

**Not requesting this now** given the size of this PR and that all three modes are already HW-validated — recommend a follow-up ticket, since the duplication (5 env-setup functions, 3 launch functions, 2 GPU-count implementations, 3 copies of the ready-regex) will only grow as more SGLang deployment modes are added.

amd-droy:
1. Bug: two of three report presets define a custom run-card builder but never wire it in - **Fixed**

2. Bug: kimi-k2.6 threshold_file path doesn't match the file on disk - **Fixed**

3. Dead code: commented-out test in sglang_single.py calls a method that doesn't exist on SglangSingle - **marked TODO**

4. Design: cleanup_log_dir() shouldn't be a new cvs/core orchestrator method - **Implemented and moved**

5. Cleanup: MMLU is dead scaffolding left over from the single-file → **Removed**

6. Design guidance (non-blocking): significant duplication across the three SGLang libs — **refactored for a and b.**

a. sglang_distributed_gpu_counts() and sglang_disagg_gpu_counts() share a byte-for-byte identical inner function. - **Refactored**

b. _SERVER_READY_RE is a copy-pasted module-level regex literal in all three files - **Refactored**

c. Five near-identical setup_*_container_env functions - **Will be refactored in next PR**

d. launch_server/launch_prefill_servers/launch_decode_servers - **Will be refactored in next PR**

e. poll_for_server_ready/poll_and_check_server_ready also duplicate the same grep-log-for-pattern-with-timeout loop - **Will be refactored in next PR**

atnair-amd:
LGTM. One small nitpick: remove the shlex import in the container.py file.

atnair-amd:
LGTM. Squash and Merge!

PR #278, opened 2026-07-28
Stop the vLLM suite from logging bulk command output three times over

atnair-amd:
Jira: [AIMVT-273](https://amd-hub.atlassian.net/browse/AIMVT-273)

## Problem

A single `test_vllm_inference` run produced a 491 MB / 7.5M-line `cvs.log`, of which
~6,100 lines were signal. Three independent causes:

1. Bulk-data reads echoed in full — the end-of-run `cat` of the GPU poller file
   (`cvs/lib/utils/gpu.py:405`, `:442`) was 99.1% of the log on its own.
2. Every remote command logged twice. `globals.py` binds handlers to the **root**
   logger, so pssh's `host_logger` — which upstream silences behind a `NullHandler`
   and CVS never enables — reached CVS's handlers by propagation. 3.7M duplicate pairs,
   affecting every command in every suite, not just vLLM.
3. vLLM log tails added a third copy of the same server-log lines.

`print_console=False` already existed on `Pssh.exec`, but was unreachable from test
code: `ContainerOrchestrator`, `BaremetalOrchestrator` and `DockerRuntime` all dropped
the kwarg on the way down.

## Change

- Thread `print_console` through the orchestrator and runtime layers to `Pssh`,
  defaulting to `True` so every existing caller is byte-for-byte unchanged.
- Drop pssh's `host_logger` records with a filter, keeping the `_process_output`
  copy — the one that honors `print_console`. A filter rather than
  `propagate = False`: CVS runs under pytest, and `_pytest.logging.catching_logs`
  attaches its capture handler to root **and** to every non-propagating logger, so
  clearing `propagate` makes pytest attach directly to `host_logger` and the
  duplicate survives. A filter drops the record before any handler is consulted.
- Opt the GPU-metrics and vLLM log-tail reads out of console echo. Return values are
  unchanged, so parsing, digests and `dump_*_log` artifacts still see full output.
- Where the suppressed echo was the *only* record of a failure, produce the diagnostic
  deliberately: the OpenAI smoke test dumps its server log, and `parse_results` quotes
  the artifact it could not parse.
- `ContainerOrchestrator.exec_on_head` gains the `detailed` parameter its baremetal
  counterpart already had — `build_mpi_cmd` passes it, which raised `TypeError` on the
  container path.

## Tests

- 11 new tests: kwarg forwarding at each layer, a default-stays-verbose guard, the
  `detailed` path, and three for the `host_logger` suppression — asserting no
  `pssh.host_logger` record is captured (once with a handler attached straight to the
  logger, once inside a real `catching_logs`), plus a guard that other loggers still
  are. The earlier test asserted on the `propagate` flag in a plain process, so it
  verified the mechanism intended rather than the outcome wanted, and passed against
  a fix that does not work under pytest.
- `run_all_unittests.py`: 929 tests, 4 failures — all 4 reproduce on untouched
  `dev/dtni` @ `3cc63c0f` (918 tests, same 4). Zero regressions.
- `test_cli.sh` passes. `ruff check` error set byte-identical to base (14 both sides).
- End-to-end through the real `Pssh._process_output` with only the network faked:
  504 → 4 logged lines, returned payload byte-identical.

## Hardware

Validated on 2x MI300 (`10.32.80.124` / `10.32.80.125`), DeepSeek R1, TP8/PP2, 1k/1k,
image `rocm/ufb-private:vllm-0.23.0-ubuntu24.04-py3.14-prereleases-device-all-cdna-rocm7.14.0rc3-0fc695fc6`
(registry digest `sha256:1fce0c8189f141c4d2ed9c56144bd660a9554c92d07db1e896a253a1f51d1ab0`,
identical to the baseline node's).

Workload unaffected: 400/400 requests succeeded, 0 failed, 790.52 tok/s total
throughput, mean TTFT 370 ms, mean TPOT 39.5 ms.

| | baseline | this PR |
|---|---|---|
| `cvs.log` size | 143.1 MB | **0.20 MB** |
| lines | 2,287,334 | **1,705** |
| pssh `host_logger` duplicate lines | 367+ | **0** |
| raw `amd-smi` JSON blobs | flood | **0** |

The duplicate-line count is measured with `grep -cP '\[\d+\.\d+\.\d+\.\d+\]\t'`,
which matches pssh's `host_logger.info("[%s]%s\t%s", ...)` format. That pattern finds
367-644,724 matches in every prior run on this cluster -- including the run carrying the
earlier `propagate = False` attempt -- and zero in the runs on this commit.

Signal is preserved: 132 `Host ==` banners, the full benchmark result block, 163
server-log and 470 client-log labeled lines. 27 adjacent-identical line pairs remain,
all benign -- 16 are the `#---#` separator rule, 9 are vLLM engine lines the server
itself repeats, 2 are legitimately repeated status messages.

> `make test` fails at `fmt-check` on this branch **and on untouched `dev/dtni`** — 22
> pre-existing unformatted files (sglang/preflight/parsers), zero overlap with this
> PR's 14. Since `ut` → `installtest` → `test-venv` runs `clean_test_venv` *before*
> `fmt-check`, a red fmt-check also leaves `.test_venv` dependency-less; the results
> above came from running the two stages directly.

### DIFF
--- a/cvs/core/orchestrators/container.py
+++ b/cvs/core/orchestrators/container.py
@@ -618,7 +618,7 @@ class ContainerOrchestrator(BaremetalOrchestrator):
         if not self.container_id:
             raise RuntimeError("No containers running. Call setup_containers() first.")
 
-        return self.runtime.exec(self.container_id, cmd, hosts, timeout, detailed, print_console)
+        return self.runtime.exec(self.container_id, cmd, hosts, timeout, detailed=detailed, print_console=print_console)
 
     def exec_cmd_list(self, cmd_list, timeout=None):
         """
@@ -659,7 +659,9 @@ class ContainerOrchestrator(BaremetalOrchestrator):
         Returns:
             Dictionary mapping head node to execution result
         """
-        return self.runtime.exec_on_head(self.container_id, cmd, timeout, detailed, print_console)
+        return self.runtime.exec_on_head(
+            self.container_id, cmd, timeout, detailed=detailed, print_console=print_console
+        )
 
     def distribute_using_mpi(
         self,
--- a/cvs/core/orchestrators/container.py
+++ b/cvs/core/orchestrators/container.py
@@ -596,7 +596,7 @@ class ContainerOrchestrator(BaremetalOrchestrator):
             container_name = f"{username}_{sanitized_image}"
         return container_name
 
-    def exec(self, cmd, hosts=None, timeout=None, detailed=False):
+    def exec(self, cmd, hosts=None, timeout=None, detailed=False, print_console=True):
         """
         Execute command in running containers.
 
@@ -605,6 +605,9 @@ class ContainerOrchestrator(BaremetalOrchestrator):
             hosts: Target hosts (if None, uses all hosts)
             timeout: Command timeout
             detailed: If True, return detailed execution info including exit_code
+            print_console: If False, the command's output is returned but not
+                logged. Use for bulk data the caller parses itself — a single
+                unfiltered dump can otherwise reach hundreds of MB.
 
         Returns:
             Dictionary mapping hosts to execution results
@@ -615,7 +618,7 @@ class ContainerOrchestrator(BaremetalOrchestrator):
         if not self.container_id:
             raise RuntimeError("No containers running. Call setup_containers() first.")
 
-        return self.runtime.exec(self.container_id, cmd, hosts, timeout, detailed)
+        return self.runtime.exec(self.container_id, cmd, hosts, timeout, detailed, print_console)
 
     def exec_cmd_list(self, cmd_list, timeout=None):
         """
@@ -640,7 +643,7 @@ class ContainerOrchestrator(BaremetalOrchestrator):
 
         return self.runtime.exec_cmd_list(self.container_id, cmd_list, timeout)
 
-    def exec_on_head(self, cmd, timeout=None, detailed=False):
+    def exec_on_head(self, cmd, timeout=None, detailed=False, print_console=True):
         """
         Execute command directly on head node (baremetal).
 
@@ -648,11 +651,13 @@ class ContainerOrchestrator(BaremetalOrchestrator):
             cmd: Command to execute on head node
             timeout: Command timeout
             detailed: If True, return detailed execution info including exit_code
+            print_console: If False, the command's output is returned but not
+                logged. See exec().
 
         Returns:
             Dictionary mapping head node to execution result
         """
-        return self.runtime.exec_on_head(self.container_id, cmd, timeout, detailed)
+        return self.runtime.exec_on_head(self.container_id, cmd, timeout, detailed, print_console)
 
     def distribute_using_mpi(
         self,
--- a/cvs/core/orchestrators/container.py
+++ b/cvs/core/orchestrators/container.py
@@ -650,7 +650,9 @@ class ContainerOrchestrator(BaremetalOrchestrator):
         Args:
             cmd: Command to execute on head node
             timeout: Command timeout
-            detailed: If True, return detailed execution info including exit_code
+            detailed: If True, return detailed execution info including
+                exit_code. Mirrors BaremetalOrchestrator.exec_on_head, whose
+                build_mpi_cmd path calls this with detailed=True.
             print_console: If False, the command's output is returned but not
                 logged. See exec().
 
--- a/cvs/core/orchestrators/unittests/test_baremetal.py
+++ b/cvs/core/orchestrators/unittests/test_baremetal.py
@@ -9,12 +9,10 @@ All code contained here is Property of Advanced Micro Devices, Inc.
 # and command-dispatch surface (exec, exec_on_head, cleanup, sudo_prefix) used by the
 # migrated rvs_cvs.py orch fixture. Mocks Pssh so tests run with no SSH.
 
-import inspect
 import unittest
 from unittest.mock import MagicMock, patch
 
 from cvs.core.orchestrators.factory import OrchestratorConfig
-from cvs.core.orchestrators.base import Orchestrator
 from cvs.core.orchestrators.baremetal import BaremetalOrchestrator
 
 
@@ -107,30 +105,6 @@ class TestBaremetalOrchestrator(unittest.TestCase):
         orch.head = MagicMock()
         orch.exec_on_head("cat /tmp/huge", print_console=False)
         self.assertIs(orch.head.exec.call_args.kwargs["print_console"], False)
-
-    def test_abc_signature_matches_implementation(self):
-        """The Orchestrator ABC must advertise the kwargs its subclasses accept.
-
-        Python's abstractmethod enforces method NAMES only, never parameter
-        lists, so a subclass may silently widen a signature. That is how
-        `detailed` and then `print_console` came to exist on every concrete
-        orchestrator while the ABC still documented neither: nothing raises,
-        the tests pass, and the gap is invisible until someone writes a new
-        backend from the ABC and omits the kwarg -- at which point the drop is
-        silent (the command still works, it just logs hundreds of MB again).
-        Pinning the ABC against the implementation is the only thing that makes
-        that class of drift loud.
-        """
-        for method in ("exec", "exec_on_head"):
-            with self.subTest(method=method):
-                abc_params = inspect.signature(getattr(Orchestrator, method)).parameters
-                impl_params = inspect.signature(getattr(BaremetalOrchestrator, method)).parameters
-                missing = [p for p in impl_params if p not in abc_params]
-                self.assertEqual(
-                    missing,
-                    [],
-                    f"Orchestrator.{method}() is missing {missing}, which BaremetalOrchestrator.{method}() accepts",
-                )
 
     @patch("cvs.core.orchestrators.baremetal.Pssh")
     def test_cleanup_returns_true(self, _mock_pssh):
--- a/cvs/core/orchestrators/unittests/test_baremetal.py
+++ b/cvs/core/orchestrators/unittests/test_baremetal.py
@@ -9,10 +9,12 @@ All code contained here is Property of Advanced Micro Devices, Inc.
 # and command-dispatch surface (exec, exec_on_head, cleanup, sudo_prefix) used by the
 # migrated rvs_cvs.py orch fixture. Mocks Pssh so tests run with no SSH.
 
+import inspect
 import unittest
 from unittest.mock import MagicMock, patch
 
 from cvs.core.orchestrators.factory import OrchestratorConfig
+from cvs.core.orchestrators.base import Orchestrator
 from cvs.core.orchestrators.baremetal import BaremetalOrchestrator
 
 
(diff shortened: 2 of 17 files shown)


cvs/lib/globals.py
```diff
@@ -9,6 +9,21 @@
...
+# CVS's handlers anyway. The result is that every line is logged twice: once by
+# pssh, once by Pssh._process_output (cvs/lib/parallel/pssh.py). Dropping the
+# pssh copy keeps the _process_output one, which is the one that honors
+# print_console and can therefore be suppressed for bulk-data commands.
+#
+# A filter, not propagate=False: pytest's catching_logs attaches its capture
+# handler to root AND to every non-propagating logger (_pytest/logging.py), so
+# clearing propagate makes pytest attach directly and the duplicate survives.
+# A filter drops the record before any handler is consulted, however attached.
+logging.getLogger('pssh.host_logger').addFilter(lambda _record: False)
```
hnimra-amd: Nit: consider replacing the anonymous `lambda _record: False` with a named filter (e.g. `_suppress_pssh_host_logger(record) -> False`). Same behavior, but easier to spot in `logging.getLogger('pssh.host_logger').filters` when debugging log routing in the field.

cvs/lib/inference/vllm_job.py
```diff
@@ -672,6 +679,9 @@ def parse_results(self):
             try:
                 raw = json.loads(text)
             except (json.JSONDecodeError, ValueError) as e:
-                raise RuntimeError(f"unparseable results artifact on {host}: {artifact}: {e}") from e
+                # The artifact is read with print_console=False, so its content
+                # is nowhere else in the log -- carry a slice into the error or
+                # the failure is undiagnosable.
+                raise RuntimeError(f"unparseable results artifact on {host}: {artifact}: {e}: {text[:500]}") from e
```
hnimra-amd: Good call carrying a snippet into the error now that `print_console=False` hides the artifact from the log. One small tweak: raw `text[:500]` can inject newlines/control chars into CI/Jira exception text. Consider `repr(text[:500])` or collapsing whitespace to keep the failure one-line readable.

cvs/core/orchestrators/container.py
```diff
@@ -615,7 +618,7 @@ def exec(self, cmd, hosts=None, timeout=None, detailed=False):
         if not self.container_id:
             raise RuntimeError("No containers running. Call setup_containers() first.")
 
-        return self.runtime.exec(self.container_id, cmd, hosts, timeout, detailed)
+        return self.runtime.exec(self.container_id, cmd, hosts, timeout, detailed, print_console)
```
hnimra-amd: Nit: `detailed` and `print_console` are forwarded positionally here. Prefer keyword args (`detailed=detailed, print_console=print_console`) so a future signature change on `ContainerRuntime.exec()` is less likely to silently mis-route arguments.

cvs/core/orchestrators/container.py
```diff
@@ -615,7 +618,7 @@ def exec(self, cmd, hosts=None, timeout=None, detailed=False):
...
 
     def exec_cmd_list(self, cmd_list, timeout=None):
```
hnimra-amd, line 623: Follow-up (non-blocking): `exec()` now supports `print_console`, but `exec_cmd_list()` still always logs full output. Same footgun as the original bug if a bulk read ever goes through the per-host list path. Fine for this PR's scope, but worth threading `print_console` here in a follow-up for parity.

cvs/core/runtimes/docker.py
```diff
@@ -280,11 +283,11 @@ def exec_cmd_list(self, container_name, cmd_list, timeout=None):
         exec_cmd_list = [f"{sudo_prefix}docker exec {container_name} bash -c {shlex.quote(cmd)}" for cmd in cmd_list]
         return self.orchestrator.all.exec_cmd_list(exec_cmd_list, timeout=timeout)
```
hnimra-amd, line 284: Same `exec_cmd_list`/`print_console` parity note as on `ContainerOrchestrator.exec_cmd_list()` — not blocking for this PR, but the runtime layer is the other half of that gap.

cvs/core/orchestrators/unittests/test_baremetal.py
```diff
@@ -63,9 +63,49 @@ def test_exec_on_head_delegates_to_head_handle(self, _mock_pssh):
         orch.head = MagicMock()
         orch.head.exec.return_value = {"10.0.0.1": "ok"}
         result = orch.exec_on_head("hostname", timeout=10)
-        orch.head.exec.assert_called_once_with("hostname", timeout=10, detailed=False)
+        orch.head.exec.assert_called_once_with("hostname", timeout=10, detailed=False, print_console=True)
         self.assertEqual(result, {"10.0.0.1": "ok"})
 
+    @patch("cvs.core.orchestrators.baremetal.Pssh")
+    def test_exec_forwards_print_console_false_to_all(self, _mock_pssh):
```
hnimra-amd, line 70: These baremetal forwarding pins are great. Consider mirroring them in `test_container.py` for the container orchestrator layer too — that's the path vLLM actually uses, and it previously dropped both `print_console` and (for a while) `detailed` on `exec_on_head`.

cvs/tests/inference/vllm/vllm.py
```diff
@@ -313,6 +313,11 @@ def test_openai_compatible_smoke(orch, variant_config, hf_token, lifecycle, requ
         summary = job.probe_openai_endpoints()
     except Exception:
         lifecycle.failed = True
+        # This job owns its own short-lived server, so unlike the sweep there
+        # is no reuse path to consult -- dump ours. Needed because the poll
+        # loop's tails are no longer echoed to the console, so without this a
+        # bringup timeout leaves nothing from the failing server in the log.
+        job.dump_server_log()
```
hnimra-amd, line 320: +1 — this closes the diagnostic gap created by suppressing poll tails. Without this, a smoke bringup timeout would leave almost no server output in `cvs.log`.

cvs/lib/unittests/test_globals.py
```diff
@@ -0,0 +1,84 @@
...
+        root.setLevel(logging.INFO)
+        try:
+            _emit_host_line()
+        finally:
+            root.setLevel(orig_level)
+            host_logger.removeHandler(handler)
+
+        self.assertEqual(handler.records, [])
+
+    def test_no_duplicate_captured_under_pytest_catching_logs(self):
```
hnimra-amd, line 54: +1 — testing through `_pytest.logging.catching_logs` is the right regression guard. The old `propagate=False` approach passed unit tests but failed under pytest; this test catches that class of bug.

cvs/lib/utils/gpu.py
```diff
@@ -202,13 +202,13 @@ def capture_gpu_metrics(orch, nodes=None, timeout_s=None) -> dict:
     all_entries = []
     if nodes is None:
         kwargs = {"timeout": timeout_s} if timeout_s is not None else {}
-        out = orch.exec_on_head("amd-smi metric --json", **kwargs)
+        out = orch.exec_on_head("amd-smi metric --json", print_console=False, **kwargs)
```
hnimra-amd, line 205: +1 — suppressing console echo on `amd-smi metric --json` and poller read-back is exactly where the 400+ MB log bloat was coming from. Return values unchanged, so parsing/reporting should be unaffected.

hnimra-amd (commented):
Nice work. Root cause analysis is clear (triple logging: pssh `host_logger` propagation, `_process_output`, and vLLM poll tails), and the fix is layered correctly: global duplicate suppression + `print_console` plumbing + targeted opt-outs + failure-path diagnostics. Tests pin the silent-kwarg-drop failure mode, which is the right thing to guard. A few small interface/consistency nits below; none block merge from my side.

**Comments on lines outside this diff (could not anchor inline):**

- **`cvs/core/orchestrators/base.py:34` (`exec`)**: `BaremetalOrchestrator`, `ContainerOrchestrator`, `ContainerRuntime`, and `DockerRuntime` all gained `print_console=True`, but the ABC still documents the old signature. Please add `print_console=True` (and keep `detailed=False` on `exec_on_head`) here so the interface matches implementations and IDE/type hints stay accurate.

- **`cvs/core/orchestrators/base.py:52` (`exec_on_head`)**: Same as the `exec()` comment above — please add `print_console=True` to this abstract signature/docstring so callers know the kwarg exists at the orchestrator layer.

- **`cvs/lib/inference/vllm_job.py:451` (`wait_ready` timeout)**: Non-blocking suggestion: on timeout this raises without any server-log dump. The vLLM test harness compensates via `dump_server_log()` in its outer `except`, but a direct `VllmJob` caller won't get server output unless early-failure regexes matched. Consider calling `self.dump_server_log()` here (or documenting that callers must dump on failure) so library use matches test behavior.

cvs/lib/globals.py
atnair-amd: Done in a83470c0 — now a module-level `_suppress_pssh_host_logger(_record)`. Added a test that reads `logging.getLogger('pssh.host_logger').filters` and asserts a filter with that `__name__` is present, so the suppression can't silently regress back to an anonymous callable.

cvs/lib/inference/vllm_job.py
atnair-amd: Done in a83470c0 — `{text[:500]!r}`. Went with `repr()` over whitespace-collapsing so the content stays recoverable rather than lossy. Covered by a new `test_unparseable_artifact_snippet_is_single_line`, which feeds an artifact containing `\n`, `\r` and `\t` and asserts no raw newline or carriage return reaches the message while the text is still readable.

cvs/core/orchestrators/container.py
atnair-amd: Done in a83470c0 for both `exec()` and `exec_on_head()`. The new `test_container.py` pins read the argument by keyword and fall back to position, so they keep holding either way and would fail loudly if a future signature change mis-routed it.

cvs/core/orchestrators/unittests/test_baremetal.py
atnair-amd: Added `TestContainerOrchestratorExecForwarding` in `test_container.py` — six pins covering `print_console` (explicit `False` and the `True` default) and `detailed` across both `exec()` and `exec_on_head()`.

Validated by mutation rather than assumed: the pins pass on unmodified code, so I deleted the forwarding args in `container.py` and confirmed 4 of the 6 fail with named-argument messages, then restored the file byte-identically. The helper calls `self.fail()` on a missing argument so a dropped kwarg reads as an assertion instead of an `IndexError`.

cvs/core/orchestrators/container.py
atnair-amd: Agreed, and thanks for pinning both halves. `exec_cmd_list` is a real parity gap — nothing in the vLLM path routes bulk reads through it today, which is why it didn't surface, but the footgun is identical. Leaving it out of this PR to keep the diff scoped to the observed 491MB regression; I'll track it as a follow-up.

cvs/core/runtimes/docker.py
atnair-amd: Same follow-up as on `ContainerOrchestrator.exec_cmd_list()` — tracking the runtime half alongside it, out of scope for this PR.

atnair-amd:
Replies to the two points in the review body that couldn't anchor inline.

**`base.py:34` / `base.py:52` — ABC signatures**

I'd like to push back on these, though say the word and I'll add them.

`BaremetalOrchestrator` is the ABC's only subclass; `ContainerOrchestrator` extends *it*, not `Orchestrator` directly, and nothing else imports the ABC. So the signature every call site actually resolves against is the concrete one, which already carries `print_console`. `abc` enforces method *names* only and never checks parameter lists, so widening the abstract signature buys no enforcement.

Worth noting the drift also pre-dates this PR: `detailed` was already missing from the abstract `exec()` on `origin/dev/dtni` before my first commit. Fixing it here would be unrelated churn in a file this PR otherwise doesn't touch. Happy to split it into its own change if you'd rather have the ABC accurate as documentation — that seems like the real argument for it.

**`vllm_job.py:451` — `wait_ready` timeout dump**

Good catch on the library-vs-harness asymmetry. Leaving it for now since every in-repo caller goes through the test harness's `dump_server_log()`, and I'd rather not change failure-path behavior without a hardware run to confirm the dump lands where expected. Tracking alongside the `exec_cmd_list` parity item.

atnair-amd:
### Correction to my reply above

The last line of my `vllm_job.py:451` reply — "Tracking alongside the `exec_cmd_list` parity item" — refers to a parity item that does not exist. I raised it as a finding on #290 and have since withdrawn it there ([#290 comment](https://github.com/ROCm/cvs/pull/290#issuecomment-5198054398)).

`exec_cmd_list` already accepts `print_console`, on both implementations, and has since [b8b2cf24](https://github.com/ROCm/cvs/commit/b8b2cf24) (2026-01-02):

- [`multiprocess_pssh.py#L249-L254`](https://github.com/ROCm/cvs/blob/d9c5179239b1974cf20db56d6b1d535eb8a5ee87/cvs/lib/parallel/multiprocess_pssh.py#L249-L254) — the class the suites actually resolve to, since `parallel_ssh_lib` aliases `MultiProcessPssh` as `Pssh` at [`parallel_ssh_lib.py#L34-L36`](https://github.com/ROCm/cvs/blob/d9c5179239b1974cf20db56d6b1d535eb8a5ee87/cvs/lib/parallel_ssh_lib.py#L34-L36)
- [`pssh.py#L354-L357`](https://github.com/ROCm/cvs/blob/d9c5179239b1974cf20db56d6b1d535eb8a5ee87/cvs/lib/parallel/pssh.py#L354-L357)

So there is no gap here and nothing left open from this PR on that front. The `wait_ready` timeout-dump point stands as written — still deferred pending a hardware run, on its own.

The `base.py:34` / `base.py:52` ABC pushback is unaffected.

PR #282, opened 2026-07-30
docs: add AGENTS.md for cross-tool AI agent

cijohnson:
compatibility

Add comprehensive AI agent instruction files to enable Cursor, GitHub Copilot and other AI coding tools to understand CVS development patterns.

Key additions:
- AGENTS.md: Complete CVS development guide covering orchestrator patterns, configuration management, error handling, custom HTML reporting, and mandatory unit testing requirements
- CLAUDE.md: Backward compatibility shim for Claude Code

## Submission Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,194 @@
+# CVS (Cluster Validation Suite)
+
+CVS is a collection of test suites that can help customers qualify AMD's ROCm cluster from a functionality and performance point of view and certify its readiness for production workloads.
+
+**Key Value Propositions:**
+- **Deterministic qualification**: Make AI infrastructure qualification more deterministic, repeatable and time bound through an automated validation process
+- **Parallel execution**: Offer ability to run single node burn-in, health tests in parallel across a cluster and collate test results
+- **Issue detection**: Quickly spot out configuration inconsistencies, hardware issues/degradation that can potentially impact AI workload performance
+- **Troubleshooting tools**: Provide troubleshooting and visibility tools to narrow down problems when failures, performance issues are encountered
+
+## Architecture
+
+- **Validation Framework**: PyTest-based customer-facing test suites under `cvs/tests/`
+- **Orchestrator Layer**: `orch` fixture from `cvs/tests/conftest.py` — primary execution abstraction (baremetal SSH or container backends via `cvs/core/orchestrators/`)
+- **Cluster Management**: JSON-configured cluster topology, SSH credentials, and node connectivity
+- **Test Suites**: health, rccl, platform, preflight, ibperf, training, inference, benchmark, and others — see `cvs list`
+- **Reporting**: HTML reports, structured logging, and real-time monitoring
+
+## Development Setup
+
+See [CONTRIBUTORS.md](CONTRIBUTORS.md) for complete development workflow, setup instructions, and quality checks.
+
+Quick start:
+```bash
+git clone https://github.com/ROCm/cvs.git
+cd cvs
+make install
+source .cvs_venv/bin/activate  # Activate CVS environment
+cvs --version  # Validate installation
+```
+
+Before declaring a change done, from the repo root:
+```bash
+make fmt-check   # ruff format --check
+make lint        # ruff check + pylint E1205/E1206 (logging arg counts)
+make ut          # sdist -> .test_venv -> run_all_unittests.py
+```
+
+`make test` = `make ut` plus `test_cli.sh`. `make build` = `fmt-check lint sdist`.
+
+## CVS Development Conventions
+
+### Test Structure
+- Test functions use `test_` prefix (PyTest requirement)
+- Customer test suites live under `cvs/tests/`; library code under `cvs/lib/`
+- Configuration files in `cvs/input/config_file/` (JSON-structured, explicit parameters). Sample configs mark mandatory user fields with `<changeme>` in value strings — users must replace every `<changeme>` before running tests; unresolved placeholders hard-exit at startup via `_resolve_placeholders_in_dict` (`cvs/lib/utils_lib.py`), reached by `resolve_test_config_placeholders`, which nearly every test module calls. The `<changeme>` validators in `cvs/parsers/schemas.py` cover only the aorta and pytorch-xdit configs.
+- Tests require `--cluster_file` and `--config_file` CLI args (wired via `cvs/conftest.py`; the `orch` fixture is in `cvs/tests/conftest.py`)
+
+### Orchestrator Patterns (Recommended)
+- Use the `orch` fixture for new test development — see `cvs/tests/conftest.py`
+- Execute cluster commands with `orch.exec(cmd, timeout=30)`
+- Supports baremetal SSH and container orchestration backends
+- Handle node failures gracefully; inspect per-node results in returned dicts
+
+### Legacy Parallel SSH Patterns
+- **Legacy**: Direct `phdl`/`shdl` usage — avoid in new code; many existing suites still use it
+- `phdl.exec(cmd)` and `phdl.exec_cmd_list(cmd_list)` remain in older test files
+
+### Error Handling & Logging
+- Use `fail_test(msg)` for test failures with descriptive context
+- Time-bounded dmesg scanning (current pattern in RCCL/ibperf suites):
+  ```python
+  start_time = handle.exec('date +"%a %b %e %H:%M:%S"')
+  # ... run test ...
+  end_time = handle.exec('date +"%a %b %e %H:%M:%S"')
+  verify_dmesg_for_errors(handle, start_time, end_time, till_end_flag=False)
+  ```
+  `handle` is `phdl` today in most suites; `orch` works if it exposes `.exec()`. Use `till_end_flag=False` to bound scans to the test window.
+- Dmesg parsing: `full_dmesg_scan` honors `CVS_DMESG_PARSER` (default `node-scraper`; set `legacy` for the regex path). Note that `verify_dmesg_for_errors` — the time-bounded scan shown above — always uses the `err_patterns_dict` regex path and ignores this variable.
+- Use `err_patterns_dict` in `cvs/lib/verify_lib.py` for failure patterns
+
+### Custom HTML Reports
+- Generate test-specific HTML reports (performance charts, detailed results) as separate files
+- Integrate with pytest HTML report using `request.config._html_report_manager.add_html_to_report()`:
+  ```python
+  # Generate custom report
+  html_file = f'/tmp/custom_report_{os.getpid()}.html'
+  # ... build HTML content ...
+
+  # Add to pytest report bundle with clickable link
+  copied_path = request.config._html_report_manager.add_html_to_report(
+      html_file, link_name="Custom Performance Report", request=request
+  )
+  ```
+- See `cvs/tests/rccl/rccl_perf.py`, `cvs/tests/preflight/preflight_checks.py` for examples
+
+### Configuration Management
+- Cluster files: node topology, SSH credentials, head-node settings (`cvs/input/cluster_file/`)
+- Config files: test-specific parameters (RDMA interfaces, GPU settings, thresholds)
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,183 @@
...
+See [CONTRIBUTORS.md](CONTRIBUTORS.md) for complete development workflow, setup instructions, and quality checks.
+
+Quick start:
+```bash
+git clone https://github.com/ROCm/cvs.git
+cd cvs
+make install
+source .cvs_venv/bin/activate  # Activate CVS environment
+cvs --version  # Validate installation
+```
```
atnair-amd, line 30: **Suggest adding the quality gates here.**

The doc names only `make install` and defers the rest to CONTRIBUTORS.md — whose own entry reads `make lint - Check code quality (linting only) (linting only)`, so the pylint step is invisible from both. An agent finishing a change gets no signal these exist.

Proposed addition after the quick-start block:

````markdown
Before declaring a change done, from the repo root:

```bash
make fmt-check   # ruff format --check
make lint        # ruff check + pylint E1205/E1206 (logging arg counts)
make ut          # sdist -> .test_venv -> run_all_unittests.py
```

`make test` = `make ut` plus `test_cli.sh`. `make build` = `fmt-check lint sdist`.
````

(`Makefile:87-101` for the pylint step, `Makefile:79-85` for `ut`/`test`.)

File: AGENTS.md [hunk 2]
```diff
@@ -0,0 +1,183 @@
...
+
+## CVS Development Conventions
+
+### Test Structure
+- Test functions use `test_` prefix (PyTest requirement)
+- Customer test suites live under `cvs/tests/`; library code under `cvs/lib/`
+- Configuration files in `cvs/input/config_file/` (JSON-structured, explicit parameters). Sample configs mark mandatory user fields with `<changeme>` in value strings — users must replace every `<changeme>` before running tests; unresolved placeholders fail validation at startup (see `cvs/parsers/schemas.py`).
```
atnair-amd: **`cvs/parsers/schemas.py` isn't the enforcement path here.**

Unresolved `<changeme>` is caught by `_resolve_placeholders_in_dict` in `cvs/lib/utils_lib.py` (`sys.exit(1)`), reached via `resolve_test_config_placeholders` — called by 43 of 44 test modules. `cvs/parsers/schemas.py` has four `<changeme>` validators covering five field names, and it's imported by only the aorta, pytorch-xdit and preflight modules.

Suggested replacement for the second half of the sentence:

```markdown
unresolved placeholders hard-exit at startup via `_resolve_placeholders_in_dict`
(`cvs/lib/utils_lib.py`), reached by `resolve_test_config_placeholders`, which
nearly every test module calls. The `<changeme>` validators in
`cvs/parsers/schemas.py` cover only the aorta and pytorch-xdit configs.
```

AGENTS.md
```diff
@@ -0,0 +1,183 @@
...
+- Tests require `--cluster_file` and `--config_file` CLI args (wired via `conftest.py`)
```
atnair-amd: **Two `conftest.py` files exist, and this doc references things from both.**

`--cluster_file` / `--config_file` are wired in `cvs/conftest.py:35,46`; the `orch` fixture cited on line 41 is in `cvs/tests/conftest.py:17`; `_html_report_manager` on line 65 is back in `cvs/conftest.py:27`. Worth disambiguating:

```markdown
- Tests require `--cluster_file` and `--config_file` CLI args (wired via
  `cvs/conftest.py`; the `orch` fixture is in `cvs/tests/conftest.py`)
```

AGENTS.md
```diff
@@ -0,0 +1,183 @@
...
+- Use `fail_test(msg)` for test failures with descriptive context
+- Time-bounded dmesg scanning (current pattern in RCCL/ibperf suites):
+  ```python
+  start_time = handle.exec('date +"%a %b %e %H:%M:%S"')
+  # ... run test ...
+  end_time = handle.exec('date +"%a %b %e %H:%M:%S"')
+  verify_dmesg_for_errors(handle, start_time, end_time, till_end_flag=False)
+  ```
+  `handle` is `phdl` today in most suites; `orch` works if it exposes `.exec()`. Use `till_end_flag=False` to bound scans to the test window.
+- Dmesg parsing: default `CVS_DMESG_PARSER=node-scraper`; set `CVS_DMESG_PARSER=legacy` for regex path
```
atnair-amd: **This variable doesn't affect the function in the snippet above.**

`use_node_scraper_dmesg()` has exactly one call site — `cvs/lib/verify_lib.py:539`, inside `full_dmesg_scan`. `verify_dmesg_for_errors` (`verify_lib.py:214-298`) never consults it; it always walks `err_patterns_dict` / `warn_patterns_dict`. I ran the function under all three env states and the emitted command and behaviour were byte-identical.

Of the production call sites of the two functions, one reaches `full_dmesg_scan` (`cvs/monitors/check_cluster_health.py`) and the rest — all of rccl, ibperf, and the training/inference libs — reach `verify_dmesg_for_errors`. As placed, an agent debugging a missed RCCL dmesg error would set `CVS_DMESG_PARSER=legacy`, see no change, and rule out the parser.

Suggested replacement:

```markdown
- Dmesg parsing: `full_dmesg_scan` honors `CVS_DMESG_PARSER` (default
  `node-scraper`; set `legacy` for the regex path). Note that
  `verify_dmesg_for_errors` — the time-bounded scan shown above — always uses
  the `err_patterns_dict` regex path and ignores this variable.
```

AGENTS.md
```diff
@@ -0,0 +1,183 @@
...
+  # Generate custom report
+  html_file = f'/tmp/custom_report_{os.getpid()}.html'
+  # ... build HTML content ...
+
+  # Add to pytest report bundle with clickable link
+  copied_path = request.config._html_report_manager.add_html_to_report(
+      html_file, link_name="Custom Performance Report", request=request
+  )
+  ```
+- See `cvs/tests/rccl/rccl_perf.py`, `preflight_checks.py` for examples
```
atnair-amd: **Unqualified path.** `preflight_checks.py` is at `cvs/tests/preflight/preflight_checks.py` — its neighbour on the same line is fully qualified.

```markdown
- See `cvs/tests/rccl/rccl_perf.py`, `cvs/tests/preflight/preflight_checks.py` for examples
```

AGENTS.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Configuration Management
+- Cluster files: node topology, SSH credentials, head-node settings (`cvs/input/cluster_file/`)
+- Config files: test-specific parameters (RDMA interfaces, GPU settings, thresholds)
+- Use `<changeme>` in sample/default JSON values for any field the user must customize; do not ship cluster-specific values as defaults
```
atnair-amd, line 90: **`<changeme>` carries a second, conflicting meaning worth flagging here.**

Several suites use it as an *auto-detect sentinel* the user is meant to leave in place — `cvs/input/config_file/health/mi300_health_config.json:15` documents it as such ("Set the placeholder changeme to auto-detect from /opt/rocm or /opt/rocm/core-*"), and the guard is at `cvs/tests/health/install/install_rvs.py:36`. There are eight such comparisons across the tree.

As written, line 81 blesses only the "user must replace this" reading, so an agent will helpfully fill in a sentinel that was supposed to stay. Suggested addition below the existing bullet:

```markdown
- **Caveat:** some suites also use `<changeme>` as an *auto-detect sentinel* the
  user is meant to leave in place — see
  `cvs/input/config_file/health/mi300_health_config.json:15` and the guard at
  `cvs/tests/health/install/install_rvs.py:36`. The two meanings conflict; check
  which applies before filling one in or adding a new one.
```

Not this PR's to fix, but for context on why the ambiguity matters: the two meanings already collide at runtime on `main` — the shipped default health config hard-exits on its own `rocm_path` sentinel before the auto-detect code can run. I'll raise that separately.

AGENTS.md
```diff
@@ -0,0 +1,183 @@
...
+- **Performance impact**: Assess cluster scalability for changes affecting parallel operations
+- **Cross-platform testing**: Validate on Ubuntu (primary supported distro)
+- **Configuration compatibility**: Ensure changes maintain backward compatibility with existing cluster/config files
+
+## Code Quality Boundaries
+
+### Unit Testing Standards
+See [UNIT_TESTING_GUIDE.md](UNIT_TESTING_GUIDE.md) for comprehensive testing procedures.
+
+- **All library functions and classes** in `cvs/lib/` must have corresponding unit tests in `cvs/lib/unittests/`
```
atnair-amd: **Two corrections here: the location, and the scope.**

**Location** — tests live in each module's own `unittests/` dir, not one central `cvs/lib/unittests/`. There are nine such dirs, two of them inside `cvs/lib/` itself (`cvs/lib/parallel/unittests/`, `cvs/lib/preflight/unittests/`, ~134 tests between them). As written, an agent adding a `cvs/lib/preflight/` test files it in the wrong place.

**Scope** — the rule covers `cvs/lib/` only, but `cvs/runners/`, `cvs/parsers/`, `cvs/reports/` and `cvs/schema/` are library code with no `unittests/` dir at all. `cvs/parsers/schemas.py` alone is 1,122 lines and reads as out of scope under the current wording.

`cvs/core/orchestrators/` is the cleanest model in the repo — every concrete module paired with a same-named test beside it, ABC exempted:

```
baremetal.py  -> unittests/test_baremetal.py
container.py  -> unittests/test_container.py
factory.py    -> unittests/test_factory.py
base.py       -> (ABC, no test)
```

Suggested replacement for this bullet:

```markdown
- **All library functions and classes** must have corresponding unit tests in the
  tested module's own `unittests/` dir — e.g. `cvs/lib/preflight/unittests/`, not a
  central `cvs/lib/unittests/`. `run_all_unittests.py` discovers them recursively
  from the repo root.
- `cvs/core/orchestrators/` is the model: every concrete module has a same-named
  test beside it (`baremetal.py` → `unittests/test_baremetal.py`, and so on;
  `base.py` is an ABC and exempt).
- This applies to every library package, not just `cvs/lib/` — also `cvs/core/`,
  `cvs/runners/`, `cvs/parsers/`, `cvs/reports/`, `cvs/schema/`, `cvs/cli_plugins/`.
  The middle four have no `unittests/` dir yet; new code there should create one.
```

AGENTS.md
```diff
@@ -0,0 +1,183 @@
...
+- **Never add review feedback comments** ("changed per review", "moved to fix X") - code should stand alone
+- **Never apply band-aid fixes** - if a cleaner refactor improves maintainability, do the refactor
+
+### Unit Testing Standards
+- **Never write tests that simulate behavior** instead of calling actual code - tests must exercise real functions
+- **Never write environment-dependent unit tests** - use explicit fixtures for consistent results
+- **Never add timeouts to unit tests** - unit tests should be fast and deterministic (remote `orch.exec(..., timeout=...)` is fine)
+- **Never add external service dependencies** to unit tests - mock parallel SSH / orchestrator calls
+- **Never mix unittest and pytest patterns** - use `unittest.TestCase` for library unit tests only
+- **Never add library functions without unit tests** - all `cvs/lib/` code requires corresponding tests in `cvs/lib/unittests/`
```
atnair-amd: Same fix as the location note above — this restates the central-dir rule:

```markdown
- **Never add library functions without unit tests** - all library code requires
  corresponding tests in that module's own `unittests/` dir
```

atnair-amd (commented):
Reviewed this against the tree rather than reading it as prose — checked each cited path, symbol and env var, and exercised a few of the claims in a venv. The structure is right and most of it holds up; the notes below are the places where an agent following the doc would land in the wrong file or write something that doesn't match the code.

Five inline comments, each with the suggested replacement text. All are small.

AGENTS.md
cijohnson: AGENTS.md is instructions/rules  for the coding agents to follow, So from agent's perspective any config option that needs user input has to be tagged as <changeme>.

If a test suite decided to auto detect it and replace it, if user did not change it like in rvs install_rvs.py it is ok.

atnair-amd: sounds good, resolving this

atnair-amd (approved):
lgtm

PR #285, opened 2026-07-31
Urtiwari/tier2

urtiwari:
Added optional Tier 2 node_smoke support to preflight via config (tier2_perf: true).

When enabled, preflight passes --tier2-perf and thresholds for:

GEMM TFLOPS floor
HBM D2D bandwidth
local multi-GPU RCCL all-reduce
Updated: node_smoke.py, preflight config/schema/docs, report summary, and unit tests.

### DIFF
--- a/cvs/input/config_file/preflight/preflight_config.json
+++ b/cvs/input/config_file/preflight/preflight_config.json
@@ -155,7 +155,7 @@
       "venv_activate": "/home/{user-id}/envs/preflight/.venv/bin/activate",
       "_comment_venv_activate": "Path to the Python virtualenv activate script used by primus-cli direct. Required when connectivity_mode is 'run'.",
 
-      "connectivity_mode": "run",
+      "connectivity_mode": "skip",
       "_comment_connectivity_mode": "Options: 'run' (host/GPU/RDMA roll-call via node_smoke) or 'skip' (default).",
 
       "gpus_per_node": 8,
--- a/cvs/input/config_file/preflight/preflight_config.json
+++ b/cvs/input/config_file/preflight/preflight_config.json
@@ -155,7 +155,7 @@
       "venv_activate": "/home/{user-id}/envs/preflight/.venv/bin/activate",
       "_comment_venv_activate": "Path to the Python virtualenv activate script used by primus-cli direct. Required when connectivity_mode is 'run'.",
 
-      "connectivity_mode": "skip",
+      "connectivity_mode": "run",
       "_comment_connectivity_mode": "Options: 'run' (host/GPU/RDMA roll-call via node_smoke) or 'skip' (default).",
 
       "gpus_per_node": 8,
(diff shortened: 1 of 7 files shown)


cijohnson (changes requested):
I see 198 files being modified, added in this commit, is this expected?

urtiwari:
AMD General

No I accidentally pushed dev/dtni to main branch.
Ignore it.

cvs/input/config_file/preflight/preflight_config.json
```diff
@@ -155,7 +155,7 @@
       "venv_activate": "/home/{user-id}/envs/preflight/.venv/bin/activate",
       "_comment_venv_activate": "Path to the Python virtualenv activate script used by primus-cli direct. Required when connectivity_mode is 'run'.",
 
-      "connectivity_mode": "skip",
+      "connectivity_mode": "run",
```
solaiys: 🔴 **Blocking — this flips the shipped default from opt-in `skip` to `run`.** node_smoke was deliberately opt-in; setting it to `"run"` in the shipped config makes every default preflight run execute node_smoke — and with `auto_setup` (default true) + the default `primus_dir`, that clones Primus and installs a torch venv on every node before running the GPU/RDMA roll-call. It looks like an accidental test leftover: the `_comment_connectivity_mode` line right below still reads `"Options: 'run' … or 'skip' (default)"`, so the value now contradicts its own comment. Fix: revert to `"connectivity_mode": "skip"`. If turning it on by default is actually intended, that's a separate decision from "add Tier 2" — please update the comment and flag it explicitly.

solaiys (commented):
Reviewed for blocking issues. @cijohnson's earlier "198 files" concern is resolved — the current head is a single clean commit touching 7 focused files.

🔴 **Blocking — shipped default `connectivity_mode` flipped `skip` → `run`** (see inline comment). This PR is scoped to *optional* Tier 2, but it also turns node_smoke ON by default in the shipped preflight config, which (with `auto_setup` default true + the default `primus_dir`) would clone Primus + install a torch venv and run the GPU/RDMA roll-call on every node for anyone using the default config. It looks accidental — the adjacent `_comment_connectivity_mode` still says `'skip' (default)`, so the value contradicts its own doc. Please revert to `"skip"` unless enabling-by-default is intended (and if so, update the comment + call it out separately).

### Everything else looks good
- **`node_smoke.py`** — Tier 2 flags are numeric (`--gemm-tflops-min {float(...)}`, etc.), so no shell-injection surface; all gated on `tier2_perf` (pass `None` when off). `_effective_ssh_timeout()` bumps the per-node budget to `max(ssh_timeout, 600)` for the heavier run. No new threads/shared state — same per-reachable-host `exec_cmd_list` pattern, so no concurrency risk.
- **`schemas.py`** — new Tier 2 fields have sensible `ge` bounds and defaults that exactly match `_load_settings` (`600.0/2000.0/100.0/64/120`), so schema and code can't drift.
- **`report.py`** — Tier 2 summary/HTML use `(… or {}).get(k, '?')` and `html.escape(...)`, safe when `tier2_thresholds` is `None`. Minor cosmetic-only note (non-blocking): the Tier 2 HTML line is added only to the all-passed branch, not the failed-nodes branch.
- Config keys, schema, docs, and unit tests are consistent with each other.

cvs/input/config_file/preflight/preflight_config.json
urtiwari: Addressed the comment

solaiys:
I see, some places the /tmp/preflight path is used inside the code as a fallback path.
In general, this is not recommended. because the ownership of that folder will be fixed to current user. when the next user runs this test on the same node, it will throw permission error. 

But I also see this below path is used from the config file. So it is good for this PR.
      "artifacts_root_dir": "/tmp/{user-id}/preflight",

May be consider this to fix it in another PR. 
  - Update the default paths, fallback paths not getting created  at the common places on the node. If it is not cleared, the next user will face the permission error for that folder.

solaiys (approved):
LGTM.

PR #287, opened 2026-08-04
Dev/dtni retain legacy inferencemax

hnimra-amd:


solaiys:
## Review: blocking issues

Reviewed the diff plus the current `InferenceBaseJob` that the new `InferenceMaxJob` and pytest suite depend on. Two blocking issues before this can merge.

### 🔴 Threshold verification is silently dead — perf regressions won't fail the test
`cvs/input/config_file/inference/inferencemax/mi300x_inferencemax_gpt_oss_120b_single.json` (and the mi355x twin)

The base `verify_inference_results()` builds `expected_key = f"ISL={isl},OSL={osl},TP={tp},CONC={conc}"` and does `result_dict.get(expected_key, {})`, skipping all validation when that key is absent. The new configs ship a **flat** `result_dict` (`output_throughput_per_sec`, `mean_ttft_ms`, `mean_tpot_ms`) with no `ISL=...,OSL=...` key, so `expected_result_dict` is always `{}` → logs `"No expected results found ... skipping validation"` and `test_gpt_oss_120_single_node` passes unconditionally regardless of measured throughput/latency. A perf-gating suite that never gates is a correctness regression.

**Fix:** nest the thresholds under the composed key, e.g.
```json
"result_dict": {
    "ISL=8192,OSL=1024,TP=8,CONC=64": {
        "output_throughput_per_sec": "4200",
        "mean_ttft_ms": "500",
        "mean_tpot_ms": "15"
    }
}
```

### 🔴 `nnodes: "4"` contradicts the single-node suite and selects the wrong server script
`cvs/input/config_file/inference/inferencemax/mi300x_inferencemax_gpt_oss_120b_single.json:7` (and the mi355x twin)

The config is named `..._single.json`, run via the single-node pytest, and `test_gpt_oss_120_single_node` hardcodes `distributed_inference=False`. But with `nnodes="4"`, `InferenceMaxJob.get_server_script_path()` computes `base = "single_node" if int(self.nnodes) == 1 else "multi_node"` → returns `benchmarks/multi_node/gptoss_fp4_mi300x.sh` instead of the single-node script, and the base `launch_server()` / `build_server_inference_job_cmd()` / `launch_client()` loops all iterate `range(0, 4)`, creating `out-node0..3` dirs and launching 4 processes against a single-node cluster. This breaks the run on a genuine single-node setup.

**Fix:** set `"nnodes": "1"` in both single configs.

---

Concurrency scan: no new threads/async/shared clients introduced (`s_phdl`/`c_phdl` are independent handles, module tests run sequentially), so no race-condition findings. Non-blocking nits omitted.

cvs/input/config_file/inference/inferencemax/mi300x_inferencemax_gpt_oss_120b_single.json
```diff
@@ -0,0 +1,55 @@
...
+            "dataset_name": "random",
+            "max_concurrency": "64",
+            "model": "openai/gpt-oss-120b",
+            "num_prompts": "1000",
+            "input_sequence_length": "8192",
+            "output_sequence_length": "1024",
+            "burstiness": "1.0",
+            "seed": "0",
+            "max_model_length": "9216",
+            "random_range_ratio": "0.8",
```
atnair-amd, line 39: **Blocking — this value is silently clamped to `0` at runtime.**

The restored `InferenceMaxJob` subclasses the *live* `cvs/lib/inference/base.py`, which on `dev/dtni` was rewritten to clamp `--random-range-ratio` via `clamped_bench_random_range_ratio_str()`. For the values in this file:

```
r_cap = MML/(ISL+OSL) - 1 = 9216/(8192+1024) - 1 = 0.0
clamped_bench_random_range_ratio_str("0.8", "8192", "1024", "9216") -> ('0', True)
```

I verified the clamp is reachable: `InferenceMaxJob` does not override `launch_client`, so `start_inference_client_job()` resolves to the base implementation containing the clamp.

The clamp itself is not the bug — it's a necessary mitigation for the bench-driver swap in the same rewrite. The two drivers have opposite semantics:

| | driver | sampling | peak at r=0.8 |
|---|---|---|---|
| main | cloned `bench_serving` | lower-bound `[len*r, len]` | 8192+1024 = **9216 = MML** — fine |
| here | vLLM-shipped `benchmarks/` | symmetric `len*(1±r)` | **16588 > 9216** — mass rejection |

Since `max_model_length == ISL+OSL`, *any* `r > 0` overflows under the new driver, so `0` is the only feasible value. Removing the clamp would make it fail harder.

Net effect: main ran randomized lengths (input 6553–8192, output 819–1024); this PR runs fixed 8192/1024 — while `result_dict` thresholds below were tuned against the randomized distribution. It's only surfaced at `log.info`.

For a deprecation grace period this defeats the purpose: the "stable old way" runs a measurably different workload. Fix is to freeze the base (see the note on `inference_max.py`), keep the legacy driver, or raise `max_model_length` above `ISL+OSL`.

cvs/input/config_file/inference/inferencemax/mi300x_inferencemax_gpt_oss_120b_single.json
```diff
@@ -0,0 +1,55 @@
+{
+    "_comment": "Legacy InferenceMax config (deprecated). Migrate to inferencex_atom_single/. See inferencemax/README.md.",
+    "config": {
+        "container_image": "rocm/7.0:rocm7.0_ubuntu_22.04_vllm_0.10.1_instinct_20250927_rc1",
+        "container_name": "inference_max_rocm",
+        "_example_nnodes": "4",
+        "nnodes": "4",
+        "inferencemax_repo": "https://github.com/SemiAnalysisAI/InferenceX.git",
+        "benchmark_script_repo": "https://github.com/kimbochen/bench_serving.git",
```
atnair-amd, line 9: **Nit — Dead key.** `benchmark_script_repo` has zero readers at this commit — `grep -rn benchmark_script_repo --include='*.py' cvs/` returns nothing.

On main, `base.py:93` had a `setdefault` for it and `base.py:302` actually ran `git clone {self.if_dict['benchmark_script_repo']}`. On `dev/dtni`, `clone_bench_serving_repo()` (`base.py:327`) is a logging no-op — the client driver now resolves from the installed vLLM package's `benchmarks/`.

So this line reads as configuration but does nothing. Either drop it, or freeze the base so it works again. Same on the mi355x config (line 9), and `inferencemax.rst` still documents it at lines 50 and 126.

**Course of action depends on the freeze-vs-inherit decision above.** If the base is frozen, this key becomes live again and nothing here changes. If the live base is kept, it should be dropped here, on the mi355x config, and in `inferencemax.rst` (lines 50, 126). I'm deliberately not prescribing one — the key matches `main`, so the config isn't what diverged.

cvs/lib/inference/inference_max.py
```diff
@@ -0,0 +1,69 @@
+'''
+Copyright 2025 Advanced Micro Devices, Inc.
+All rights reserved. This notice is intended as a precaution against inadvertent publication and does not imply publication or any waiver of confidentiality.
+The year included in the foregoing notice is the year of creation of the work.
+All code contained here is Property of Advanced Micro Devices, Inc.
+
+.. deprecated::
+   Legacy InferenceMax host+docker job. New work should use
+   ``cvs.lib.inference.inferencex_atom_orch.InferenceXAtomJob`` and the
```
atnair-amd, line 9: **Nit — Wrong module path** — `cvs.lib.inference.inferencex_atom_orch` is not importable:

```
>>> import cvs.lib.inference.inferencex_atom_orch
ModuleNotFoundError: No module named 'cvs.lib.inference.inferencex_atom_orch'
```

`cvs/lib/inference/__init__.py` is empty, so no re-export rescues it. The real path is `cvs.lib.inference.inferencex_atom.inferencex_atom_orch` — which `inference_lib.py` (lines 15, 38) already uses correctly. This docstring is the only outlier in the repo.

**Suggested fix** (one-click):

```suggestion
   ``cvs.lib.inference.inferencex_atom.inferencex_atom_orch.InferenceXAtomJob`` and the
```

cvs/lib/inference/inference_max.py
```diff
@@ -0,0 +1,69 @@
...
+
+from cvs.lib.inference.base import InferenceBaseJob
+from cvs.lib.verify_lib import fail_test
+
+
+class InferenceMaxJob(InferenceBaseJob):
+    """Legacy InferenceMAX-specific implementation (deprecated)."""
+
+    def __init__(self, *args, **kwargs):
+        warnings.warn(
```
atnair-amd, line 26: **Nit — This warning never reaches the user.** `pytest.ini` line 3 sets `addopts = -p no:warnings`, which disables the warnings plugin entirely — so under `cvs run` (the only way this suite executes) the `DeprecationWarning` is swallowed.

The sole surviving signal is `log.warning(_LEGACY_INFERENCEMAX_MSG)` in the factory, which fires only via `InferenceJobFactory`. The pytest suite instantiates `InferenceMaxJob` directly (`inferencemax_gpt_oss_120b_single.py:282`), bypassing the factory — so a user running the legacy suite the documented way sees **no deprecation notice at all**.

Since announcing the deprecation is the point of the PR, consider a `log.warning` here alongside the `warnings.warn`.

**Two viable remedies, both reasonable — flagging rather than prescribing.** (a) Route the suite through `InferenceJobFactory`, which already emits `log.warning(_LEGACY_INFERENCEMAX_MSG)`; or (b) add a `log.warning` here. Note (b) means importing `globals`/`log` into this module — it currently imports only `re`, `time`, `warnings`, `InferenceBaseJob`, `fail_test` — so it adds a line `main` doesn't have.

cvs/lib/inference/inference_max.py
```diff
@@ -0,0 +1,69 @@
...
+   ``cvs.lib.inference.inferencex_atom_orch.InferenceXAtomJob`` and the
+   ``inferencex_atom_single`` suite. This module is retained for OSS users
+   migrating gradually.
+'''
+
+import re
+import time
+import warnings
+
+from cvs.lib.inference.base import InferenceBaseJob
```
atnair-amd, line 18: **Design decision — the root cause of the drift findings.**

This imports the *live* base, which has diverged from main by ~119/-37 lines. The restored suite therefore inherits four behavior changes it was written against:

1. `--random-range-ratio` now clamped (0.8 → 0 for the shipped config)
2. `benchmark_script_repo` dropped; bench driver swapped to vLLM-shipped `benchmarks/`
3. parsed key `'Total generated tokens:'` → `'total_generated_tokens'`
4. readiness polling tail+match → whole-file grep

Each is individually defensible in the new suite; together they mean the "stable old way" is not the old way. For a 3-month grace period whose purpose is bit-compatible behavior, that's the core tension.

There's a worked alternative on `atnair/vllm-legacy-deprecation`: a frozen `base_legacy.py` — a verbatim copy of main's 711-line `base.py` plus 25 lines (a docstring enumerating exactly these four drifts, and a `LEGACY_BENCH_DEFAULTS` constant a guard test asserts against). Legacy suites import the frozen copy; new suites keep the live base; both get deleted together when the window closes.

Worth an explicit decision — freeze or inherit — rather than inheriting by default.

atnair-amd (commented):
Reviewed against `origin/main` in a clean worktree at `abba2b8`. Short version: **the restoration is faithful in content but not in behavior.**

## What checks out

`cvs/lib/inference/inference_max.py` is byte-identical to `origin/main`'s copy except for the deprecation notes — I diffed the ASTs after stripping docstrings and the `warnings` import/call and they are equal. Same for the two configs and the pytest suite. The factory rewiring in `inference_lib.py` (`'inferencemax'` → `InferenceMaxJob`, replacing the `NotImplementedError` placeholder) is the one substantive change and it's correct.

I also separated PR-caused problems from the known-red `dev/dtni` baseline by building a pristine baseline worktree and comparing failure *identities* (rule + file + line + column), not counts. The 23 `fmt-check` failures, 14 lint failures, and 4 unit-test failures are all pre-existing; every file this PR touches is ruff-clean, and there are **zero test regressions**. The 1081-vs-1103 test-count gap is just the branch being 2 commits behind — I confirmed with a control build at merge-base `b3842e14`, which yields 1081 with an identical ID set.

One thing I checked and am explicitly *not* flagging: the composite-key threshold handling. It looked wrong, but the function is byte-identical to `main`, so it's pre-existing and out of scope here.

## The core issue

`InferenceMaxJob` subclasses the **live** `cvs/lib/inference/base.py`, which diverged from `main` by roughly +119/−37 lines. The restored suite therefore inherits four behavior changes it was never written against — clamped `--random-range-ratio`, a swapped bench driver, a renamed parsed metric key, and rewritten readiness polling.

The concrete consequence is the blocking comment on the mi300x config: `random_range_ratio: "0.8"` resolves to `0` at runtime, because `max_model_length == ISL+OSL` makes `r_cap` exactly `0.0`. `main` ran randomized sequence lengths; this PR runs fixed ones, against `result_dict` thresholds that were tuned on the randomized distribution. It's surfaced only at `log.info`.

For a 3-month grace period whose whole purpose is that the old way keeps behaving like the old way, that's the tension worth resolving before merge. Freeze-or-inherit is a real decision and I've left it as one rather than prescribing an answer — `atnair/vllm-legacy-deprecation` has a worked `base_legacy.py` alternative if it's useful as a reference.

## One item I couldn't anchor inline

`cvs/lib/inference_lib.py:27` falls between diff hunks, so GitHub won't take an inline comment. Inside `_LegacyVllmInferenceJobPlaceholder`, the `NotImplementedError` points users at:

```
cvs.lib.inference.vllm_single.VllmJob
```

That module does not exist. The real class is `VllmJob` at `cvs/lib/inference/vllm_job.py:78`, so the path should be `cvs.lib.inference.vllm_job.VllmJob`. Same class of pathing nit as the one on `inference_max.py:9` — worth fixing in the same pass, since this string is the only guidance a user hits at that failure point.

## Severity

One **Blocking** (the clamped ratio), one **Design decision** (freeze vs inherit), one **Scope** note (legacy vLLM/sglang suites are still deleted — if the 3-month rule covers them too, they need the same treatment), one **Suggestion** (a cheap unit test that would have caught the clamp), and five **Nits**, two of which have one-click `suggestion` blocks.

Nothing here is a request to change your approach — the restoration itself is well done, and per the review scope I've flagged divergence from `main` rather than rewriting anything.

cvs/tests/inference/inferencemax/inferencemax_gpt_oss_120b_single.py
```diff
@@ -0,0 +1,297 @@
...
+    out_dict = s_phdl.exec('docker ps')
+    for node in out_dict.keys():
+        if not re.search(f'{container_name}', out_dict[node], re.I):
+            fail_test(f'Failed to launch container on node {node}')
+    update_test_result()
+
+
+def test_gpt_oss_120_single_node(c_phdl, s_phdl, gpu_type, inference_dict, benchmark_params_dict, hf_token):
+    globals.error_list = []
+    im_obj = InferenceMaxJob(
```
atnair-amd, line 282: **Suggestion — No unit test covers this path.** The PR adds no test files, and nothing in the repo exercises `InferenceMaxJob` or `InferenceJobFactory`:

- `cvs/lib/unittests/test_inference_max.py` is a misnomer — despite the name it imports only `textwrap_for_yml` from `base.py` and never references `InferenceMaxJob`. (Pre-existing, not introduced here.)
- `grep -rln InferenceJobFactory --include='*.py' cvs/` returns only `inference_lib.py` itself — zero tests, even though flipping that registry entry is the substantive change in this PR.

This suite needs hardware, so it can't gate CI. But a cheap unit test would catch the drift above: construct with the shipped config against a fake handle and assert the emitted client command still carries `--random-range-ratio 0.8`. That test would fail today. The `vllm_legacy` branch does this with an 18-method guard test.

Related: the deprecation notice is invisible from this entry point — see the note on `inference_max.py:26`.

cvs/input/config_file/inference/inferencemax/README.md
```diff
@@ -0,0 +1,33 @@
+# InferenceMax configs (legacy, deprecated)
+
+**Status:** Legacy — retained for OSS users migrating gradually. Scheduled for removal after a deprecation window (target: post Aug 2026 main merge stabilization).
```
atnair-amd, line 3: **Nit — No concrete removal date.** AMD's rule is a 3-month grace period, but "post Aug 2026 main merge stabilization" names a *start* condition, not an end — and `inferencemax.rst:11` says only "after a deprecation window" with no date at all. I grepped all 11 changed files: no ISO date, month-year, or quarter anywhere.

Without a date nobody can tell when the window closes, and the removal PR has nothing to cite. Suggest a fixed date in both places, e.g. `2026-11-11` — which is what `atnair/vllm-legacy-deprecation` uses, stated in 8 files and asserted in a unit test so the window can't quietly slip.

docs/reference/configuration-files/inferencemax.rst
```diff
@@ -0,0 +1,218 @@
...
+
+   **Legacy (deprecated).** This page documents the monolithic ``config`` + ``benchmark_params``
+   InferenceMAX layout and the ``inferencemax_gpt_oss_120b_single`` pytest suite. New work should
+   use :doc:`InferenceX ATOM </reference/configuration-files/inferencex_atom>` and
+   ``cvs run inferencex_atom_single``. The legacy suite remains in the repo for OSS users
+   migrating gradually; it is scheduled for removal after a deprecation window.
+
+*******************************************
+InferenceMAX inference configuration file (legacy)
+*******************************************
```
atnair-amd, line 15: **Nit — Title overline too short.** The title text grew to 50 chars but the overline/underline are still 43:

```
line 13  len=43  *******************************************
line 14  len=50  InferenceMAX inference configuration file (legacy)
line 15  len=43  *******************************************
```

docutils reports `(WARNING/2) Title overline too short.` on this file; the same parse of `origin/main`'s version is clean, so it's introduced here. Cosmetic — no `-W` in `.readthedocs.yaml`, so the build won't fail. Just pad both rules to 50.

**Suggested fix** (one-click):

```suggestion
**************************************************
InferenceMAX inference configuration file (legacy)
**************************************************
```

cvs/input/config_file/inference/inferencex_atom_single/README.md
```diff
@@ -20,7 +20,10 @@ W1 **DeepSeek R1 FP8** on 8× GPU, ISL=OSL=1024, TP8.
 
 Each shipped config sets `"threshold_json"` to the sibling threshold filename (resolved relative to the config directory). You may also use an absolute path (vLLM-style).
 
-Legacy nested layouts (`deepseek_r1_fp8_mi300x_atom_perf/`, `inferencemax/`, etc.) are **removed** from the repo tree. Use only the flat stems below.
+Legacy nested layouts (`deepseek_r1_fp8_mi300x_atom_perf/`, etc.) are **not** used for new work.
```
atnair-amd, line 23: **Scope — legacy vLLM and sglang suites are still deleted.** This PR restores the InferenceMax lineage only. Relative to `origin/main`, these remain absent:

```
cvs/lib/inference/vllm.py
cvs/input/config_file/inference/vllm/mi355x_vllm_single.json
cvs/tests/inference/vllm/vllm_{deepseek31_685b,gpt_oss_120b,qwen3_235b,qwen3_80b}_single.py
cvs/tests/inference/sglang/sglang_{deepseek_r1_671b,llama_70b}_distributed.py
```

I checked these are genuine deletions, not renames: `git diff -M10% --find-copies-harder` classifies all 8 as `D`, and brute-force content matching against every file at head tops out at 42–56% — all pytest fixture boilerplate, no suite-specific logic. Head's unified `cvs/tests/inference/vllm/vllm.py` is a different harness (orchestrator + `schema_version 1` configs), and the 371-line legacy-shape `mi355x_vllm_single.json` those suites consumed is gone, so the legacy vLLM path isn't runnable here.

If the 3-month rule applies to those suites too, they need the same treatment — `atnair/vllm-legacy-deprecation` already implements it under `vllm_legacy/`.

PR #238, opened 2026-06-23
Hnimrama/ix atom

hnimra-amd:
# PR #238 — InferenceX ATOM W1 (MI300X perf gates)

**PR:** https://github.com/ROCm/cvs/pull/238  
**Head:** `hnimrama/IX-atom` → **Base:** `dev/dtni`

**Jira:** [AIMVT-236](https://amd-hub.atlassian.net/browse/AIMVT-236) · [AIMVT-244](https://amd-hub.atlassian.net/browse/AIMVT-244)

---

## Motivation

CVS needs a first-class **InferenceX ATOM** automation path aligned with the **DTNI Validation Tracker (IX ATOM)** — not the legacy `inferencemax_single` uplift. This PR:

- Introduces the **`inferencex_atom_single`** suite with an ATOM-native driver (`atom.entrypoints.openai_server` + `atom.benchmarks.benchmark_serving`).
- Ships **W1 DeepSeek R1 FP8** variant configs for **MI300X** and **MI355X** (perf, smoke, MTP3) with calibrated / CI-seeded thresholds.
- Closes **M1 / Phase A** on MI300X: W1 perf with `enforce_thresholds: true` after lab confirmation.
- Documents the IX-atom roadmap (W1–W18, accuracy, metric tiers, parity frameworks) in `plans/inferencex-atom-cvs-automation-plan.md`.

**MI355X** variant configs ship with `enforce_thresholds: false` until hardware is available — they do not block merge or MI300X milestone work.

**Base branch:** `dev/dtni`.

---

### Suite and orchestration

- Rename **`inferencemax_single` → `inferencex_atom_single`**; legacy `inferencemax/` configs removed.
- New **`InferenceXAtomJob`** (`inferencex_atom_orch.py`):
  - `params.driver=atom` → ATOM server + `benchmark_serving`; JSON artifacts parsed via `to_client_metrics`.
  - `params.driver=vllm` retained for interim GPT-OSS uplift variants only.
- Canonical config layout: flat sibling pairs under `cvs/input/config_file/inference/inferencex_atom_single/`:
  - `{gpu}_inferencex-atom-single_{model}_{precision}[_{mode}]_config.json`
  - `{gpu}_inferencex-atom-single_{model}_{precision}[_{mode}]_threshold.json`
- `schema_version: 1`, typed loader, and `ix_recipes.json` recipe pins (`dsr1-fp8-mi300x-atom`, etc.).
[...]

PR #290, opened 2026-08-05
Restore legacy vLLM suites under a 3-month deprecation notice

atnair-amd:
## Summary

`dev/dtni` replaced the per-model vLLM suites with the unified `vllm` suite, deleting the old
entry points outright. AMD's OSS policy requires a deprecated interface to stay available for
three months before removal, so this restores them behind a dated notice rather than leaving
the merge as a silent breaking change.

Notice served 2026-08-11; removal 2026-11-11.

## Change

- Legacy suites live in `cvs/tests/inference/vllm_legacy/`, not alongside the new ones. The new
  suite's `conftest.py` reorders collection from a rank table that has no entries for the legacy
  test names — sharing a directory ties them all at rank 99 and runs inference before the
  container launches. A separate directory keeps that conftest out of scope.
- `base_legacy.py` is a frozen copy of `main`'s `base.py`. The legacy suites pin to it so the
  unified suite's base can keep evolving without dragging them along.
- Importing `cvs.lib.inference.vllm` raises a `DeprecationWarning` naming the replacement.
- Out of scope: sglang and inferencemax, which were deleted the same way and need the same
  treatment separately.

## Tests

- `test_vllm_legacy_deprecation.py` (new, 15 tests) — pins the quarantine boundary, the frozen
  base, and the removal date, so re-merging the suites or re-pointing them at the live base fails.
- Hardware smoke on splinter-126-016d (1x MI300X, ROCm 7.14 GA, gpt-oss-120b): container
  lifecycle, server env contract, launch, readiness polling, and report generation all exercised
  on the restored path. `test_vllm_inference` did not complete — the GPU dropped off the PCIe bus
  (`amdgpu: device lost from bus!`) four seconds after vLLM finished warmup, and the harness
  failed cleanly on its readiness timeout. The client benchmark step remains unvalidated.
- Gates: 15/15 new unit tests pass; `ruff format --check` and `ruff check` clean on all 8 changed
  files. Full `make ut` is not runnable here — `dev/dtni` already fails `fmt-check` (23 files) and
  `lint` (14 errors) before this branch, verified against an extracted pristine baseline.

### DIFF
--- a/cvs/input/config_file/inference/vllm_legacy/mi355x_vllm_single.json
+++ b/cvs/input/config_file/inference/vllm_legacy/mi355x_vllm_single.json
@@ -1,0 +1,371 @@
+{
+    "config": {
+        "container_image": "rocm/7.0:rocm7.0_ubuntu_22.04_vllm_0.10.1_instinct_20250927_rc1",
+        "container_name": "vllm_inference_rocm",
+        "nnodes": "1",
+        "benchmark_server_script_path": "/home/{user-id}/benchmark_server_scripts/",
+        "benchmark_script_repo": "https://github.com/kimbochen/bench_serving.git",
+        "hf_token_file": "/home/{user-id}/.hf_token",
+        "shm_size": "16G",
+        "log_dir": "/home/{user-id}/LOGS",
+        "data_cache_dir": "/it-share/models/",
+        "container_config": {
+            "device_list": [
+                "/dev/dri",
+                "/dev/kfd",
+                "/dev/mem"
+            ],
+            "volume_dict": {
+                "/home/{user-id}": "/home/{user-id}",
+                "/it-share/models/": "/models"
+            },
+            "env_dict": {
+                "HF_HUB_CACHE": "/models/huggingface-cache"
+            }
+        }
+    },
+    "benchmark_params": {
+        "gpt-oss-120b": {
+            "container_image": "rocm/7.0:rocm7.0_ubuntu_22.04_vllm_0.10.1_instinct_20250927_rc1",
+            "backend": "vllm",
+            "base_url": "http://0.0.0.0",
+            "port_no": "8888",
+            "_example_dataset_name": "sharegpt|hf|random|sonnet|burstgpt",
+            "dataset_name": "random",
+            "concurrency_levels": [
+                16,
+                32,
+                64
+            ],
+            "model": "openai/gpt-oss-120b",
+            "num_prompts": "3200",
+            "sequence_combinations": [
+                {
+                    "isl": "1024",
+                    "osl": "1024",
+                    "name": "balanced"
+                },
+                {
+                    "isl": "1024",
+                    "osl": "8192",
+                    "name": "long_generation"
+                },
+                {
+                    "isl": "8192",
+                    "osl": "1024",
+                    "name": "long_context"
+                }
+            ],
+            "burstiness": "1.0",
+            "seed": "0",
+            "request_rate": "inf",
+            "max_model_length": "9216",
+            "random_range_ratio": "0.8",
+            "random_prefix_len": "0",
+            "tensor_parallelism": "1",
+            "_example_tokenizer_mode": "auto|slow|mistral|custom",
+            "tokenizer_mode": "auto",
+            "percentile_metrics": "ttft,tpot,itl,e2el",
+            "metric_percentiles": "99",
+            "server_script": "gpt-oss-120b_fp4_mi355x_vllm_docker.sh",
+            "bench_serv_script": "benchmark_serving.py",
+            "result_dict": {
+                "ISL=1024,OSL=1024,TP=1,CONC=16": {
+                    "total_throughput_per_sec": "4651",
+                    "mean_ttft_ms": "70",
+                    "mean_tpot_ms": "8"
+                },
+                "ISL=1024,OSL=1024,TP=1,CONC=32": {
+                    "total_throughput_per_sec": "7043",
+                    "mean_ttft_ms": "180",
+                    "mean_tpot_ms": "9"
+                },
+                "ISL=1024,OSL=1024,TP=1,CONC=64": {
+                    "total_throughput_per_sec": "10677",
+                    "mean_ttft_ms": "76",
+                    "mean_tpot_ms": "13"
+                },
+                "ISL=1024,OSL=8192,TP=1,CONC=16": {
+                    "total_throughput_per_sec": "2735",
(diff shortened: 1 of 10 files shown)


atnair-amd:
### Code review

Found 2 issues:

1. The two `TestFrozenLegacyBase` regression guards assert against `LEGACY_BENCH_DEFAULTS`, a module-level constant that no runtime code reads. It is written only so these tests can assert against it, so neither test guards the behaviour its comment describes. Verified by mutation: changing the real `self.bp_dict.setdefault('random_range_ration', ...)` at [base_legacy.py#L191](https://github.com/ROCm/cvs/blob/122e6f0b3013fb9af981af0e6ed25c0c471dadd0/cvs/lib/inference/base_legacy.py#L189-L193) to the corrected `random_range_ratio` spelling — precisely the drift the comment says must not happen — leaves all 15 tests passing. Gutting `clone_bench_serving_repo` to a no-op likewise passes. To actually pin the frozen behaviour, assert on the real `setdefault` result (e.g. instantiate and check `bp_dict`) and on `clone_bench_serving_repo`'s body rather than on the side constant.

https://github.com/ROCm/cvs/blob/122e6f0b3013fb9af981af0e6ed25c0c471dadd0/cvs/lib/inference/unittests/test_vllm_legacy_deprecation.py#L76-L88

2. The restored readiness polling routes bulk log reads through `exec_cmd_list`, which reintroduces the log-bloat class fixed in #278. That PR added `print_console` to `exec()`/`exec_on_head()` but deliberately left `exec_cmd_list` without it; the follow-up reply there argued it was safe because "nothing in the vLLM path routes bulk reads through it today". This PR makes that no longer true — `base_legacy.py` has 10 `exec_cmd_list` call sites, including a `tail -30`-per-node loop running up to `default_server_poll_count` iterations. Observed on a smoke run: 1008 duplicated poll-tail lines in a single run that never reached steady state. Worth either noting explicitly in the README as a known limitation of the frozen path, or threading `print_console` through `exec_cmd_list` as the #278 follow-up.

https://github.com/ROCm/cvs/blob/122e6f0b3013fb9af981af0e6ed25c0c471dadd0/cvs/lib/inference/base_legacy.py#L412-L420

Checked and found clean: all four restored suites and the config are byte-identical to their pre-#223 copies on `main` apart from the added deprecation header, so `/dev/mem` in the config, the `qwen3_80b` fixture divergence, and the missing fail-gate in `test_cleanup_stale_containers` are all faithfully preserved pre-existing behaviour rather than new defects. Dates are consistent across all files, the `DeprecationWarning` target `cvs.lib.inference.vllm_job.VllmJob` resolves, and `ruff format --check` / `ruff check` are clean on all 8 changed files at this head SHA.

atnair-amd:
Follow-up — one more finding from a git-history pass that finished after the review above.

3. The frozen base keeps `server_launch_poll_count=20` where the live base now defaults to 60, and that budget is marginal for exactly the models these suites serve. The `20 -> 60` bump landed on `dev/dtni` only (in #238, after #223 had already deleted this code), motivated by large weight loads exceeding the 20-minute window — so `main` never had it and the verbatim restore faithfully preserves the smaller budget. This is a preserved pre-existing limitation, not a regression, but it is worth knowing before anyone runs these suites: on the gpt-oss-120b smoke run I did against this branch, polling reached iteration 19 of 20 with weights already cached on local disk (`Model loading took 68.7 GiB memory and 593.389452 seconds`). A first run that has to pull weights from HF would exhaust the budget and fail on readiness timeout rather than on anything real. Three of the four suites take this default; only `vllm_qwen3_80b_single.py` overrides it, to 30.

https://github.com/ROCm/cvs/blob/122e6f0b3013fb9af981af0e6ed25c0c471dadd0/cvs/lib/inference/base_legacy.py#L78-L80

Also, for the record on the `inference_lib.py` pointer fix in this PR: `InferenceJobFactory` has no callers anywhere in the tree, and the restored suites instantiate `VllmJob` directly rather than going through it. The corrected pointer is right, but nothing exercises it.

atnair-amd:
### Correction to the review above

Adversarial re-checking of my own findings turned up errors in two of them. Retracting and correcting, with the fixes pushed in d9c51792.

**Finding 2 is withdrawn entirely — it was wrong twice over.**

- `exec_cmd_list` does accept `print_console`, and has since [b8b2cf24](https://github.com/ROCm/cvs/commit/b8b2cf24) (2026-01-02), not since #136 — that PR only moved `parallel_ssh_lib.py` to `parallel/pssh.py`. My claim that #278 "deliberately left it without" the flag was simply false.
- The suites do not use `Pssh` at all. `from cvs.lib.parallel_ssh_lib import *` aliases **`MultiProcessPssh`** ([parallel_ssh_lib.py#L35](https://github.com/ROCm/cvs/blob/d9c51792/cvs/lib/parallel_ssh_lib.py#L34-L36)), which forwards the flag at [multiprocess_pssh.py#L250-L253](https://github.com/ROCm/cvs/blob/d9c51792/cvs/lib/parallel/multiprocess_pssh.py#L249-L254).
- The premise fails outright: live `base.py` runs `tail -30` **plus** a grep per iteration over up to 60 iterations (~1860 lines) versus the frozen path's 30 x 20 = 600. **Live is strictly noisier.** Nothing is reintroduced, and my reply on #278 was not falsified by this PR.

**Finding 1 stands, but I mischaracterised the defect.** The guards are inert — six independent mutations left the suite green. But the misspelled `setdefault` is not harmless dead code. `base_legacy.py:191` setdefaults `random_range_ration` while the reads at `:286`/`:461` use `random_range_ratio`, so the default never applies:

```
legacy  keys after defaults = ['random_range_ration']  -> read raises KeyError
live    keys after defaults = ['random_range_ratio']   -> '1.0'
```

Any legacy config omitting `random_range_ratio` raises `KeyError` while building the server command. `origin/main` has the identical hole, so it stays frozen — but the accurate framing is "the typo defeats the default, the key is mandatory in config", not "unread constant".

**Finding 3 stands as filed.**

d9c51792 replaces both guards with behavioural tests parametrised over the frozen and live base, so mutating `:124` or `:191` turns them red for a runtime reason rather than a text mismatch; deletes `LEGACY_BENCH_DEFAULTS`, which was additive and only existed for the old asserts; adds a git-free sha256 content pin over the five restored files (CI checks out at depth 1 with no `origin/main` ref, and these tests also run from the sdist copy with no `.git`); and documents four known limitations of the frozen path, led by the fact that the four `*_vllm_docker.sh` server scripts named in the config ship in no repository and must be pre-staged.

Issue #283, opened 2026-07-31
Make ANC install reloacatble for tar installations


ashutom:
### Suggestion Description

Currently ANC installs on to /opt/amdtools/anc  as a default path
For deb and rpm this is fine but for the tar extraction, this may be a limitation.
Hence we should allow the user to change the installation base path with a parameter in the config file.

### ROCm Component

cvs

ashutom:
CC @cijohnson , @solaiys  for notification.

ashutom:
PR merged hence closing this one

PR #292, opened 2026-08-06
 ANC installations relocatable

ashutom:
This PR resolves the issue #283. 
This PR Make ANC installations relocatable and add support for ANC's new 1.5.0+ packaging layout

Technical Details :

1) Relocatable tar installs (ANC_INSTALL_PATH)
  - New optional ANC_INSTALL_PATH config tag lets tar releases extract to a user-chosen prefix; blank/absent falls back to the legacy /opt/amdtools. Supports {home}/~ expansion to the runner user's home.
  - Applies to tar releases only — deb/rpm packages carry predefined absolute locations in their archives and ignore the key.
  - resolve_anc_install_prefix / resolve_anc_paths thread the per-install prefix through version check, install verification, group run, and readiness.
  - Relocated tar installs rewrite content YAML exe_path entries from /opt/amdtools to the chosen prefix, leaving relative/system paths intact.
  - Auto-sudo: extraction skips sudo when the target prefix is user-writable.

2) 1.5.0+ direct packaging support
  - detect_package_flavour now returns (pkg_type, is_direct). Legacy token forms (-<flavour>- / -<flavour>.) are matched before the direct extension fallback so a legacy tar is never mistaken for a direct
  one.
  - Direct installers: deb (dpkg -i) / rpm (dnf install) download-and-install; tar untars straight into the relocatable prefix and rewrites exe_path.
  - exe_path validation is shared between legacy and direct tar installers.

3) Centralized install-path resolution
  - resolve_anc_install_location probes the node for anc.py by URL flavour, logs and caches the resolved path for the session, and fails clearly when absent. ensure_anc_ready populates it; run_anc_groups
  builds its anc.py -g command from the cached, node-verified path.

4) Fail-fast config guard
  - check_version_matches_url aborts the run before contacting any node when anc_version disagrees with the version parsed from anc_release_url (both packaging generations). Wired into
  cvs/tests/anc/conftest.py config validation.

All the UTs are passing with this change.

### DIFF
--- a/cvs/input/config_file/anc/anc_config.json
+++ b/cvs/input/config_file/anc/anc_config.json
@@ -16,6 +16,9 @@
         "_comment_anc_release_url": "URL of the ANC release archive to download and install. Flavour (deb/rpm/tar) is auto-detected from the filename.",
         "anc_release_url": "REPLACE_ME",
 
+        "_comment_ANC_INSTALL_PATH": "TAR installs ONLY: relocatable install prefix ANC is extracted into (its anc/ dir, tool folders, and content live under here). A leading '{home}' expands to the runner user's home directory (e.g. '/home/{id}/my/anc/install/path'); a leading '~' is also expanded. Applies exclusively to tar releases -- deb/rpm packages carry predefined absolute locations baked into their archive and IGNORE this key entirely. Leave blank or omit to keep the legacy default (/opt/amdtools).",
+        "ANC_INSTALL_PATH": "",
+
         "_comment_print_all_to_console": "'True' echoes ANC group output to the console; 'False' suppresses it (install/ldconfig/version diagnostics still print).",
         "print_all_to_console": "True",
 
--- a/cvs/input/config_file/anc/anc_config.json
+++ b/cvs/input/config_file/anc/anc_config.json
@@ -13,17 +13,17 @@
         "_comment_anc_version": "Expected ANC version. The install pre-task skips (re)installation when this version is already present on the node, and post-verifies the install matches.",
         "anc_version": "1.4.9",
 
-        "_comment_anc_release_url": "URL of the ANC release archive to download and install. Both packaging generations are supported and auto-detected from the filename: LEGACY (<=1.4.x) outer tarballs carry a '-deb-'/'-rpm-'/'-tar-' token (e.g. anc-release-helios-nda-1.4.9-tar-linux-x64.tar.gz); DIRECT (1.5.0+) URLs point straight at a '.deb'/'.rpm'/'.tar.gz' with no flavour token (e.g. anc-release-helios-nda-1.5.5-x86_64.tar.gz). When anc_version is set it must match the version in this URL or the run aborts before contacting any node.",
-        "anc_release_url": "REPLACE_ME",
+        "_comment_anc_release_url": "URL of the ANC release archive to download and install. Replace the placeholder changeme with the real URL before running. Both packaging generations are supported and auto-detected from the filename: LEGACY (<=1.4.x) outer tarballs carry a '-deb-'/'-rpm-'/'-tar-' token (e.g. anc-release-helios-nda-1.4.9-tar-linux-x64.tar.gz); DIRECT (1.5.0+) URLs point straight at a '.deb'/'.rpm'/'.tar.gz' with no flavour token (e.g. anc-release-helios-nda-1.5.5-x86_64.tar.gz). When anc_version is set it must match the version in this URL or the run aborts before contacting any node.",
+        "anc_release_url": "<changeme>",
 
-        "_comment_ANC_INSTALL_PATH": "TAR installs ONLY: relocatable install prefix ANC is extracted into (its anc/ dir, tool folders, and content live under here). A leading '{home}' expands to the runner user's home directory (e.g. '/home/{id}/my/anc/install/path'); a leading '~' is also expanded. Applies exclusively to tar releases -- deb/rpm packages carry predefined absolute locations baked into their archive and IGNORE this key entirely. Leave blank or omit to keep the legacy default (/opt/amdtools).",
+        "_comment_ANC_INSTALL_PATH": "TAR installs ONLY: relocatable install prefix ANC is extracted into (its anc/ dir, tool folders, and content live under here). A leading '{home}' expands to the runner user's home directory (e.g. '/home/{user-id}/my/anc/install/path'); a leading '~' is also expanded. Applies exclusively to tar releases -- deb/rpm packages carry predefined absolute locations baked into their archive and IGNORE this key entirely. Leave blank or omit to keep the legacy default (/opt/amdtools).",
         "ANC_INSTALL_PATH": "",
 
         "_comment_print_all_to_console": "'True' echoes ANC group output to the console; 'False' suppresses it (install/ldconfig/version diagnostics still print).",
         "print_all_to_console": "True",
 
-        "_comment_log_folder_path": "Controller-side destination PREFIX for ALL ANC artifacts (collected logs AND the auto-collected HTML report). Set it to a plain directory path (a leading '~' is expanded to your home directory) under which CVS lays down its own per-node/per-test/per-run folder structure. Required: replace 'REPLACE_ME' before running. To send the HTML report elsewhere, pass --html on the command line.",
-        "log_folder_path": "REPLACE_ME",
+        "_comment_log_folder_path": "Controller-side destination PREFIX for ALL ANC artifacts (collected logs AND the auto-collected HTML report). Set it to a plain directory path (a leading '~' is expanded to your home directory; '{home}'/'{user-id}' tokens are also resolved) under which CVS lays down its own per-node/per-test/per-run folder structure. Required: replace the placeholder changeme before running. To send the HTML report elsewhere, pass --html on the command line.",
+        "log_folder_path": "<changeme>",
 
         "_comment_ADD_ANC_LOGS_TO_HTML_REPORTS": "'True' always bundles the collected ANC log tree into the pytest-html report zip. 'False' (default) bundles it only when the test fails, so passing runs stay lean while failures always carry their logs.",
         "ADD_ANC_LOGS_TO_HTML_REPORTS": "False",
--- a/cvs/input/config_file/anc/anc_config.json
+++ b/cvs/input/config_file/anc/anc_config.json
@@ -13,7 +13,7 @@
         "_comment_anc_version": "Expected ANC version. The install pre-task skips (re)installation when this version is already present on the node, and post-verifies the install matches.",
         "anc_version": "1.4.9",
 
-        "_comment_anc_release_url": "URL of the ANC release archive to download and install. Flavour (deb/rpm/tar) is auto-detected from the filename.",
+        "_comment_anc_release_url": "URL of the ANC release archive to download and install. Both packaging generations are supported and auto-detected from the filename: LEGACY (<=1.4.x) outer tarballs carry a '-deb-'/'-rpm-'/'-tar-' token (e.g. anc-release-helios-nda-1.4.9-tar-linux-x64.tar.gz); DIRECT (1.5.0+) URLs point straight at a '.deb'/'.rpm'/'.tar.gz' with no flavour token (e.g. anc-release-helios-nda-1.5.5-x86_64.tar.gz). When anc_version is set it must match the version in this URL or the run aborts before contacting any node.",
         "anc_release_url": "REPLACE_ME",
 
         "_comment_ANC_INSTALL_PATH": "TAR installs ONLY: relocatable install prefix ANC is extracted into (its anc/ dir, tool folders, and content live under here). A leading '{home}' expands to the runner user's home directory (e.g. '/home/{id}/my/anc/install/path'); a leading '~' is also expanded. Applies exclusively to tar releases -- deb/rpm packages carry predefined absolute locations baked into their archive and IGNORE this key entirely. Leave blank or omit to keep the legacy default (/opt/amdtools).",
--- a/cvs/lib/anc_lib.py
+++ b/cvs/lib/anc_lib.py
@@ -44,10 +44,20 @@ from cvs.lib.run_config_paths import resolve_runner_results_base
 log = globals.log
 
 # --- Install locations ---------------------------------------------------
+# Default install prefix laid down by the deb/rpm packages and by a legacy tar
+# install. TAR installs may relocate elsewhere via config anc.ANC_INSTALL_PATH
+# (see resolve_anc_install_prefix); deb/rpm always use this fixed prefix because
+# their archives bake in absolute locations.
 ANC_TOOLS_PREFIX = "/opt/amdtools"
-# ANC directory and entrypoint laid down by the deb/rpm packages.
+# ANC directory and entrypoint at the DEFAULT prefix. Per-install paths (which
+# honour a relocated tar prefix) come from resolve_anc_paths, not these; these
+# remain the deb/rpm layout and the fallback when no relocation is configured.
 ANC_DIR = f"{ANC_TOOLS_PREFIX}/anc"
 ANC_BIN = f"{ANC_DIR}/anc.py"
+
+# config anc.ANC_INSTALL_PATH: TAR-ONLY relocatable install prefix. Blank/absent
+# -> ANC_TOOLS_PREFIX. deb/rpm ignore it entirely (fixed archive locations).
+ANC_INSTALL_PATH_KEY = "ANC_INSTALL_PATH"
 
 # validate_exe_paths.py ships beside the anc test suites.
 RESOURCES_DIR = os.path.join(
@@ -500,9 +510,69 @@ def detect_package_type(anc_release_url):
     )
 
 
-def node_version_matches(phdl, expected_version):
+# Install paths for one run: the prefix ANC is extracted under, plus the derived
+# anc/ dir and anc.py entrypoint. For deb/rpm these are always the fixed default
+# layout; for tar they honour a relocated config anc.ANC_INSTALL_PATH.
+AncPaths = namedtuple("AncPaths", ["prefix", "anc_dir", "anc_bin"])
+
+
+def resolve_anc_install_prefix(config_dict, pkg_type):
     '''
-    Query ``ANC_BIN --version`` on every node and report version matches.
+    Resolve the install PREFIX ANC is laid down under for this run.
+
+    Only tar installs are relocatable: when pkg_type == "tar" and config
+    anc.ANC_INSTALL_PATH is set to a non-blank value, that value (leading "~"
+    expanded; "{home}"/other tokens are already substituted by the config
+    placeholder pass in the ANC conftest) is the prefix. A blank/absent value,
+    or ANY deb/rpm install, resolves to the fixed ANC_TOOLS_PREFIX because those
+    packages carry predefined absolute locations baked into their archive.
+
+    The returned prefix is normalised to an absolute path with no trailing slash
+    so it composes cleanly into "<prefix>/anc/anc.py".
+    '''
+    prefix = ANC_TOOLS_PREFIX
+    if pkg_type == "tar":
+        raw = config_dict.get("anc", {}).get(ANC_INSTALL_PATH_KEY)
+        if raw and str(raw).strip():
+            prefix = os.path.expanduser(str(raw).strip())
+    return os.path.abspath(prefix)
+
+
+def resolve_anc_paths(config_dict, pkg_type):
+    '''
+    Build the AncPaths (prefix, anc_dir, anc_bin) for this install flavour.
+
+    anc_dir is "<prefix>/anc" and anc_bin is "<prefix>/anc/anc.py", mirroring the
+    layout the release archives lay down under any prefix.
+    '''
+    prefix = resolve_anc_install_prefix(config_dict, pkg_type)
+    anc_dir = f"{prefix}/anc"
+    return AncPaths(prefix=prefix, anc_dir=anc_dir, anc_bin=f"{anc_dir}/anc.py")
+
+
+def resolve_anc_paths_from_config(config_dict):
+    '''
+    Resolve AncPaths from config alone, detecting the package flavour from
+    anc.anc_release_url. Used by run-time consumers (version check, group run,
+    ldconfig-independent path lookups) that need the live install location
+    without re-passing pkg_type. Falls back to the default prefix if the URL is
+    missing or its flavour cannot be determined.
+    '''
+    url = config_dict.get("anc", {}).get("anc_release_url") or ""
+    try:
+        pkg_type = detect_package_type(url) if url and str(url).strip() else ""
+    except ValueError:
+        pkg_type = ""
+    return resolve_anc_paths(config_dict, pkg_type)
+
+
+def node_version_matches(phdl, expected_version, anc_bin=ANC_BIN):
+    '''
+    Query ``<anc_bin> --version`` on every node and report version matches.
+
+    ``anc_bin`` is the anc.py entrypoint path; it defaults to the fixed-prefix
+    ANC_BIN but callers pass the per-install path (which may be a relocated tar
+    prefix) so a version check hits the location ANC was actually installed to.
 
     The version is matched as a whole token (bounded by non-version characters),
     not a substring, so expecting "1.4.7" does NOT match "1.4.70" or "1.4.7-rc".
@@ -512,7 +582,7 @@ def node_version_matches(phdl, expected_version):
       standalone version token in the node's ``anc.py --version`` output (False
       if ANC is absent or reports a different version).
     '''
-    out_dict = phdl.exec(f"{ANC_BIN} --version 2>&1 || true", timeout=60)
+    out_dict = phdl.exec(f"{anc_bin} --version 2>&1 || true", timeout=60)
     print_test_output(log, out_dict)
     # Boundaries are any char that isn't part of a version token (digits, dots,
     # hyphens, alphanumerics) so "1.4.7" won't match inside "1.4.70"/"1.4.7-rc".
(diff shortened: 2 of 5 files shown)


cvs/lib/anc_lib.py
```diff
@@ -545,8 +772,20 @@ def install_anc(phdl, cluster_dict, config_dict):
     _assert_shell_safe(anc_cfg, ("anc_release_url",))
```
solaiys: **Blocking — the new `ANC_INSTALL_PATH` is missing from this guard.**

This validates only `anc_release_url`, but `ANC_INSTALL_PATH` follows the exact same path: a config value interpolated into remote shell commands, several of them under `sudo`. It never reaches this boundary, so nothing checks it.

```python
_assert_shell_safe(anc_cfg, ("anc_release_url", ANC_INSTALL_PATH_KEY))
```

Adding the key is necessary but **not sufficient on its own**: `_assert_shell_safe` rejects only `'`, and one of the sinks (the `rm -rf` on lines 1174 and 1308) is double-quoted, so `$` and backtick need rejecting too. See the other inline notes.

cvs/lib/anc_lib.py
```diff
@@ -478,41 +491,252 @@ def _fail_unreachable_nodes(cluster_dict, out_dict, action):
...
+    packages carry predefined absolute locations baked into their archive.
+
+    The returned prefix is normalised to an absolute path with no trailing slash
+    so it composes cleanly into "<prefix>/anc/anc.py".
+    '''
+    prefix = ANC_TOOLS_PREFIX
+    if pkg_type == "tar":
+        raw = config_dict.get("anc", {}).get(ANC_INSTALL_PATH_KEY)
+        if raw and str(raw).strip():
+            prefix = os.path.expanduser(str(raw).strip())
```
solaiys, line 720: **Blocking (root cause).** Neither `expanduser` here nor the `abspath` below strips shell metacharacters, so whatever the user puts in `ANC_INSTALL_PATH` reaches the remote shell verbatim — `'`, `$(...)`, backticks, `\`, and newlines all survive path normalisation.

That resolved prefix is then interpolated into commands that run on every node (lines 1082, 1152, 1172, 1174, 1988), so it needs validating at a boundary. Two options:

- Add `ANC_INSTALL_PATH_KEY` to `_assert_shell_safe` (line 772) **and** extend that guard to reject `$`, backtick, `\` and newline in addition to `'`; or
- validate the resolved prefix against a safe-path allowlist such as `^[A-Za-z0-9_@%+=:,./-]+$`.

Worth also rejecting a prefix of `/` while you're here — see the note on line 1174 for why.

cvs/lib/anc_lib.py
```diff
@@ -723,6 +972,120 @@ def _install_anc_deb(phdl, cluster_dict, config_dict):
...
+
+    We cannot test the (not-yet-existing) prefix directly, so walk up to its
+    deepest EXISTING ancestor and test whether THAT is writable by the current
+    user. A user-writable ancestor (e.g. the home dir for a relocated install)
+    means we can mkdir/extract without sudo, so files land user-owned and can be
+    cleaned up later without root. A non-writable ancestor (e.g. /opt) selects
+    sudo. The result is exported as ``$SUDO`` for the extract/rewrite commands.
+    '''
+    return (
+        f"probe='{prefix}'; "
```
solaiys, line 1179: Single-quoted interpolation of an unvalidated value: a `'` anywhere in `ANC_INSTALL_PATH` closes this quote and the remainder of the path becomes shell syntax, so the command fails on every node with a confusing remote parse error rather than a clear message.

This is precisely the failure mode `_assert_shell_safe` was written to prevent for `anc_release_url` — the new key simply isn't covered (line 772).

cvs/lib/anc_lib.py
```diff
@@ -780,39 +1169,49 @@ def _install_anc_tar(phdl, cluster_dict, config_dict):
...
-        f"sudo mkdir -p '{ANC_TOOLS_PREFIX}'; "
-        f"echo 'Replacing top-level folders under {ANC_TOOLS_PREFIX}...'; "
-        f"for name in $tops; do echo \"  {ANC_TOOLS_PREFIX}/$name\"; sudo rm -rf \"{ANC_TOOLS_PREFIX}/$name\"; done; "
-        f"echo 'Extracting anc-tool archive to {ANC_TOOLS_PREFIX}...'; "
-        f"sudo tar -xzf anc-tool*.tar.gz -C '{ANC_TOOLS_PREFIX}'; "
-        f"echo 'Extracting anc-content archive to {ANC_TOOLS_PREFIX}...'; "
-        f"sudo tar -xzf anc-content*.tar.gz -C '{ANC_TOOLS_PREFIX}'; "
+        f"$SUDO mkdir -p '{prefix}'; "
+        f"echo 'Replacing top-level folders under {prefix}...'; "
+        f"for name in $tops; do echo \"  {prefix}/$name\"; $SUDO rm -rf \"{prefix}/$name\"; done; "
```
solaiys: **Blocking — sharpest sink, and it needs no quote character to go wrong.**

Unlike the other interpolations, `"{prefix}/$name"` is **double-quoted**, so the shell still expands inside it:

- `$(...)` or backticks in `ANC_INSTALL_PATH` are command-substituted right here.
- A stray `$VAR` expands to empty, so this `rm -rf` silently targets a **different path** than the operator intended.

Single-quoting the prefix closes both:

```
$SUDO rm -rf '{prefix}'/"$name"
```

Separately: there's no check that the prefix isn't `/`. With `ANC_INSTALL_PATH="/"` this loop becomes `rm -rf /anc`, `/rvs`, and so on for every top-level name the archive ships.

cvs/lib/anc_lib.py
```diff
@@ -839,6 +1238,97 @@ def _install_anc_tar(phdl, cluster_dict, config_dict):
...
+        f"cd \"$STAGE\"; "
+        f"echo 'Downloading ANC release...'; "
+        f"{_download_with_progress_snippet(anc_release_url, 'anc.tar.gz')}; "
+        # Top-level folders shipped by the archive; strip a leading './' and drop
+        # the '.' root entry so only real folder names remain.
+        f"echo 'Determining top-level tool folders...'; "
+        f"tops=$( tar -tzf anc.tar.gz | sed 's#^\\./##' | awk -F/ '{{print $1}}' | grep -vE '^[.]?$' | sort -u ); "
+        f"$SUDO mkdir -p '{prefix}'; "
+        f"echo 'Replacing top-level folders under {prefix}...'; "
+        f"for name in $tops; do echo \"  {prefix}/$name\"; $SUDO rm -rf \"{prefix}/$name\"; done; "
```
solaiys: Same double-quoted `rm -rf` problem as line 1174, duplicated here in the direct-tar installer — it needs the identical fix, otherwise fixing only the legacy path leaves 1.5.0+ installs exposed.

solaiys (commented):
Line-specific notes on the one blocking issue are inline below; a plain-language summary is in the PR comment.

solaiys:
## Review summary: one blocking issue

The relocatable-install work reads well overall. In particular I checked and am happy with: the legacy `-<flavour>-` token being matched *before* the direct extension fallback (so a legacy `...-tar-linux-x64.tar.gz` is never mistaken for a direct tar), the fail-fast version guard being wired into ANC config validation so a version/URL mismatch aborts before any node is contacted, and the new session-scoped install-path cache, which introduces no concurrency risk (single writer, no new threads or background jobs).

**The one blocking issue: the new `ANC_INSTALL_PATH` config value is interpolated into remote shell commands without ever passing the `_assert_shell_safe` guard.**

In plain terms: `anc_release_url` gets validated at that boundary, but `ANC_INSTALL_PATH` does not, and `resolve_anc_install_prefix` only applies `expanduser`/`abspath` — neither of which strips shell metacharacters. So whatever the user types into that config key is handed to the shell verbatim on every node, in commands that mostly run under `sudo`.

Two concrete consequences:

- A path containing a quote (e.g. `/home/o'brien/anc`) breaks the command's quoting and fails across the whole cluster with a confusing remote parse error, instead of the clear fail-fast message the guard was written to produce.
- One sink is **double-quoted** rather than single-quoted, so it misbehaves without any quote character at all: a `$(...)` there is command-substituted, and a stray `$VAR` expands to empty — which means a `rm -rf` can silently delete the **wrong** path.

The fix is small, but needs both halves:

1. Add the key to the guard: `_assert_shell_safe(anc_cfg, ("anc_release_url", ANC_INSTALL_PATH_KEY))`.
2. Harden it, since the guard currently rejects only `'` — also reject `$`, backtick, backslash and newline, or validate the resolved prefix against a safe-path allowlist. Single-quoting the `rm -rf` argument closes that specific sink.

I've left **inline comments on the exact lines** (the guard, the prefix resolution, and the three interpolation sites, including the duplicated one in the direct-tar installer) so the line references are attached to the code rather than listed here. Nothing else is blocking.

cvs/input/config_file/anc/anc_config.json
```diff
@@ -13,9 +13,12 @@
         "_comment_anc_version": "Expected ANC version. The install pre-task skips (re)installation when this version is already present on the node, and post-verifies the install matches.",
         "anc_version": "1.4.9",
 
-        "_comment_anc_release_url": "URL of the ANC release archive to download and install. Flavour (deb/rpm/tar) is auto-detected from the filename.",
+        "_comment_anc_release_url": "URL of the ANC release archive to download and install. Both packaging generations are supported and auto-detected from the filename: LEGACY (<=1.4.x) outer tarballs carry a '-deb-'/'-rpm-'/'-tar-' token (e.g. anc-release-helios-nda-1.4.9-tar-linux-x64.tar.gz); DIRECT (1.5.0+) URLs point straight at a '.deb'/'.rpm'/'.tar.gz' with no flavour token (e.g. anc-release-helios-nda-1.5.5-x86_64.tar.gz). When anc_version is set it must match the version in this URL or the run aborts before contacting any node.",
         "anc_release_url": "REPLACE_ME",
```
solaiys: "REPLACE_ME",  change this to "**\<changeme\>**"

<changeme> is a tracked tag which will be warned to user if it is present in the config file. So that user will know that he has to update something in the config file.

Also **{user-id}** is the right re-defined tag which gets auto resolved during the config_file fixture.

solaiys:
**one more IMPORTANT change is needed in the full ANC code:**
There are places you have used /tmp/ locations and paths to create folders and files.
It seems I missed this in the initial PR#269.

Please change it to /tmp/{user-id}/   ( if it is in config files , {user-id} will be auto-resolved)
Else, inside the code, you can change the path to be /tmp/current-user-id/    ( current user-id shuold be resolved dynamically)  . 

**Reason:**
when another user run the same ANC test on the same node, if those /tmp/* folders or files exists under your userid ownership, it will throw the permission error for the other user.

cvs/lib/anc_lib.py
ashutom: Fixed in d0335091. Added ANC_INSTALL_PATH_KEY to the _assert_shell_safe call in install_anc, and broadened the guard itself — it now rejects ', ", `, $, \, and newline (via a SHELL_UNSAFE_CHARS set), not just the single quote. 
So the double-quoted sinks you flagged (rm -rf, chown) can no longer be reached with a $/backtick payload. Unit tests assert each character is rejected.

cvs/lib/anc_lib.py
ashutom: Fixed both halves. 
The cleanup loop now single-quotes the untrusted prefix and only leaves the archive-derived name expandable: $ sudo rm -rf '{prefix}'/"$name" — so $(...)/backtick/$VAR in ANC_INSTALL_PATH can no longer be substituted or collapse the target. The / case is handled separately in resolve_anc_install_prefix (rejects root, incl. //). Additionally, $name comes from the archive listing, so I filter ./.. out of the top-level set and added a per-name case */*|.|..) continue guard, so a malicious archive entry can't make the rm -rf escape the prefix either.

cvs/lib/anc_lib.py
ashutom: Done. 
ANC_INSTALL_PATH is now validated at its single resolve choke point (resolve_anc_install_prefix), which every consumer — install, group run, version check — reaches: it runs _assert_shell_safe before expanduser/abspath (which only normalize, never sanitize), and rejects a prefix that resolves to filesystem root. I took your first option (extend the guard) rather than the allow list, since it composes with the existing URL guard. Note on the / check: I test not resolved.strip("/") rather than == "/" because abspath preserves a POSIX-special // — so /, //, /// are all rejected now.

cvs/lib/anc_lib.py
ashutom: Agreed — that was the gap. ANC_INSTALL_PATH now goes through _assert_shell_safe (which rejects ' among others), so a quote in the path is caught at the boundary with a clear message instead of producing a confusing remote parse error. The probe='{prefix}' line and the other single-quoted sinks are covered by that same guard.

cvs/lib/anc_lib.py
ashutom: Applied the identical fix to _install_anc_tar_direct as well as the legacy _install_anc_tar — single-quoted prefix, ./.. filter, and the per-name guard — so the 1.5.0+ path isn't left exposed. Both installers have unit tests asserting the safe command shape.

cvs/input/config_file/anc/anc_config.json
ashutom: Done.
Migrated anc_release_url and log_folder_path to <changeme> and removed the bespoke REPLACE_ME machinery, so the standard resolver's hard-exit now covers ANC like every other suite (verified it aborts on the shipped config). One clarification: {user-id} auto-resolves for paths that embed the username, so I used it in the path-style comments but it doesn't apply to anc_release_url (a release URL has no user in it), so that field just uses <changeme>. Also fixed a {id} → {user-id} typo in the ANC_INSTALL_PATH comment.

ashutom:
> **one more IMPORTANT change is needed in the full ANC code:** There are places you have used /tmp/ locations and paths to create folders and files. It seems I missed this in the initial PR#269.
> 
> Please change it to /tmp/{user-id}/ ( if it is in config files , {user-id} will be auto-resolved) Else, inside the code, you can change the path to be /tmp/current-user-id/ ( current user-id shuold be resolved dynamically) .
> 
> **Reason:** when another user run the same ANC test on the same node, if those /tmp/* folders or files exists under your userid ownership, it will throw the permission error for the other user.


Fixed. 
All remote temp paths are now namespaced under /tmp/<username> (username from the cluster file): the validate_exe_paths.py upload, the log tarball in _pull_log_dir, and the quiet-mode run-output file — each with a mkdir -p first. So a second user running ANC on the same node no longer collides with another user's /tmp/* ownership. I used the SSH username rather than the {user-id} config token since these paths are built in code, not config, but the effect is the same per-user isolation you described.

ashutom:
> ## Review summary: one blocking issue
> 
> The relocatable-install work reads well overall. In particular I checked and am happy with: the legacy `-<flavour>-` token being matched _before_ the direct extension fallback (so a legacy `...-tar-linux-x64.tar.gz` is never mistaken for a direct tar), the fail-fast version guard being wired into ANC config validation so a version/URL mismatch aborts before any node is contacted, and the new session-scoped install-path cache, which introduces no concurrency risk (single writer, no new threads or background jobs).
> 
> **The one blocking issue: the new `ANC_INSTALL_PATH` config value is interpolated into remote shell commands without ever passing the `_assert_shell_safe` guard.**
> 
> In plain terms: `anc_release_url` gets validated at that boundary, but `ANC_INSTALL_PATH` does not, and `resolve_anc_install_prefix` only applies `expanduser`/`abspath` — neither of which strips shell metacharacters. So whatever the user types into that config key is handed to the shell verbatim on every node, in commands that mostly run under `sudo`.
> 
> Two concrete consequences:
> 
>     * A path containing a quote (e.g. `/home/o'brien/anc`) breaks the command's quoting and fails across the whole cluster with a confusing remote parse error, instead of the clear fail-fast message the guard was written to produce.
> 
>     * One sink is **double-quoted** rather than single-quoted, so it misbehaves without any quote character at all: a `$(...)` there is command-substituted, and a stray `$VAR` expands to empty — which means a `rm -rf` can silently delete the **wrong** path.
> 
> 
> The fix is small, but needs both halves:
> 
>     1. Add the key to the guard: `_assert_shell_safe(anc_cfg, ("anc_release_url", ANC_INSTALL_PATH_KEY))`.
> 
>     2. Harden it, since the guard currently rejects only `'` — also reject `$`, backtick, backslash and newline, or validate the resolved prefix against a safe-path allowlist. Single-quoting the `rm -rf` argument closes that specific sink.
> 
> 
> I've left **inline comments on the exact lines** (the guard, the prefix resolution, and the three interpolation sites, including the duplicated one in the direct-tar installer) so the line references are attached to the code rather than listed here. Nothing else is blocking.



Both halves are in d0335091.

1) ANC_INSTALL_PATH is now passed through _assert_shell_safe; 
2) the guard is hardened to reject ', ", `, $, \, newline. 

Beyond that I closed the related sinks you pointed to inline: single-quoted, the rm -rf prefix in both installers, escaped the sed-replacement metacharacters (\ & #) in the exe_path rewrite, and — found while hardening — validated the cluster username (it reaches sudo chown but never went through the config guard) and namespaced remote /tmp paths per user. All fail-fast checks are consolidated into validate_anc_config so a bad prefix/URL/username aborts cleanly before any node is contacted, rather than as a mid-run traceback.

cvs/tests/anc/README.md
```diff
@@ -129,25 +131,30 @@ The ANC config lives at `cvs/input/config_file/anc/anc_config.json`:
         "description": "AMD Node Check",
         "inactivity_timeout": 900,
         "install_timeout": 1800,
-        "anc_version": "1.4.7",
-        "anc_release_url": "https://.../anc-release-helios-nda-1.4.7-rpm-linux-x64.tar.gz",
+        "anc_version": "1.4.9",
+        "anc_release_url": "REPLACE_ME",
```
solaiys: This PR adds these `REPLACE_ME` lines to the README, but it also deletes `REPLACE_ME_SENTINEL` and migrates the shipped config to `<changeme>` — so nothing recognises this string any more.

Following this example now gives a worse failure than before: a literal `REPLACE_ME` passes the non-blank check in `validate_anc_config`, passes `_prefix_problem`, and `_resolve_prefix` hands it back as a plain relative path. The result is ANC trying to download `REPLACE_ME` as a URL and writing logs to a relative directory, instead of the clean "replace the placeholder" abort the old sentinel gave.

Suggest `REPLACE_ME` -> `<changeme>` here, in the `log_folder_path` line below, and in the sentence at line 145 (and the table cell at 159 mentions the "shipped `REPLACE_ME`" too). Docs-only, but this PR is what introduced the mismatch.

solaiys (commented):
## Re-review @ `d0335091` — the blocking issue is properly fixed

I re-traced `ANC_INSTALL_PATH` from config parsing to every remote shell sink, and the hardening is more thorough than what I asked for. One small documentation inconsistency this PR introduces is worth a fix before merge; nothing else is blocking.

### The shell-injection blocker is closed, on both halves

**The guard** (`anc_lib.py:416`) now rejects `'`, `"`, backtick, `$`, `\` and newline, so the double-quoted sink that worried me most is no longer reachable with a `$(...)` or a bare `$VAR`.

**Coverage is better than the two-line fix I suggested.** Rather than only adding the key to the `install_anc` call, validation moved to a single choke point in `resolve_anc_install_prefix` (`:719`). I checked that this actually holds: `ANC_INSTALL_PATH_KEY` is read in exactly one place in the whole file (`:717`), and every consumer — install, group run, version check — reaches it through `resolve_anc_paths`. The extra `_assert_shell_safe(anc_cfg, ("anc_release_url", ANC_INSTALL_PATH_KEY))` at `:869` is then belt-and-braces rather than the load-bearing check. That's the right shape; it can't be bypassed by a future caller that forgets the guard.

**The root check is sharper than my suggestion.** I proposed rejecting `== "/"`; the implementation tests `not resolved.strip("/")` (`:726`) with a comment explaining that `abspath` preserves a POSIX-special leading `//`, so `//` and `//x/..` are also filesystem root. My version would have missed those.

**Both `rm -rf` loops are fixed**, legacy (`:1275`) and direct-tar (`:1417`), and the fix is layered rather than relying on the guard alone:

```1275:1275:cvs/lib/anc_lib.py
        f"echo \"  {prefix}/$name\"; $SUDO rm -rf '{prefix}'/\"$name\"; done; "
```

The prefix is single-quoted (and `'` is now rejected upstream), and `$name` can't escape the prefix through three independent filters: `awk -F/ '{print $1}'` keeps only the first path component, `grep -vE '^([.]{1,2})?$'` drops `.`/`..`/empty, and a `case "$name" in */*|.|..) continue` guard repeats the check in the loop. So a malicious or corrupt archive can't redirect the delete.

**The `sed` sinks are escaped**, at `:1249` and `:1396`, via `_sed_replacement_safe`. The escape order is correct — backslash first, then `&` and the `#` delimiter — so the added escapes aren't themselves re-escaped.

### The two issues you found while hardening are real ones

The cluster **username** reaching `sudo chown` without ever passing the config guard was a genuine gap I missed; `validate_cluster_username` rejecting whitespace as well as the shell set is the right call, since a space would have split into two `chown` arguments. And consolidating everything into `validate_anc_config`, wired into the conftest at `conftest.py:101`, means a bad prefix, URL or username now aborts before any node is contacted instead of surfacing as a mid-run traceback.

### My `/tmp` request is addressed

All three sites are namespaced under `/tmp/<user>` — the `validate_exe_paths.py` upload (`:1319`), the log tarball (`:1717`), and the quiet-mode run output (`:2120`) — each with its own `mkdir -p` first, and each sanitised to a single path component. Using the SSH username instead of the `{user-id}` config token is fine; these paths are built in code, and the isolation is what I was after. The log-tarball case is handled correctly too: the archive is created by root inside the user-owned directory and then chowned back so SFTP can pull it.

### Checked and clear

- **The `<changeme>` migration doesn't regress the placeholder check.** This one needed care, because the removed `find_replace_me_placeholders` deliberately skipped `_comment*` keys and required an exact match, whereas the standard resolver (`utils_lib.py:293`) does a **substring** match and skips nothing — so a comment merely *mentioning* the placeholder would hard-exit a correctly-filled config. I walked the shipped config: exactly two strings contain `<changeme>` (`anc.anc_release_url`, `anc.log_folder_path`, both intended), and the two comments that discuss it deliberately write "changeme" without the angle brackets. No false positives.
- **Tests.** 86 new tests, all passing, and they cover the security fixes specifically rather than just the happy path (`TestTarCleanupFiltersDotDot`, `TestValidateClusterUsername`, the shell-safe cases). Full suite goes 484 → 570 with no new failures.
- **Lint.** `ruff format --check` and `ruff check` clean on all three Python files, pylint 10.00/10 on the logging gate.
- **Concurrency**, re-checked after the refactor: still a single writer to the module-level cache, no threads or background jobs introduced.

### One thing to fix before merge (docs, not code)

This PR adds three README lines that tell users to write `REPLACE_ME`, while the same PR deletes the code that recognised it:

```135:145:cvs/tests/anc/README.md
        "anc_release_url": "REPLACE_ME",
        "log_folder_path": "REPLACE_ME",
`anc_release_url` and `log_folder_path` ship as the `REPLACE_ME` sentinel and
```

The shipped config now uses `<changeme>`, and `REPLACE_ME_SENTINEL` is gone, so nothing recognises that string any more. Reading the code paths: a literal `REPLACE_ME` passes the non-blank check in `validate_anc_config`, passes `_prefix_problem`, and then `_resolve_prefix` returns it as a plain relative path — so an operator who follows the README gets ANC trying to download `REPLACE_ME` as a URL and writing logs to a relative directory, rather than the clean "replace the placeholder" message the old sentinel produced. It's a three-line docs edit (`REPLACE_ME` → `<changeme>` in the example block and the sentence at line 145), but worth doing in this PR since this PR is what created the mismatch.

Minor, entirely optional: `_assert_shell_safe` runs on the raw value before `expanduser`, so in principle a home directory containing a metacharacter could reintroduce one after validation. Pathological and out of scope for the config-is-trusted threat model this guard operates in — noting it only for completeness.

Aside from the README lines, this looks good to me — the security fix is well-constructed and the tests back it up.

solaiys (approved):
## Approving @ `fd6053e0`

The one item left from my re-review at `d0335091` — the README still telling users to write `REPLACE_ME` after this PR deleted the code that recognised it — is fixed, and fixed more thoroughly than I asked for. All seven review threads are resolved and nothing blocking remains.

### The README fix is correct, and the added claims are accurate

`fd6053e0` is docs-only (`README.md`, +13/-9). Rather than just swapping the three strings, it also documents the validation and placeholder behaviour. I verified each new claim against the code rather than taking it at face value:

| README claim | Verified |
| --- | --- |
| `anc_release_url` / `log_folder_path` ship as `<changeme>` | Matches the shipped config; the example block is now key-for-key identical to `anc_config.json` |
| unresolved `<changeme>` "hard-exits before any node is contacted" | Ran the real resolver on the shipped config: `SystemExit(1)` at load. `conftest.py:94` resolves before `validate_anc_config` at `:101`, so it precedes every node contact |
| `{home}` / `{user-id}` "resolved during config load" | Both are real tokens in `resolve_test_config_placeholders` (`utils_lib.py:422-427`), and the whole `anc` dict passes through it |
| metacharacters `'` `"` `` ` `` `$` `\` newline rejected | Exactly `SHELL_UNSAFE_CHARS = ("'", '"', "`", "$", "\\", "\n")` at `anc_lib.py:416` — the enumeration matches the guard with nothing missing or over-claimed |
| prefix resolving to root (`/`, `//`) rejected | `not resolved.strip("/")` at `:726`, which covers `/`, `//`, `///` |

Two things I checked specifically because they were the failure modes worth worrying about: `REPLACE_ME` is now **gone from the repo entirely** (no stale references left anywhere), and the migration still produces **no false positives** — the standard resolver does a *substring* match and skips nothing, but the `_comment_*` text deliberately writes "changeme" without angle brackets, so a correctly-filled config loads cleanly. Confirmed by running it.

### Verification on the actual merge result

CI is green on this exact head (`container-smoke` and `test`, both `success` for `fd6053e0`). Since CI ran against an older base, I also built the merge with current `main` and tested that:

- Merge is clean, no conflicts. `main` has made **zero** commits to any ANC file, so there is no semantic overlap with the 66 commits it has moved ahead.
- `main` did touch `utils_lib.py`, which this PR depends on, so I checked it directly: `resolve_test_config_placeholders` has an identical signature and an identical `<changeme>` guard on both sides, so the merged call site is sound.
- Merged tree: **1341 tests, OK**, zero failures or errors, including all 86 ANC tests.
- `ruff format --check` and `ruff check` clean on all three changed Python files; `anc_config.json` parses.

The security work from the earlier rounds (the choke-point prefix validation, the layered `rm -rf` filters, the escaped `sed` sinks, the per-user `/tmp` namespacing, and the cluster-username guard you found yourself) is unchanged since I traced it at `d0335091`.

Non-blocking, informational only: the branch is 66 commits behind `main`, so a rebase or merge before landing would get it tested on the current base — though as noted above I already ran that combination locally and it is clean.

PR #293, opened 2026-08-06
Close SSH sessions left open by temporary subset Pssh handles

atnair-amd:
## Summary

`BaremetalOrchestrator.exec`, `BaremetalOrchestrator.setup_env` and
`DockerRuntime.exec` build a throwaway `Pssh` for a host subset and drop it on
return. On timeout the per-host greenlet in `ParallelSSHClient.cmds` never
finishes; it holds a bound method of the client, so `SSHClient.__del__` never
runs and the sshd session survives. vLLM readiness polling does this per rank up
to 60 times a run, until sshd's per-user limit starts rejecting unrelated logins.

Ticket: AIMVT-283

## Change

- `Pssh.destroy_clients()` was `del self.client`, a no-op here. It now kills
  pending greenlets and `_disconnect()`s each host client, tolerating errors so
  one dead host can't abort the rest.
- All three subset call sites tear down in `try/finally`. `DockerRuntime.exec`
  builds its own `Pssh` instead of delegating, so container mode needs its own
  fix. `setup_env` has no callers today but is an `abstractmethod` on the base
  orchestrator, so it gets the same treatment.
- Shared `orch.head`/`orch.all` are reused, not rebuilt, and measure flat across
  repeated timeouts; left alone, with a guard test.
- Uses parallel-ssh internals (`_host_clients`, `_disconnect()`) — the public
  `disconnect()` is a documented no-op. Re-verify on any parallel-ssh bump.
- Out of scope: caching subset handles; sshd-side session limits.

## Tests

7 unit tests; reverting the fix turns them red. Against a live sshd, 8
timing-out subset execs grew sessions 1..8 before and hold flat at 1 after.
Clean execs never leaked.

Gate: 1110 tests, 4 failures, all 4 identical to the `dev/dtni` baseline
(1103/same 4). `fmt-check` and `lint` clean on every file touched here.

### DIFF
--- a/cvs/core/orchestrators/baremetal.py
+++ b/cvs/core/orchestrators/baremetal.py
@@ -108,7 +108,10 @@ class BaremetalOrchestrator(Orchestrator):
                 host_key_check=False,
                 stop_on_errors=self.stop_on_errors,
             )
-            return pssh.exec(cmd, timeout=timeout, detailed=detailed, print_console=print_console)
+            try:
+                return pssh.exec(cmd, timeout=timeout, detailed=detailed, print_console=print_console)
+            finally:
+                pssh.destroy_clients()
 
     def sudo_prefix(self):
         """
@@ -170,7 +173,10 @@ class BaremetalOrchestrator(Orchestrator):
                 host_key_check=False,
                 stop_on_errors=self.stop_on_errors,
             )
-            result = pssh.exec(f"bash {env_script}", timeout=60, detailed=True)
+            try:
+                result = pssh.exec(f"bash {env_script}", timeout=60, detailed=True)
+            finally:
+                pssh.destroy_clients()
 
         # Check if all hosts succeeded
         success = all(output['exit_code'] == 0 for output in result.values())
--- a/cvs/core/orchestrators/unittests/test_baremetal.py
+++ b/cvs/core/orchestrators/unittests/test_baremetal.py
@@ -279,5 +279,64 @@ class TestBaremetalOrchestratorSudoPrefix(unittest.TestCase):
         pssh_instance.exec.assert_called_once_with("sudo -n true >/dev/null 2>&1; echo $?")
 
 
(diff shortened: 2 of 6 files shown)


solaiys:
No Blocking issues found.

**What the fix does:** destroy_clients() in cvs/lib/parallel/pssh.py now kills pending command greenlets and calls _disconnect() on each per-host client before dropping the reference, and both subset call sites (BaremetalOrchestrator.exec, DockerRuntime.exec) invoke it from a try/finally.


**Issue 1 (Medium — same bug, not fixed): setup_env in baremetal.py**

The setup_env method has the exact same leak pattern. Look at the current code around line 172:


```
def setup_env(self, hosts, env_script=None):
    ...
    else:
        pssh = Pssh(self.log, hosts, ...)
        result = pssh.exec(f"bash {env_script}", timeout=60, detailed=True)
        # ← pssh dropped here without destroy_clients()

```
It needs the same try/finally treatment. **It's less likely to bite in practice (setup runs once, not in a polling loop),** but for consistency it should be fixed in this PR since it's literally the same bug in the same file.

atnair-amd:
Fixed in 23e2d663 — `setup_env` now wraps its subset `exec` in the same `try/finally`, with a subTest-parametrized test over the returns/raises paths.

Two notes on the reasoning, since I'd originally left it out deliberately:

- It has zero callers repo-wide. The other `setup_env` hits are the `@abstractmethod` declaration in `base.py:65` and `mori_benchmark_test.py:203`, a pytest function that calls `create_env_script()` and is unrelated. So the practical risk today is nil, not just lower. The reason to fix it anyway is that it's an abstractmethod on the base class — an implementation could reach it, and then it ships live.
- Minor: the construction is at `baremetal.py:167`, not ~172.

solaiys (approved):
LGTM..

amd-droy (approved):
looks good to me.

PR #295, opened 2026-08-07
report: comment out Value/Unit columns in inference suite HTML reports

atnair-amd:
Ticket: AIMVT-284

## Summary
Removes the Value/Unit columns from the pytest-html results table across
the vllm, sglang, and inferencex_atom inference suites.

## Change
- Commented out `pytest_html_results_table_header`/`_row` in each suite's
  `conftest.py`, plus the shared `html_metric_table_header`/`_row` helper
  in `inference_suite_lifecycle.py` (and ATOM's now-unused import of it)
- vllm and inferencex_atom previously populated real per-metric values;
  sglang's columns were already always blank (no sglang test sets
  metric_value/metric_unit)
- Out of scope: sglang's separate per-stage lifecycle mini-table
  (stage/value/unit), which is a different, smaller HTML extra untouched
  by this change

## Tests
- `.test_venv/bin/python run_all_unittests.py`: 1110 tests, same 4
  pre-existing failures as on `dev/dtni` baseline, no new failures
- `dev/dtni` baseline fmt-check/lint are already red (unrelated,
  pre-existing); diffed before/after and confirmed this change adds no
  new fmt-check/lint violations

### DIFF
--- a/cvs/lib/inference/utils/inference_suite_lifecycle.py
+++ b/cvs/lib/inference/utils/inference_suite_lifecycle.py
@@ -256,23 +256,23 @@ def attach_lifecycle_html_table(item, report):
     report.extras = extras
 
 
-def html_metric_table_header(cells):
-    cells.insert(-1, "<th>Value</th>")
-    cells.insert(-1, "<th>Unit</th>")
-
-
-def html_metric_table_row(report, cells):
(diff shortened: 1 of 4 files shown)


amd-droy (approved):
looks good. thank you.

PR #297, opened 2026-08-07
Document the unified vLLM suite

atnair-amd:
## Summary

The vLLM docs described a schema the suite no longer has: a `vllm_single` suite
name, a config path under `inference/vllm_single/<variant>/`, and a `config` +
`benchmark_params.<model>` layout. The reference page also named a config file
(`mi355x_singlenode_vllm.json`) that ships nowhere in the repo, and failed to
build.

## Change

- Add `reference/configuration-files/vllm.rst`: the current schema, the four
  distinct "backend" settings, container/Docker keys, sweep cell keys and server
  reuse, all six threshold kinds and both coverage axes, and all four metric
  namespaces (37 client / 5 gpu / 4 prom / accuracy)
- Add `how-to/run-vllm-benchmarks.rst` as the task-shaped entry point
- Delete `vllm_singlenode_mi355x.rst` — documented the removed schema, and broke
  the build (`Unknown target name: "7.0:rocm7.0_ubuntu_22.04"`)
- Fix the stale suite name and config path in `run-cvs-tests.rst`; test listing
  taken from live `cvs list vllm` output
- Ray is documented as opt-in, not required: `mp` is the default multinode
  backend, and ray's effect is to relax the pipeline-parallelism requirement
  rather than to enable multinode
- Out of scope: three code defects found while reading (`env` under
  `container.runtime.args` crashes launch; `runtime.args.shm_size` silently
  dropped; `pull_image` hardcodes bare `sudo`) — documented around, not fixed

## Tests

Docs-only; no Python in the diff. `sphinx -b html` succeeds with zero warnings
attributed to either new page (the 27 remaining are pre-existing, in
`megatron.rst`, `jax.rst`, `cluster-file.rst`, orphan toctrees, and unreachable
intersphinx inventories).


hnimra-amd:
Nice work bringing the vLLM docs in line with the unified `vllm` suite — deleting the broken MI355X page and adding the how-to + reference split is the right shape.

One accuracy fix I'd ask for before merge: the **Coverage checking** section in `vllm.rst` says axis 2 enforces a spec for every gated metric, and the client-metrics intro says gated metrics "require a threshold spec in every cell." That doesn't match the unified loader on `dev/dtni`: `vllm_config_loader._check_thresholds_cover_sweep` passes `gated_metrics=set()`, and the docstring there explicitly says metric specs are optional. At runtime, `test_metric` / `test_gpu_metric` / `test_prom_metric` also no-op when a spec is absent even with `enforce_thresholds=true`.

Suggestion: reword coverage to **cell coverage only** at load time (every sweep cell has a threshold entry; no orphan cell keys), and describe gated metrics as "asserted only when a spec is present." The optional/null-spec behavior you already document a paragraph later is the actual contract — I'd lead with that in the coverage section so readers don't think all 23 client specs are mandatory before they can flip enforcement on.

Smaller follow-up (could be this PR or a quick follow-on): `docs/install/cvs-install.rst` still points at `cvs copy-config inference/mi355x_singlenode_vllm.json`, which this PR removes from the doc set. Worth updating that install snippet to `inference/vllm/mi300x_vllm_llama31-70b_fp8_single.json` (or `cvs copy-config inference/vllm`) so new users don't hit a missing template.

solaiys (commented):
## Review — no additional blocking issues; confirming @hnimra-amd's accuracy fix

I verified this one by executing the checks rather than reading, and the documentation holds up well. The one thing I'd also hold merge for is the coverage-section inaccuracy @hnimra-amd already raised — I reached the same conclusion independently and have the evidence below.

### Confirming the existing review

**Coverage checking is cell-only — @hnimra-amd is right.** `vllm.rst:806` states:

> 2. **Metric coverage** — every cell must carry a spec for each gated metric.

That axis is a no-op for this suite. `vllm_config_loader.py:253-258` calls the shared validator with an empty set:

```python
validate_thresholds_cover_sweep(
    expected_cells=self.expected_cells(),
    thresholds=self.thresholds,
    enforce_thresholds=self.enforce_thresholds,
    gated_metrics=set(),
)
```

and in `inferencing_config_loader.validate_thresholds_cover_sweep`, `gated_keys` is built from exactly that set (L104-107), so `absent` is always `[]`, `gated_gaps` is always empty, and the "cells missing gated-metric specs" problem can never be appended. Only axis 1 (missing cells / orphaned cell keys) actually fires. Rewording as suggested — cell coverage at load time, gated metrics "asserted only when a spec is present" — matches the implementation, and leading with it makes the null-spec paragraph at L812 consistent rather than contradictory.

**The stale install snippet is real.** `docs/install/cvs-install.rst:396` still says:

```
cvs copy-config inference/mi355x_singlenode_vllm.json --output ~/my_vllm_config.json
```

`mi355x_singlenode_vllm.json` exists nowhere in the repo — not in `cvs/input/config_file/**`, and not at the merge base either, so this is pre-existing breakage rather than something this PR introduces. But since this PR is the one bringing the vLLM docs in line, it's the natural place to fix it, and a new user following the install page hits a failing command today.

### What I verified as correct

Everything else I checked came back clean:

- **The deletion is justified, exactly as described.** Parsing the removed page from the merge base reproduces the build failure verbatim:

```
docs/reference/configuration-files/vllm_singlenode_mi355x.rst:113: (ERROR/3) Unknown target name: "7.0:rocm7.0_ubuntu_22.04".
```

- **The removal is clean.** No remaining reference to `vllm_singlenode_mi355x` anywhere under `docs/`, so no dangling toctree entry or cross-reference.
- **The new pages parse without reference errors.** `vllm.rst`, `run-vllm-benchmarks.rst`, and `run-cvs-tests.rst` all come back with zero unknown-target/undefined-reference errors — notably none of the same unescaped-colon pattern that broke the deleted page.
- **All 22 toctree entries resolve** to real files (the only unresolved name is `license.md`, which is generated at build time and pre-dates this PR).
- **Every documented command resolves.** `cvs copy-config inference/vllm/mi300x_vllm_llama31-70b_fp8_single.json` and the `_distributed` variant both exist. `cvs copy-config cluster_container.json` also works — it lives under `cvs/input/cluster_file/`, and `copy_config_plugin` searches the `cluster_file` root alongside `config_file`, so the shorthand in the how-to is correct.
- **The metric reference tables are accurate.** I cross-checked every metric name in the tables against the source: all 38 `client.*`, all 5 `gpu.*`, and all 4 `prom.*` names appear in the Python code. The four `prom.*` entries match `vllm_server_metrics` exactly (`queue_time_p50_ms`, `queue_time_p95_ms`, `prefill_time_p50_ms`, `prefill_time_p95_ms`). That's the part of a reference page most likely to drift, and it doesn't.

### Blocking issues

✅ None beyond @hnimra-amd's coverage rewording, which I'd also treat as merge-blocking since it inverts the contract a reader needs before enabling `enforce_thresholds`.

docs/reference/configuration-files/vllm.rst
```diff
@@ -0,0 +1,1210 @@
...
+
+For ``min_ratio``, the ``reference`` names another metric in the same cell. If that reference is missing, ``None``, or zero, the check fails with a message naming the reason.
+
+Coverage checking
+-----------------
+
+Threshold files are validated against the sweep on two axes:
+
+1. **Cell coverage** — every sweep cell must have a threshold entry, and no threshold key may name a cell that the sweep does not produce. This catches keys left behind after a sweep edit.
+2. **Metric coverage** — every cell must carry a spec for each gated metric.
```
solaiys: Pinning @hnimra-amd's coverage point to the exact line, since I confirmed it independently.

This axis never fires for the vLLM suite. `vllm_config_loader.py:253-258` calls the shared validator with `gated_metrics=set()`, and `validate_thresholds_cover_sweep` builds `gated_keys` from exactly that set (`inferencing_config_loader.py:104-107`) — so `absent` is always `[]`, `gated_gaps` stays empty, and the "cells missing gated-metric specs" problem can never be raised. Only axis 1 is enforced.

```suggestion
2. **Metric coverage** — not enforced for this suite. A gated metric is asserted only when the cell carries a spec for it; a metric with no spec is measured and reported but never asserted.
```

As written, a reader would think all 38 `client.*` specs are mandatory before they can turn on `enforce_thresholds`, which also makes the null-spec sentence at L812 read as a contradiction rather than an elaboration.

docs/reference/configuration-files/vllm.rst
```diff
@@ -0,0 +1,1210 @@
...
+
+The ``accuracy`` key is exempt from cell-coverage checking, since it is keyed by task rather than by cell.
+
+Setting ``enforce_thresholds`` to ``false`` downgrades **both** axes to warnings and stops threshold violations from failing tests. The run still measures and records everything, which makes it the right setting for a first calibration run on new hardware.
```
solaiys: Follow-on from the L806 correction: since axis 2 never fires, "downgrades **both** axes to warnings" is describing a downgrade of something that was never enforcing. Once L806 is reworded, this is clearer as `enforce_thresholds=false` downgrading cell-coverage failures to warnings and stopping threshold violations from failing tests — which is what the loader actually does.

solaiys:
## Re-review of `88aff03f` — good to go from my side

Re-verified every claim the new commit makes against the code. **Both points are fully resolved**, and the rewrite is more accurate than a minimal fix would have been — it also caught two spots neither @hnimra-amd nor I had called out (the "record-only by default" note at the bottom of the client-metrics section, and the `Gated` column gloss).

### Coverage semantics now match the implementation

The rewritten section says cell coverage is the only load-time axis, that there is no per-metric requirement, and that assertion is decided per metric at evaluation time. All three are correct:

- **One axis.** `vllm_config_loader.py` passes `gated_metrics=set()`, so `gated_keys` is empty and the "cells missing gated-metric specs" branch in `validate_thresholds_cover_sweep` is unreachable. Only the missing/orphaned-cell checks can fire.
- **`enforce_thresholds` wording.** "Downgrades coverage problems to warnings" is exactly right — `inferencing_config_loader.py:118-120` raises `ValueError` when enforcing and calls `warnings.warn` otherwise. Dropping the old "**both** axes" phrasing was the right call.
- **`accuracy` exemption.** `NON_SWEEP_THRESHOLD_KEYS = {"accuracy"}` is subtracted from `present` at L96, so that sentence holds.
- **Per-metric assertion.** This is the part I most wanted to check, and the new text is precisely right:

```python
spec = (variant_config.thresholds.get(cell) or {}).get(full)
if spec is None:
    return
evaluate_all(actuals, {full: spec})
```

`vllm.py:480-483`. Because the guard is `spec is None`, an absent key and an explicit `null` take the identical path — so "A spec of ``null`` is the explicit way to say the same thing" is accurate rather than approximate.

### The "gate matrix" framing is real, not a hand-wave

I checked this because it's new language. `vllm_parsing.py:223-225` states that `METRIC_TIERS` membership "is seeded from `GATED_METRICS` so the report's tier partition exactly mirrors what the suite enforces: every name in `GATED_METRICS` must land in exactly one non-record tier (pinned by a unit test)", and `METRIC_TIER_ORDER` appends a `record` tier for everything else. `report/render/gate_matrix.py` renders exactly that partition. So describing the gated mark as "files it under one of the report's gate-matrix tiers" is correct, and separating that from "requires a spec" is the right distinction to draw.

### Install guide fix verified end to end

Both replacement paths exist, and all four documented keys are real, with the right shapes:

```
container.image        EXISTS   value='<changeme>'            <- correct thing to tell users to edit
paths.shared_fs        EXISTS   value='/mnt/dtni/{user-id}'
paths.models_dir       EXISTS   value='{shared_fs}/models'
params.nnodes          EXISTS   value='1'
```

Good catch replacing the flat `container_image` / `nnodes` / `data_cache_dir` names too — the shipped config is nested (`container`, `paths`, `params`, `model`, `roles`, `sweep`), so the old key names would have sent people looking for fields that don't exist.

### Build integrity

`vllm.rst` and `cvs-install.rst` both parse with zero unknown-target/undefined-reference errors, and the new `_vllm-threshold-coverage` label is defined once and referenced once, so the `:ref:` resolves.

### Blocking issues

✅ None. Good to go from my side.

I've left this as a comment rather than a formal approval so the approving click stays with you — happy to convert it to an approval if you'd like.

### One follow-up, outside this PR's diff

Now that the docs are correct, the stale statement lives in the code instead. `cvs/lib/inference/utils/vllm_parsing.py:167-168`, just above `GATED_METRICS`, says:

> Add a metric to this set the moment it becomes a pass/fail criterion -- the loader then forces a spec for it in every cell before the suite can run green.

That is the same misconception this PR just removed from the docs, and it's the most likely origin of it. For the vLLM loader nothing is forced, because `gated_metrics=set()` is passed. Worth a one-line comment fix in a follow-on so the next person doesn't re-derive the old behaviour from the source. (Note it *is* accurate for ATOM, which passes `gated_metrics=GATED_METRICS` in `atom_config_loader.py:174` — so the comment is probably best reworded to say which loaders enforce it.)

solaiys (approved):
LGTM

PR #298, opened 2026-08-07
feat(vllm): add MI300X workload config set (14 models, single + distributed)

atnair-amd:
## Summary

Adds a vLLM workload config set for MI300X: 14 models, each as a `single` and
`distributed` pair — 28 configs with 28 sibling threshold files. Purely
additive; no existing file is touched.

Every environment-specific value (`model.id`, `container.image`, models mount,
`ib_netdev`, `master_addr`) ships as `<changeme>`, so the set is committable but
not runnable until filled in. `cvs/input/config_file/inference/vllm_mi300x_workloads/README.md`
documents each field.

## Change

- Naming and flat sibling-pair layout follow `inferencex_atom_single`.
- TP is per model (4 or 8); `distributed` spans 2 nodes via PP=2, so a TP=4
  distributed variant uses 4 GPUs per node.
- Each sweep carries all three conc16 shapes (1k1k / 1k8k / 8k1k) but selects
  only `1k1k` in `sweep.runs`, so the other two are staged for later without
  needing a matching threshold cell.
- Four sparse-attention (DSA) models set `VLLM_ROCM_USE_AITER=1` plus
  `GPU_ARCHS=gfx942`: their ROCm path has no non-AITER implementation, and the
  image leaves `GPU_ARCHS` empty, which breaks aiter's JIT build.
- `enforce_thresholds` is false everywhere and all 896 threshold values are `0`,
  meaning "not yet measured". On a `max`/`max_ms` kind `0` is an impossible
  bound, so enabling enforcement before calibrating fails loudly rather than
  passing silently.
- `accuracy` blocks ship empty in both files: config-side `accuracy.tasks`
  selects tasks, threshold-side `accuracy` holds gating values keyed by task id,
  so the keys can't be pre-enumerated. Present-but-empty keeps the slot visible.
- Out of scope: threshold calibration, and the mxfp4 variants' non-MI300X targets.

## Tests

- All 28 configs load through `load_variant`: 0 failures, 0 warnings.
- Coverage-check exemption verified as discriminating — a typo'd `accuracie`
  threshold key warns, `accuracy` does not.
- `.test_venv/bin/python run_all_unittests.py` — 1150 tests, 4 failures +
  7 errors, an identical named set with and without this branch's directory
  (verified by moving it aside and re-running). All in ATOM/loader suites.
- `make fmt-check` / `make lint` are red on pristine `dev/dtni` already
  (29 files / 20 errors, measured against a clean `git archive` extract);
  identical with this branch applied. This diff is JSON and Markdown only —
  zero `.py` files.
- Rebased onto the rewritten `dev/dtni` after an upstream force-push. The
  metric constants this set mirrors (`GATED_METRICS`, `GPU_METRICS`,
  `PROM_METRICS`) hash identically across that rewrite, and regenerating all
  56 files against the new base is a no-op (`git status` clean).
- Hardware validation is in progress and **not** complete: 3 variants have
  passed end-to-end (llama33-70b single + distributed, deepseek-v4-flash
  single), 23 are unrun. No model has failed for model/hardware reasons.
  Draft until that sweep finishes.
- Those runs executed on **MI325X** nodes, not MI300X. Both are CDNA3 and
  report `gfx942`, so the configs are valid for either, but this set has not
  been exercised on MI300X hardware.


hnimra-amd:
Solid scaffold set — flat sibling pairs, explicit `threshold_json`, per-model TP, and the DSA/AITER + fp8 kv-cache fixes for DeepSeek V4 / GLM look well reasoned from the commit history. README guidance on `ib_netdev` needing a routable IPv4 is especially useful.

A few items before I'd merge:

1. **Hardware validation gap (called out in the PR body).** Only 3/26 variants have passed E2E and those runs were on MI325X, not MI300X. The directory/name says MI300X — I'd either keep this as draft until the MI300X sweep finishes, or add a prominent README note that validation is MI325X-only so far (gfx942 parity isn't the same as sign-off on MI300X fleet configs).

2. **`num_prompts: 320` vs `3200` default.** All workload configs use 320 while the shipped `inference/vllm/` examples and schema default are 3200. If 320 is intentional for a faster characterization pass, worth one sentence in the README (and maybe `_comment`) so operators don't assume they're getting a full-length benchmark.

3. **Threshold README wording vs loader behavior.** The README implies filling all 32 metric slots is required for the coverage check. Unified `vllm_config_loader` only enforces **cell** coverage (`gated_metrics=set()`); the full placeholder grid is good practice for future calibration but not loader-mandatory. Minor tweak to avoid over-stating the requirement — aligns with what #297's reference page should say once merged.

4. **Cross-link docs.** Once #297 lands, a one-liner in the README pointing at the vLLM reference/how-to for threshold kinds and multinode setup would reduce duplication here.

No issues spotted in the JSON structure itself (sibling threshold paths, cell keys, AITER env on the four DSA models). Happy to re-review after the MI300X hardware sweep.

atnair-amd:
Thanks — addressed 2, 3 and 4 in 0a3554fc.

**2. `num_prompts: 320`.** Intentional, and you're right that nothing said so.
Confirmed the gap: all 28 configs here use `320`, while both shipped examples
(`mi300x_vllm_llama31-70b_fp8_{single,distributed}.json`) and the schema
default at `vllm_config_loader.py:172` are `3200`. It's a characterization
pass — enough to shake out topology, AITER and kv-cache settings on new
hardware at roughly a tenth the wall-clock. Now documented in the README and
in each config's `_comment`, with the reason and the instruction to raise it
to `3200` before quoting numbers meant to line up with the examples.

**3. Threshold wording.** I've made the non-mandatory part explicit, but I
don't think the README actually claimed otherwise — the only coverage sentence
was about *sweep cells* ("add the matching cell key to the threshold file — the
coverage check compares the two"), which is what the loader does. The 32-slot
table was introduced as what each file *carries*. Either way the ambiguity
wasn't worth defending, so the section now states outright that the grid is
calibration convenience and that an absent metric spec means "don't gate this
metric".

Worth flagging for #297 though, since it's the same distinction: the reference
page's Coverage checking section currently says "**Metric coverage** — every
cell must carry a spec for each gated metric." That reads as unconditional, but
metric coverage only engages when the caller passes `gated_metrics` — the vLLM
loader passes `set()` (`vllm_config_loader.py:253-257`), so for vLLM that axis
never fires. ATOM passes a real set, so it does there. Might be worth scoping
that bullet per-suite before #297 lands.

**4. Cross-link.** Added a See also pointing at
`docs/reference/configuration-files/vllm.rst` and
`docs/how-to/run-vllm-benchmarks.rst`. Note this is a forward reference until
#297 merges — happy to reorder if you'd rather that PR land first.

**1. Hardware validation.** Leaving as-is for now; that's a call for @atnair on
how to sequence against the MI300X sweep. The remaining variants are blocked on
jumphost access at the moment regardless.

PR #299, opened 2026-08-10
Feature/rccl install

ahskabir:
## Motivation

This feature will allow an user to install RCCL test and optionally build RCCL library as well. Both `rccl` library and `rccl-tests` will be built and installed from the rocm-systems repository:

## Technical Details

The rocm-systems repository: https://github.com/rocm/rocm-systems. As for RCCL library is concerned, user has either of the two choices: Use library from within the rocm installation /opt/rocm/lib Or build it from rocm-systems
The option will be provided whether or not user wants to build RCCL library. Note that rccl-tests build requires installation of RCCL library plus installation of OpenMPI. The configuration file for this test suite will ask for: Path to OpenMPI installation

## Test Plan

Installation is tested. Both RCCL library and rccl-tests were built. After built RCCL test was run and performance of all_reduce was validated.

## Test Result
all_reduce perf result: The RCCL test completed successfully and achieved the expected bandwidth (Size: 16GB → BusBW: 360 GB/s.

## Submission Checklist


cvs/tests/rccl/install_rccl_tools.py
```diff
@@ -0,0 +1,702 @@
+'''
```
cijohnson, line 1: Can we  rename it to cvs/tests/rccl/install_rccl_tests.py

cvs/input/config_file/rccl/rccl_config.json
```diff
@@ -1,5 +1,34 @@
 {
     "rccl": {
+        "installation_params": {
```
cijohnson, line 3: should we have separate config file named rccl_tests_install_config.json?

cvs/input/config_file/rccl/rccl_config.json
```diff
@@ -1,5 +1,34 @@
...
+            "nfs_install": "True",
+            "_comment_nfs_install": "If nfs_install is True, rccl-tests and rccl library (optional) will be installed in NFS. Otherwise, all nodes will have local installation",
+            "rccl_lib_install": "True",
+            "_comment_rccl_lib_install": "Set True to sparse-clone projects/rccl from rocm-systems and build RCCL for rccl-tests (e.g. when bundled ROCm RCCL mismatches rccl-tests). rccl-tests then builds with CUSTOM_RCCL_LIB.",
+            "rccl_lib_build_timeout": "10800",
+            "_comment_rccl_lib_build_timeout": "SSH exec timeout in seconds for RCCL library source build. Some rccl lib tags take long time to build, thus the high timeout value, this upper bound doesn't have effect on actual build time",
+            "rccl_lib_install_dir": "<changeme>",
```
speriaswamy-amd: Blocking: These <changeme> values break rccl_perf and rccl_regression. Both fixtures resolve the whole rccl object through resolve_test_config_placeholders, which sys.exit(1)s on the first <changeme> — so adding installation_params here aborts both suites on the shipped default, even for users who never run the installer. Verified: this branch exits 1, main resolves clean. Please move installation_params to its own config file, or have the perf suites resolve only the sub-keys they use.

cvs/tests/rccl/install_rccl_tools.py
```diff
@@ -0,0 +1,702 @@
...
+        f"export LD_LIBRARY_PATH=${{OMPI_PREFIX}}/lib:${{ROCM_PREFIX}}/lib:${{LD_LIBRARY_PATH:-}}; "
+    )
+
+
+def _git_clone_source(hdl, repository, dest_dir, sparse_path=None, git_tag=""):
+    """
+    Clone repository at dest_dir. When sparse_path is set, use partial clone +
+    sparse-checkout for a single rocm-systems project path.
+    """
+    hdl.exec(f"rm -rf {dest_dir}", timeout=60)
```
speriaswamy-amd: Blocking: This rm -rf is unguarded and runs on a user-supplied path. Combined with the /lib stripping in _rccl_install_prefix, setting rccl_lib_install_dir=/opt/rocm/lib (which the docstring on line 278 explicitly suggests) produces rm -rf /opt/rocm on every node. Please confine the delete to a subdirectory the installer owns, and reject system prefixes.

cvs/tests/rccl/install_rccl_tools.py
```diff
@@ -0,0 +1,702 @@
...
+        rccl_repository,
+        repo_dir,
+        sparse_path=sparse_path,
+        git_tag=rccl_git_tag,
+    )
+
+    rccl_build_dir = f"{rccl_project_dir}/build"
+    hdl.exec(
+        f"bash -c '{build_env}cd {rccl_project_dir} && mkdir -p build && cd build && cmake .. && make -j $(nproc)'",
+        timeout=14400,
```
speriaswamy-amd: Should Fix: Timeout is hardcoded to 14400 while the new rccl_lib_build_timeout config key says 10800 and is never read by any code. Either wire the key through here or drop it from the sample config.

cvs/tests/rccl/install_rccl_tools.py
```diff
@@ -0,0 +1,702 @@
...
+    nfs_install = config_dict["nfs_install"]
+    rccl_lib_install = config_dict["rccl_lib_install"]
+
+    rccl_lib_install_dir = config_dict["rccl_lib_install_dir"].rstrip('/')
+    rccl_tests_install_dir = config_dict["rccl_tests_install_dir"].rstrip('/')
+    rccl_repository = config_dict["rccl_repository"]
+    ompi_install_dir = config_dict["ompi_install_dir"].rstrip('/')
+
+    # Resolve the orchestrator handle
+    if nfs_install == "True":
```
speriaswamy-amd: Should Fix: == "True" is case- and whitespace-sensitive, so "true" silently selects per-node installs instead of NFS. This file already has _config_bool_str for this; please use it here and on lines 650 and 672.

cvs/tests/rccl/install_rccl_tools.py
```diff
@@ -0,0 +1,702 @@
...
+    log.info("%s", cluster_dict)
+    env_vars = cluster_dict.get("env_vars")
+    node_list = list(cluster_dict['node_dict'].keys())
+    if len(node_list) < 2:
+        raise ValueError('At least 2 nodes are required to run this test')
+    if len(node_list) % 2 != 0:
+        log.info(
+            f'Odd number of nodes ({len(node_list)}) detected; popping last node from the cluster to make the count even'
+        )
+        node_list.pop()
```
speriaswamy-amd: Should Fix: Popping the last node means node 3 of a 3-node cluster never gets rccl-tests when nfs_install=False, and the failure only surfaces later in rccl_perf. An installer also shouldn't require 2 nodes at all. This is copied from install_ibperf_tools.py, but it's worth breaking the pattern in a new file: install on every node in node_dict.

cvs/tests/rccl/install_rccl_tools.py
```diff
@@ -0,0 +1,702 @@
...
+    _git_clone_source(
+        hdl,
+        rccl_repository,
+        repo_dir,
+        sparse_path=sparse_path,
+        git_tag=rccl_git_tag,
+    )
+
+    rccl_build_dir = f"{rccl_project_dir}/build"
+    hdl.exec(
```
speriaswamy-amd: Should Fix: No error check on the library build, unlike the tests build on line 598. A cmake or compile failure surfaces as "library not found" from _resolve_rccl_shared_lib instead of the real compiler error. Please apply the same \berror: scan here.

cvs/tests/rccl/install_rccl_tools.py
```diff
@@ -0,0 +1,702 @@
...
+    rccl_lib_install_dir may point at the install root or at .../lib when reusing
+    the ROCm layout.
+    """
+    install_dir = config_dict["rccl_lib_install_dir"].rstrip("/")
+    if install_dir.endswith("/lib"):
+        return install_dir[: -len("/lib")]
+    return install_dir
+
+
+def _rccl_source_layout(config_dict):
```
speriaswamy-amd: Nit: this function doesn't seem to be used anywhere

cvs/tests/rccl/install_rccl_tools.py
```diff
@@ -0,0 +1,702 @@
...
+        f"\\( -name librccl.so -o -name \"libnccl.so\" -o -name \"libnccl.so.*\" \\) "
+        f"2>/dev/null; done | head -1'",
+        timeout=60,
+    )
+    for node, output in out_dict.items():
+        lib_path = output.strip().splitlines()[0] if output.strip() else ""
+        if not lib_path:
+            return ""
+        log.info("Node %s: resolved RCCL shared library: %s", node, lib_path)
+    return next(iter(out_dict.values())).strip().splitlines()[0]
```
speriaswamy-amd: Nit: next(iter(...)) raises a bare StopIteration when out_dict is empty (no nodes reachable). The loop above also logs per-node but the return only reflects the first node — worth simplifying so the intent is explicit.

cvs/tests/rccl/install_rccl_tools.py
```diff
@@ -0,0 +1,702 @@
...
+    """
+    node_list = list(cluster_dict['node_dict'].keys())
+    env_vars = cluster_dict.get("env_vars")
+    head_node = node_list[0]
+    shdl = Pssh(log, [head_node], user=cluster_dict['username'], pkey=cluster_dict['priv_key_file'], env_vars=env_vars)
+    return shdl
+
+
+@pytest.fixture(scope="module")
+def vpc_node_list(cluster_dict):
```
speriaswamy-amd: Nit: This function is not used anywhere in this module, could be dropped

cijohnson (changes requested):

PR #300, opened 2026-08-10
[JAX][Training] Add JAX MaxText single-node & distributed training suites (sweeps, metrics, thresholds, reporting)

solaiys:
**Summary**
Adds config-driven JAX MaxText pre-training validation suites for AMD Instinct GPUs, covering single-node and multi-node (distributed) runs. Each run drives a MaxText training job inside a container, parses the training log, and gates the run on performance and correctness metrics with a per-(sweep, metric) PASS/FAIL HTML report and console summary.

**Suites**

- cvs run jaxmaxtext_single — single-node (no RDMA stage).
- cvs run jaxmaxtext_distributed — multi-node (adds the RDMA setup stage).

Both share their implementations from _common.py; sweep parametrization and fixtures/hooks live in conftest.py. The mode (single/distributed) is reflected in the suite filename, the metric-results HTML title, and the loss-curve titles/artifacts.

```
cvs run jaxmaxtext_distributed \
  --cluster_file ./p3_2n_cluster.json \
  --config_file cvs/input/config_file/training/jaxmaxtext/mi325x_jaxmaxtext_llama-3.3-70b_distributed.json \
  --html ./logs/jaxmaxtext_distributed.html --self-contained-html
```
**Config-driven design**
Typed pydantic config loader (training_config_loader.py) with placeholder substitution ({user-id}, {shared_fs}, {paths.*}).
One config + sibling threshold file per arch/mode (MI300X single/distributed, MI325X distributed).
Sweeps: one full training run per entry (training.sweeps + enabled_sweep_list); the sweep name is the threshold cell key. Currently precision is the swept dimension (BF16, FP8). Per-sweep MaxText overrides (per_device_batch_size, max_target_length, precision).
coordinator_ip: auto resolves to the first cluster node; gpus_per_node is config-driven (no assumed 8-GPU topology).
FP8 recipe by arch: nanoo_fp8 (MI300X/MI325X, CDNA3) vs fp8 (MI355X/MI350X, CDNA4).

**Metrics (validation-sheet rows)**

- Throughput: tokens_per_sec_per_gpu, tokens_per_sec_total, tflops_per_sec_per_gpu.
- Step time: mean / p50 / p95 (ms) for tail-latency/jitter.
- Scaling efficiency % vs a 1-node baseline.
- Convergence: steps_to_target / time_to_target_seconds.
- Loss curve (row 32): per-sweep PNG + least-squares downward-trend gate.
- Gating is threshold-driven (enforce_thresholds), per (sweep, metric): PASS / FAIL / N/A (metric not produced) / RECORD. Threshold kinds include min, max, max_ms, within, min_tok_s, min_ratio, and info (always-pass, record-only with a placeholder value to calibrate later).

**Reporting**

- Results table with per-metric PASS/FAIL.
- Single consolidated metric_results.html (Sweep | Metric | Expected | Actual | Unit | Status) linked from every metric row, plus each test's own Full Log link.
- Per-sweep loss-curve PNGs.
- Aggregated failure summary in the pytest final summary via globals.error_list / fail_test / update_test_result.

**Training-log error detection**

- Config-driven training.error_patterns (name → regex) scanned per node during polling; a match fails that sweep's test_training_run with the matched signature. Defaults cover NCCL, GPU HW faults, JAX/assertion errors, ROCm init, Python fatal, TF coordination, OOM (RESOURCE_EXHAUSTED), and segfault. Falls back to built-in defaults when omitted.
- Lingering ranks are reaped on failure/timeout (stop_training, self-safe pkill) so the next sweep does not launch on top of orphans.

**Files**

- Suites: cvs/tests/training/jaxmaxtext/{jaxmaxtext_single.py, jaxmaxtext_distributed.py, _common.py, conftest.py, README.md}.
- Lib: cvs/lib/training/jaxmaxtext/{jaxmaxtext_training_lib.py, utils/maxtext_parsing.py, utils/training_config_loader.py, utils/loss_curve.py} + unittests/.
- Shared: cvs/lib/utils/verdict.py (adds the info threshold kind).
- Inputs: cvs/input/config_file/training/jaxmaxtext/ configs + thresholds + README.md.
- requirements.txt: adds matplotlib (headless loss-curve rendering).
- 

**Tests**
Unit tests for maxtext_parsing, training_config_loader, loss_curve, and MaxTextTrainingJob (mocked orchestrator): parsing, convergence, loss-curve sampling/slope, scaling efficiency, coordinator resolution, gpus_per_node, config-driven error patterns, and self-safe stop_training.

**Test plan**

- cvs run jaxmaxtext_distributed on a 2-node MI325X cluster (BF16 + FP8 sweeps) — verify per-sweep rows, thresholds, loss curves, and metric-results HTML.
- cvs run jaxmaxtext_single on MI300X — verify no RDMA stage and correct rows.
- python -m unittest discover -s cvs/lib/training/jaxmaxtext/unittests -p "test_*.py" — all pass.

**Notes**
FP8 nanoo_fp8/fp8 and threshold values are placeholders; calibrate against measured hardware numbers.
Per-sweep steps must match training.steps (labels vs the actual completion check).
The HTML report Environment "Config/Cluster File" hyperlink fix is submitted separately (targets cvs/lib/report_plugins.py).


solaiys:
[jaxmaxtext_distributed_2026-08-10T183653.zip](https://github.com/user-attachments/files/30910509/jaxmaxtext_distributed_2026-08-10T183653.zip)
Attached the 2Node training results and html report.

atnair-amd:
### Code review

Found 2 issues:

1. Distributed configs ship concrete cluster-specific RDMA/NIC device names (`rdma0..rdma7`, `eno0`) as defaults instead of `<changeme>` ([AGENTS.md](https://github.com/ROCm/cvs/blob/573227ba2b507495c434e53372de516cb603b805/AGENTS.md) says "Use `<changeme>` in sample/default JSON values for any field the user must customize; do not ship cluster-specific values as defaults", and "**Never hardcode cluster-specific values** in test code (IP addresses, hostnames, interface names)"). The sibling training suites use `"<changeme>"` with an `_example_*` key for exactly these fields — see [torchtitan](https://github.com/ROCm/cvs/blob/573227ba2b507495c434e53372de516cb603b805/cvs/input/config_file/training/torchtitan/mi3xx_torchtitan_llama_distributed.json#L16-L20), and the same in `training/jax/` and `training/megatron/`. As shipped, a user on different hardware silently runs with the wrong NIC/RDMA devices rather than hitting the `<changeme>` startup hard-exit.

https://github.com/ROCm/cvs/blob/573227ba2b507495c434e53372de516cb603b805/cvs/input/config_file/training/jaxmaxtext/mi325x_jaxmaxtext_llama-3.3-70b_distributed.json#L147-L152

(same block at `mi300x_jaxmaxtext_llama-3.3-70b_distributed.json` L149-L152)

2. `validate_thresholds_cover_training()` is never called from the suite, so a sweep-name mismatch silently disables gating. It and its `cell_key()`/`expected_cells()` helpers are implemented and unit-tested, but `grep` finds no caller in `conftest.py`, `_common.py`, `jaxmaxtext_single.py`, or `jaxmaxtext_distributed.py` — only `unittests/test_training_config_loader.py`. When a `sweeps[].name` does not match a threshold key, `metric()` takes the `spec is None` path and emits a non-gating `RECORD` row with no error or warning, so the suite reports green while gating nothing. Separately, the validator's key format (`STEPS=..,BATCH=..,SEQLEN=..,NODES=..`) does not match the sweep-name keys the shipped threshold files actually use (`NNODES=2,STEPS=30,PRECISION=BF16,BATCH=3,GBS=48,SEQLEN=8192`), so wiring it in as-is would report every real key as both missing and extra.

https://github.com/ROCm/cvs/blob/573227ba2b507495c434e53372de516cb603b805/cvs/lib/training/jaxmaxtext/utils/training_config_loader.py#L152-L173

https://github.com/ROCm/cvs/blob/573227ba2b507495c434e53372de516cb603b805/cvs/tests/training/jaxmaxtext/_common.py#L322-L325

cvs/lib/training/jaxmaxtext/utils/training_config_loader.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        t = self.training
+        mc = t.maxtext_config
+        batch = mc.get("per_device_batch_size", "-")
+        seqlen = mc.get("max_target_length", "-")
+        return f"STEPS={t.steps},BATCH={batch},SEQLEN={seqlen},NODES={num_nodes}"
+
+    def expected_cells(self, num_nodes=1):
+        """The single cell this training config produces."""
+        return [self.cell_key(num_nodes=num_nodes)]
```
amd-droy: Runtime lookup in _common.py uses sweep name as the threshold key (e.g. NNODES=2,STEPS=30,PRECISION=BF16,...), not cell_key(). If you wire the validator using expected_cells() as-is, it will false-fail every real config. Align expected_cells() with [s.name for s in self.enabled_sweeps()] (or drop cell_key()).

cvs/lib/training/jaxmaxtext/utils/maxtext_parsing.py
```diff
@@ -0,0 +1,386 @@
...
+    ("tokens_per_sec_per_gpu", "tok/s/GPU"),
+    ("tokens_per_sec_total", "tok/s total"),
+    ("scaling_efficiency_pct", "%"),
+    ("step_time_seconds", "s/step"),
+    ("step_time_mean_ms", "ms/step"),
+    ("step_time_p50_ms", "ms/step"),
+    ("step_time_p95_ms", "ms/step"),
+    ("final_loss", "loss"),
+    ("loss_decreased", "bool"),
+    ("eval_loss", "loss"),
```
amd-droy, line 34: these two not in GATED:
("final_loss", "loss"),
("loss_decreased", "bool"),
For training, do you think these should be verified against a threshold?

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
```diff
@@ -0,0 +1,501 @@
...
+# out of the box.
+_TRAINING_ERR_PATTERNS = {
+    'NCCL ERROR': r'NCCL ERROR|NCCL timeout|local work queue catastrophic error',
+    'GPU HW ERROR': r'HW Exception by GPU|GPU Hang|Uncorrectable error|GPU Reset',
+    'AssertionError': r'AssertionError|ValueError:|JaxStackTrace|During handling of the above exception|triggered the following exception',
+    'rocm Err': r'FAILED_PRECONDITION: No visible GPU devices|failed call to hipInit: HIP_ERROR_NoDevice|librocm reported version is: NOT_FOUND',
+    'python err': r'ModuleNotFoundError: No module named|Fatal Python error:',
+    'tensorflow': r'tensorflow.CoordinationServiceError|tensorflow.BarrierError|CoordinationServiceError',
+    'resource': r'RESOURCE_EXHAUSTED: Out of memory|failed: RESOURCE_EXHAUSTED',
+}
```
amd-droy, line 57: segfault - error pattern is missing. You can check with legacy tests if they have this.

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
```diff
@@ -0,0 +1,501 @@
...
+
+_COMPLETION_RE = re.compile(r'completed step:\s*(\d+)')
```
amd-droy: _COMPLETION_RE - not used. You can remove this.

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
```diff
@@ -0,0 +1,501 @@
...
+            text = text if isinstance(text, str) else (text or {}).get("output", "")
+            text = text or ""
+            i = node_of.get(host, "?")
+            if _NAN_INF_RE.search(text):
+                raise RuntimeError(f"NaN/Inf in training metrics on {host} (node {i}): {text[-500:]}")
+            for err_name, err_pattern in self.error_patterns.items():
+                if not err_pattern:
+                    continue
+                if re.search(err_pattern, text, re.I):
+                    raise RuntimeError(f"Training error '{err_name}' on {host} (node {i}): {text[-500:]}")
```
amd-droy, line 444: is this covered - verify_dmesg_for_errors()?

cvs/tests/training/jaxmaxtext/_common.py
```diff
@@ -0,0 +1,488 @@
...
+        try:
+            job.stop_training()
+        except Exception:  # noqa: BLE001
+            pass
+        pytest.fail(f"training run failed for sweep '{sweep_name}': {e}")
+
+    results["training.wall_time_seconds"] = wall_time
+    results["training.convergence_steps"] = variant_config.training.steps
+    results["training.convergence_wall_time"] = wall_time
+
```
amd-droy, line 264: Not in TRAINING_METRICS; never asserted.

cvs/tests/training/jaxmaxtext/conftest.py
```diff
@@ -0,0 +1,196 @@
...
+        d = json.load(fp)
+    return resolve_cluster_config_placeholders(d)
+
+
+@pytest.fixture(scope="module")
+def variant_config(pytestconfig, cluster_dict):
+    config_file = pytestconfig.getoption("config_file")
+    if not config_file:
+        pytest.fail("--config_file is required")
+    return load_training_variant(config_file, cluster_dict)
```
amd-droy: Both jaxmaxtext_single and jaxmaxtext_distributed share this conftest, but variant_config only calls load_training_variant(config_file, cluster_dict) and never checks that training.distributed matches the suite being run. That allows a bad pairing — e.g. cvs run jaxmaxtext_single with a distributed config (skips RDMA, still launches multi-node JAX) or jaxmaxtext_distributed with a single-node config — to start and fail late with confusing errors.

solaiys:
> ### Code review
> Found 2 issues:
> 
> 1. Distributed configs ship concrete cluster-specific RDMA/NIC device names (`rdma0..rdma7`, `eno0`) as defaults instead of `<changeme>` ([AGENTS.md]
> 
> 2. `validate_thresholds_cover_training()` is never called from the suite, so a sweep-name mismatch silently disables gating. 

**FIXED.**

cvs/tests/training/jaxmaxtext/_common.py
solaiys: removed.

cvs/tests/training/jaxmaxtext/conftest.py
solaiys: Done.

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
solaiys: Added now. It was not covered.

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
solaiys: Removed.

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
solaiys: It was already added from the config file. But in this file fallback list of patterns if config file doesn't have the patterns.
Now here also added.

cvs/lib/training/jaxmaxtext/utils/maxtext_parsing.py
solaiys: Yes. For now I have all those metrics as info only. Thee are few others need to be added as gated.
I wanted it to be controlled from the threshold json file itself using the kind value.

if kind = "info"  Dont gate the metric for FAIL. mark PASS always.
Else, gate the metric for the FAIL/PASS

For now, i have added those 2 metrics in to gated.
I will submit another PR, to enable the other metrics for gating.

cvs/lib/training/jaxmaxtext/utils/training_config_loader.py
solaiys: Done. Thanks.

amd-droy (approved):
looks good to me. Thanks Saravanan.

PR #301, opened 2026-08-10
fix(html-report): handle the missing link in the html ENV table.

solaiys:
**Handles the following 2 cases:**
1. Config hyperlink — _update_environment_config_links previously decided which Environment row to link by checking whether the filename contained the substring "cluster"/"config". If there is no substring "cluster| config" the it will not show the link.

2. The argument parsing only handled the space-separated form (--cluster_file value), but not the equals form (--cluster_file=value), so the scan never matched and defaulted to "Not specified".

  Added a shared helper cli_option_value(name, argv=None) in cvs/lib/report_plugins.py that
  handles both --opt value and --opt=value, and used it in the two places that parsed sys.argv.

**Both cases handled correctly (verified functionally):**

- cli_option_value("--config_file", ...) returns the value for both --config_file=./a/b.json (equals) and --config_file ./c.json (space) forms.
- Environment table linking is now keyed by the recorded kind (cluster/config), not a filename substring — so a config whose name has no "config" token (e.g. mi325x_..._distributed.json) still gets a hyperlink. Confirmed: both Config File and Cluster File render \<a href=...\>  cells .
- The config_/cluster_ bundle prefixes are retained (bundle hygiene), and _config_files now stores {"rel", "kind"} — all readers/writers (copy_config_files_to_bundle, _update_environment_config_links, the truthiness check) are consistent; no other consumer expects the old string form.

### DIFF
--- a/cvs/lib/report_plugins.py
+++ b/cvs/lib/report_plugins.py
@@ -18,6 +18,26 @@ from cvs.lib import globals
 
 log = globals.log
 
+
+def cli_option_value(name, argv=None):
+    """Return the value passed for CLI option `name`, or None if absent.
+
+    Supports both `--name value` (space-separated) and `--name=value` (equals)
+    forms. Needed because callers invoke the suite either way (e.g. via
+    `cvs run ... --config_file=...`), and a plain `arg == name` scan silently
+    misses the equals form -- which is why the HTML Environment table showed
+    "Not specified" and the config files were never bundled/linked.
+    """
+    argv = sys.argv if argv is None else argv
+    prefix = name + "="
+    for i, arg in enumerate(argv):
+        if arg == name and i + 1 < len(argv):
+            return argv[i + 1]
+        if arg.startswith(prefix):
+            return arg[len(prefix):]
+    return None
+
+
 REPORT_STYLE_OVERRIDES = """<style>
     .collapse { display: none !important; }
     .col-result:hover::after { content: none !important; }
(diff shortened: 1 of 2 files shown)


cvs/lib/report_plugins.py
```diff
@@ -261,15 +285,12 @@ def copy_config_files_to_bundle(self, cluster_file_path, config_file_path):
...
-
-        for i, arg in enumerate(sys.argv):
-            if arg == "--cluster_file" and i + 1 < len(sys.argv):
-                cluster_file_path = str(Path(sys.argv[i + 1]).resolve())
-            elif arg == "--config_file" and i + 1 < len(sys.argv):
-                config_file_path = str(Path(sys.argv[i + 1]).resolve())
+        # Parse command line arguments to get file paths.
+        # Handles both `--opt value` and `--opt=value` forms.
+        cluster_arg = cli_option_value("--cluster_file")
+        config_arg = cli_option_value("--config_file")
```
amd-droy, line 291: I think, pytest has already parsed --cluster_file / --config_file into that config object. these are redundant.

solaiys: This fix is for the html report Environment table section.
use the below commands in your execution, you will see the issue.

1. cvs run <test_suite> --cluster_file=<filename> --config_file=<filename> .....
2. cvs run <test_suite> --cluster_file <filename without **cluster** word> --config_file <config filename without **config** word> .....

**For  #1st case :** cluster file and config file is shown as **"Not Specified"**
image

**For #2nd case:**   **hyperlink is missed in config file due to there is no config word in that file..**

image

amd-droy (approved):
looks good to me.

PR #302, opened 2026-08-10
reformatted to pass ruff formatter check in the CI

urtiwari:
reformatted to pass ruff formatter check in the CI

### DIFF
--- a/cvs/lib/preflight/node_smoke.py
+++ b/cvs/lib/preflight/node_smoke.py
@@ -260,7 +260,6 @@ class NodeSmokeCheck(PreflightCheck):
         self.master_port = int(get_nested_config(cfg, "node_smoke", "master_port", 1234))
         self.ssh_timeout = int(get_nested_config(cfg, "node_smoke", "ssh_timeout", 300))
 
-        get_nested_config(cfg, "reporting", "artifacts_root_dir", "/tmp/preflight")
         self.dump_path = _resolve_dump_path(cfg)
 
         rdma_ifaces = node_check.get("rdma_interfaces") or []
--- a/cvs/lib/preflight/node_smoke.py
+++ b/cvs/lib/preflight/node_smoke.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 import json
 import re
 import shlex
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional
 
 from cvs.lib.preflight.base import PreflightCheck
 
@@ -260,7 +260,7 @@ class NodeSmokeCheck(PreflightCheck):
         self.master_port = int(get_nested_config(cfg, "node_smoke", "master_port", 1234))
         self.ssh_timeout = int(get_nested_config(cfg, "node_smoke", "ssh_timeout", 300))
 
-        artifacts_root = get_nested_config(cfg, "reporting", "artifacts_root_dir", "/tmp/preflight")
+        get_nested_config(cfg, "reporting", "artifacts_root_dir", "/tmp/preflight")
         self.dump_path = _resolve_dump_path(cfg)
 
         rdma_ifaces = node_check.get("rdma_interfaces") or []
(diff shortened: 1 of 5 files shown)


cvs/lib/preflight/node_smoke.py
```diff
@@ -260,7 +260,7 @@ def _load_settings(self):
         self.master_port = int(get_nested_config(cfg, "node_smoke", "master_port", 1234))
         self.ssh_timeout = int(get_nested_config(cfg, "node_smoke", "ssh_timeout", 300))
 
-        artifacts_root = get_nested_config(cfg, "reporting", "artifacts_root_dir", "/tmp/preflight")
+        get_nested_config(cfg, "reporting", "artifacts_root_dir", "/tmp/preflight")
```
solaiys: This line itself is not required i think.
If so, could you please remove it.

solaiys:
## Review summary — no blocking issues

Reviewed all 5 files. This is behavior-preserving as intended. I paid particular attention to the two places where a pure reformat can silently change semantics, and verified both rather than eyeballing them.

### 1. The `f` prefix removal in `report.py` (`html_out = f"""` → `html_out = """`)

This one is worth calling out because the diff shape *can* be dangerous, but here it is **safe**. Verified four ways:

- The string literal contains **zero brace characters** — no `{`, no `}`, no `{{`, no `}}`. It is 439 characters of static markup, byte-identical on both sides.
- Evaluated as an f-string vs. a plain string, the two produce equal values.
- End-to-end: running `_generate_node_smoke_html` from the base commit and from this PR head against the same failing-node input produces **byte-identical HTML** (same SHA-256, 979 chars).
- Ruff itself reports the original line as `F541` ("f-string without any placeholders → Remove extraneous `f` prefix"), which is one of ruff's *safe* fixes.

For future reference, the case where this change **does** break things: if the string had contained `{{` / `}}` escapes — common in HTML with embedded CSS, e.g. `<style>body {{ margin: 0 }}</style>` — dropping the `f` converts those escapes into literal doubled braces and corrupts the output. Ruff's autofix unescapes them correctly; a hand-applied removal would not. Not applicable here since the block has no braces at all.

### 2. Shell-command builders in `primus_setup.py`

These reflows join multi-line f-strings that build shell commands, which is where a joined line can silently lose or gain a space. I executed every affected builder on both versions across all branches — `recurse_submodules` both ways, `force_reclone` both ways, and all three `pip_install_mode` values (`skip` / `requirements` / `minimal`) — and **every generated command string is byte-identical**.

### 3. Other changes

- `Tuple` dropped from the `typing` import in `node_smoke.py`: no remaining references in the file.
- The `x or None` and `list(... or [...])` reflows are parenthesization only; precedence is unchanged.
- All 5 files compile and import cleanly; the 27 unit tests in the two changed test files pass.
- `ruff format --check` and `ruff check` both pass on the 5 changed files.

---

### Heads-up: this won't turn CI green yet

`.github/workflows/ci.yml` runs `make test`, which chains `ut` → `installtest` → `build`, and `build: fmt-check lint sdist`. Both gates run **repo-wide** (`ruff format --check .` and `ruff check .`), not just on changed files.

After this PR, on top of `dev/dtni`:

- **24 files** still fail `ruff format --check` (baseline was 29)
- **17** `ruff check` errors remain (baseline was 20): 11 `F401`, 4 `E402`, 1 `F841`, 1 `F541`

So this is genuine progress on 5 files, but the CI check on this PR will still be red. Most of the remainder is in the sglang / inference tree, and `make fmt` + `make lint-fix` from the repo root would clear the bulk of it mechanically.

Two leftovers are the same rules this PR fixes instances of elsewhere:

- `cvs/lib/utils/unittests/test_gpu.py:1141` — `F541`
- `cvs/lib/inference/sglang/sglang_distributed_lib.py:464` — `F841` (`pp` assigned but never used)

Happy to approve either way — just flagging it so the CI status isn't a surprise, and so we can decide whether to finish the sweep in this PR or a follow-up.

### Minor

`cvs/lib/preflight/node_smoke.py:263` — see my inline comment. Confirming the analysis: `get_nested_config` is a pure nested-dict lookup with no side effects, and `artifacts_root` was a local in `_load_settings` that nothing in that method used (the one at lines 61–62 is a separate local inside `_resolve_dump_path`). So the leftover bare call computes a value and discards it — safe to delete the whole line, no `NameError` risk. Deleting the statement is the right resolution for `F841` rather than only dropping the assignment.

solaiys:
The other ruff format check is on the other files in dev/dtni.

After removing the unused get_nested_config call line, we can merge this patch.

solaiys (approved):
LGTM

PR #306, opened 2026-08-11
Torchtitan dtni

amd-rthummal:
This PR introduces a complete Torchtitan pre-training validation suite for MI355X GPUs, covering single-node and distributed (multi-node) runs. The suite drives Torchtitan training jobs inside a container, parses training logs, and gates results against configurable per-combo performance and correctness thresholds with a linked HTML report.

**How to run**
**Single-node (Llama 3.1 8B)**

cvs run torchtitan_single \
  --cluster_file ./cluster.json \
  --config_file cvs/input/config_file/training/torchtitan/mi355_llama3_1_8b_single_config.json \
  --html ./logs/torchtitan _single.html --self-contained-html

**Distributed (Llama 3.3 70B)**

cvs run torchtitan_distributed \
  --cluster_file ./cluster.json \
  --config_file cvs/input/config_file/training/torchtitan/mi355_llama3_3_70b_distributed_config.json \
  --html ./logs/torchtitan_distributed.html --self-contained-html

**What's included**

torchtitan_single.py — single-node lifecycle: container launch → tokenizer download → smoke → training sweeps → metric gating → loss curve → teardown.
torchtitan_distributed.py — multi-node variant with RDMA networking and scaling efficiency tracking.
Config files (mi355x_torchtitan_*) for Llama 3.1 8B (single), Llama 3.3 70B (single + distributed), and DeepSeek V3 Lite (single), with sibling threshold files.
Sweep combos (precision, batch size) declared in the config; the suite runs each combo sequentially inside a single shared container, freeing GPU memory between combos.

**Loss curve validation**

Replaced the old monotonic step-pair check with a slope-based approach. parse_all_loss_points extracts every (step, lm_loss) pair from the training log; sample_loss_curve reduces these to stride points, configured milestones, and always the first and last. evaluate_loss_decreasing fits a least-squares line across sampled points — robust against individual step noise. A PNG is rendered locally via matplotlib and linked in the HTML report at the test_loss_curve row.

**Security fixes**

Wrapper scripts and scripts_dir hardened from chmod 777 to chmod 700 — prevents other cluster users from reading the embedded HF token or tampering with scripts between write and execution (TOCTOU).
HF token removed from huggingface-cli argv (was visible in ps to all users); now passed as HF_TOKEN environment variable scoped to the download command only.

**Bug fixes**

NCCL_IB_HCA was sourced from nccl_ib_hca_list instead of nccl_ib_hca, silently ignoring the narrower HCA config key.
training.mem_usage threshold was 0 in all cells (always failing); updated to 0.85.
Trailing commas in threshold JSON files caused JSONDecodeError at startup.
scan_for_training_errors docstring claimed it scanned all nodes; it only scans the last node.


cvs/tests/training/torchtitan/torchtitan_single.py
```diff
@@ -0,0 +1,231 @@
...
+    if not variant_config.enforce_thresholds:
+        log.info("enforce_thresholds=false; recorded metrics for combo %s, skipping verdict", combo_key)
+        return
+
+    # Use new cell_key format for threshold lookup
+    cell_key = variant_config.cell_key(
+        micro_batch_size=micro_batch_size,
+        global_batch_size=global_batch_size,
+        precision=precision,
+    )
```
amd-droy: Wrong cell_key() call — passes micro_batch_size / global_batch_size / precision; loader expects a sweep combo key string. API mismatch

cvs/tests/training/torchtitan/torchtitan_single.py
```diff
@@ -0,0 +1,231 @@
...
+        threshold_specs = variant_config.threshold_dict[cell_key]
+    elif result_dict:
+        # Legacy mode: inline result_dict - convert to threshold spec format
+        threshold_specs = {f"training.{k}": {"kind": "min", "value": v} for k, v in result_dict.items()}
+    else:
+        log.warning("no thresholds defined for combo %s; skipping threshold checks", combo_key)
+        return
+
+    # Evaluate thresholds using evaluate_all
+    from cvs.lib.utils.verdict import evaluate_all
```
amd-droy: you can move this to the top in the library call section.

cvs/tests/training/torchtitan/torchtitan_single.py
```diff
@@ -0,0 +1,231 @@
...
+
+    actuals = train_res_dict[combo_key]
+    evaluate_all(
+        threshold_specs,
+        actuals,
+        cell_key,
+        tolerance_pct=5.0,
+    )
```
amd-droy: signature is evaluate_all(actuals, thresholds) only

cvs/lib/training/torchtitan/torchtitan_lib.py
```diff
@@ -0,0 +1,559 @@
...
+            log_files = [f'{self.log_dir}/torchtitan-logs/out-node0/training.log']
+
+        for log_file in log_files:
+            out_dict = self.orch.exec(f'tail -1000 {log_file}')
+            for host, output in out_dict.items():
+                if not output:
+                    continue
+                for err_type, pattern in training_err_dict.items():
+                    if re.search(pattern, output, re.I):
+                        fail_test(f'{err_type} detected in training log on {host}')
```
amd-droy, line 504: verify_dmesg_for_errors - is missing, I think it was there in the legacy files. You can add verify_dmesg_for_errors(orch, start, end, till_end_flag=False)

cvs/tests/training/torchtitan/_shared.py
amd-droy, line 1: is this file referred anywhere? Or is this for later use?

cvs/lib/training/torchtitan/torchtitan_lib.py
```diff
@@ -0,0 +1,559 @@
...
+            if int(self.global_batch_size) > 32:
+                if int(self.global_batch_size) % 32 == 0:
+                    per_gpu_batch_size = int(self.global_batch_size) / 32
+                    self.global_batch_size = str(int(per_gpu_batch_size * total_gpus))
+
+    def run_pretraining_tasks(self):
+        """Snapshot network stats before training (distributed only)."""
+        if self.distributed_training:
+            self.rdma_stats_dict_before = linux_utils.get_rdma_stats_dict(self.orch)
+            self.ethtool_stats_dict_before = linux_utils.get_nic_ethtool_stats_dict(self.orch)
```
amd-droy, line 269: is this function called from anywhere? or can be removed.

cvs/tests/training/torchtitan/torchtitan_distributed.py
```diff
@@ -0,0 +1,227 @@
...
+        lifecycle.record(nodeid, "container_launch", time.monotonic() - t)
+        if not ok:
+            pytest.fail(f"setup_containers() returned False for {name}")
+        if not orch.verify_containers_running(name):
+            pytest.fail(f"container {name} not running after setup_containers()")
+
+        # Stage 2: start sshd — distributed runs always require inter-node MPI
+        # so sshd on 2224 is mandatory on all nodes.
+        # t = time.monotonic()
+        # ok = orch.setup_sshd()
```
amd-droy: ok = orch.setup_sshd() this is commented out in distributed, but running in single? Are they implemented in different ways?

amd-droy:
Also, please update the description section.

cvs/tests/training/torchtitan/torchtitan_single.py
```diff
@@ -0,0 +1,231 @@
...
+
+    # Use new cell_key format for threshold lookup
+    cell_key = variant_config.cell_key(
+        micro_batch_size=micro_batch_size,
+        global_batch_size=global_batch_size,
+        precision=precision,
+    )
+
+    # Prefer external threshold_dict, fallback to legacy result_dict
+    if variant_config.threshold_dict:
```
solaiys: **Blocking — `AttributeError`.** The field on `TorchTitanVariantConfig` is `thresholds`, not `threshold_dict` (see `training_config_loader.py:137`). Verified against a real config:

```
AttributeError: 'TorchTitanVariantConfig' object has no attribute 'threshold_dict'
```

The model is `_Forbid` (`extra="forbid"`), so it can't be populated dynamically either. Also affects L211 and L214.

This is a third failure on top of the `cell_key` and `evaluate_all` signature mismatches @amd-droy flagged — all three are currently masked because every shipped config sets `enforce_thresholds: false`, so the function returns at L198 before reaching any of them.

cvs/tests/training/torchtitan/torchtitan_single.py
```diff
@@ -0,0 +1,231 @@
...
+        threshold_specs = {f"training.{k}": {"kind": "min", "value": v} for k, v in result_dict.items()}
+    else:
+        log.warning("no thresholds defined for combo %s; skipping threshold checks", combo_key)
+        return
+
+    # Evaluate thresholds using evaluate_all
+    from cvs.lib.utils.verdict import evaluate_all
+
+    actuals = train_res_dict[combo_key]
+    evaluate_all(
```
solaiys: Adding one detail to @amd-droy's signature note: the arguments are in the wrong **order** as well as the wrong count. The real signature is `evaluate_all(actuals, thresholds)`, and this passes `(threshold_specs, actuals, ...)` — so just deleting `cell_key` and `tolerance_pct` would silently swap the two remaining arguments and compare thresholds as if they were measurements.

```suggestion
    evaluate_all(train_res_dict[combo_key], threshold_specs)
```

`evaluate_all` raises `ThresholdViolation` on failure, so there's no return value to inspect.

cvs/tests/training/torchtitan/torchtitan_distributed.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    combo_key = request.node.callspec.id
+    if combo_key not in train_res_dict:
+        pytest.skip(f"no recorded results for combo '{combo_key}' (training did not run)")
+
+    if not variant_config.enforce_thresholds:
+        log.info("enforce_thresholds=false; recorded metrics for combo '%s', skipping verdict", combo_key)
+        return
+
+    if not result_dict:
```
solaiys: **Blocking — the distributed suite never reads its threshold files.** This path only consults the inline per-combo `result_dict` and returns here when it's empty. No config defines one:

```
mi355_llama3_1_405b_distributed_config.json  combos=1  with result_dict=NONE
mi355_llama3_1_70b_distributed_config.json   combos=2  with result_dict=NONE
mi355_llama3_3_70b_distributed_config.json   combos=2  with result_dict=NONE
```

And one can't be added, because `TorchTitanSweepCombo` is `_Forbid`:

```
ValidationError: Extra inputs are not permitted [type=extra_forbidden]
```

So the four distributed `*_threshold.json` files are unreachable no matter what `enforce_thresholds` is set to. This suite should use the same `cell_key()` + `variant_config.thresholds` + `evaluate_all()` path as the single-node suite (once that path is fixed), rather than a second and weaker min-only comparison.

cvs/lib/training/torchtitan/torchtitan_lib.py
```diff
@@ -0,0 +1,559 @@
...
+                except ValueError:
+                    fail_test(f'Cannot parse value {val} for metric {metric}')
+
+        # Check network errors if requested
+        if self.distributed_training and self.verify_network_errors == 'True':
+            self.rdma_stats_dict_after = linux_utils.get_rdma_stats_dict(self.orch)
+            self.ethtool_stats_dict_after = linux_utils.get_nic_ethtool_stats_dict(self.orch)
+
+            # Verify no significant error increments
+            for node in self.rdma_stats_dict_before:
```
solaiys: **Blocking — this loop never executes.** `self.rdma_stats_dict_before` is only populated by `run_pretraining_tasks()`, which neither `torchtitan_single.py` nor `torchtitan_distributed.py` ever calls (only the legacy `torchtitan_deepseek_16b_single.py` does). So the dict is always `{}` and no counter is ever compared; `ethtool_stats_dict_before` is never compared at all.

Second problem once that's fixed: `delta > 100` only calls `log.warning`, never `fail_test`, so a NIC degrading mid-run cannot fail the suite. `megatron_lib.verify_training_results` fails on any increase in a matching counter.

This is the concrete consequence of @amd-droy's question about `run_pretraining_tasks` at L267.

cvs/lib/training/torchtitan/torchtitan_lib.py
```diff
@@ -0,0 +1,559 @@
...
+
+                log_path = f'{self.log_dir}/torchtitan-logs/out-node{i}/training.log'
+                self.orch.exec(f'mkdir -p $(dirname {log_path})')
+
+                full_cmd = cmd + f': > {log_path}; nohup {torchrun_cmd} > {log_path} 2>&1 & disown'
+
+                script_cmd = (
+                    f"cat > {self.scripts_dir}/distributed_wrapper_script_{i}.sh << 'WRAPPER_EOF'\n"
+                    f"#!/bin/bash\n{full_cmd}\nWRAPPER_EOF\n; "
+                    f'chmod 777 {self.scripts_dir}/distributed_wrapper_script_{i}.sh'
```
solaiys: **Blocking — secret in a world-writable file.** The script body contains `export HF_TOKEN={self.hf_token}` (L287 and L387), and it is `chmod 777`'d here inside a `scripts_dir` that is itself `chmod 777`'d at L252.

On a shared cluster that exposes the HF token to every local user, and because the file is world-*writable* and then executed, another user can rewrite it between creation and launch. `0600` on the scripts and `0700` on `scripts_dir` addresses both.

cvs/input/config_file/training/torchtitan/mi355_llama3_1_8b_single_config.json
```diff
@@ -0,0 +1,84 @@
+{
+  "schema_version": 1,
+  "framework": "torchtitan_single",
+  "gpu_arch": "mi355",
+  "enforce_thresholds": false,
```
solaiys, line 5: All eight new configs set `enforce_thresholds: false`, so `test_throughput` returns at its guard and no threshold is ever asserted by either suite — even though the threshold files themselves are correct and their cell keys match `cell_key()` output exactly in all 8 cases.

Record-only is a defensible choice for a first drop, but note it is also what hides the three `TypeError`/`AttributeError` bugs in `torchtitan_single.py::test_throughput`. Those need fixing first, otherwise flipping this to `true` is a guaranteed crash rather than a config change.

cvs/lib/training/torchtitan/torchtitan_lib.py
```diff
@@ -0,0 +1,559 @@
...
+        for pat in patterns:
+            matches = re.findall(pat, output, re.I)
+            if matches:
+                # TorchTitan may emit comma-grouped numbers
+                out[metric] = [m.replace(',', '') for m in matches]
+                break
+    return out
+
+
+def _is_training_complete(output, iterations):
```
solaiys, line 73: Non-blocking: `rf'step:\s+{iterations}'` has no word boundary, so `iterations=10` also matches `step: 100` and `step: 1000` and would report completion early. Every shipped config uses `training_iterations: 10` so steps never exceed the target and it can't misfire today.

`megatron_lib._is_training_complete` handles this with `rf'iteration\s+{n}\s*/\s*{n}\b'`; adding `\b` here would do the same job.

solaiys (commented):
## Review — 5 blocking issues

Reviewed the new TorchTitan suite (24 files, +2487). Good news first: all 16 new JSON files parse cleanly, `ruff format --check` / `ruff check` pass on all 16 Python files, and the threshold cell keys line up exactly with `cell_key()` output in all eight model configs — I diffed every one:

```
mi355_llama3_1_8b_single_threshold.json    keys ['MBS=1,GBS=8,PRECISION=bf16', 'MBS=1,GBS=8,PRECISION=fp8']   match: YES
mi355_llama3_3_70b_distributed_threshold.json                                        ... 8/8 match: YES
```

The problem is that **nothing in the suite ever reads them**. Every threshold path is either broken or disabled, so both suites currently run training and record numbers without gating on anything. Four independent reasons, each verified by executing the real code:

### Blocking issues

**1. `test_throughput` in `torchtitan_single.py` raises on all three of its API calls**

Loading a real config and running lines 203-231 exactly as written:

```
REAL signatures:
   cell_key     (self, combo_key: 'str') -> 'str'
   evaluate_all (actuals, thresholds)

L203  cell_key(micro_batch_size=..., global_batch_size=..., precision=...)
      -> TypeError: TorchTitanVariantConfig.cell_key() got an unexpected keyword argument 'micro_batch_size'
L210  variant_config.threshold_dict
      -> AttributeError: 'TorchTitanVariantConfig' object has no attribute 'threshold_dict'
L226  evaluate_all(threshold_specs, actuals, cell_key, tolerance_pct=5.0)
      -> TypeError: evaluate_all() got an unexpected keyword argument 'tolerance_pct'
```

@amd-droy already flagged the `cell_key` and `evaluate_all` signatures. Two things to add beyond that:

- `L210`/`L211`/`L214` reference `variant_config.threshold_dict`, but the field on `TorchTitanVariantConfig` is `thresholds`. Since the model is `_Forbid` (`extra="forbid"`), there is no chance of it being populated dynamically — it's a plain `AttributeError`.
- The `evaluate_all` arguments are also in the wrong **order**, not just the wrong count: the real signature is `evaluate_all(actuals, thresholds)`, and the call passes `(threshold_specs, actuals, ...)`. Dropping the two extra arguments alone would silently swap them.

The corrected block is:

```python
cell_key = variant_config.cell_key(combo_key)
if not variant_config.thresholds:
    log.warning("no thresholds defined for combo %s; skipping threshold checks", combo_key)
    return
if cell_key not in variant_config.thresholds:
    log.warning("no threshold entry for cell %s; skipping", cell_key)
    return
evaluate_all(train_res_dict[combo_key], variant_config.thresholds[cell_key])
```

Note `evaluate_all` raises `ThresholdViolation` itself, so the result does not need to be inspected — but as written the current call discards a return value that never existed.

**2. All eight shipped configs set `enforce_thresholds: false`, which is what hides issue 1**

```
mi355_deepseek_v3_16b_single_config.json      enforce_thresholds=False
mi355_llama3_1_405b_distributed_config.json   enforce_thresholds=False
mi355_llama3_1_70b_distributed_config.json    enforce_thresholds=False
mi355_llama3_1_8b_single_config.json          enforce_thresholds=False
mi355_llama3_3_70b_distributed_config.json    enforce_thresholds=False
mi355_llama3_3_70b_single_config.json         enforce_thresholds=False
mi355_mixtral_8x22b_single_config.json        enforce_thresholds=False
mi355_qwen3_32b_single_config.json            enforce_thresholds=False
```

`test_throughput` returns at the `if not variant_config.enforce_thresholds` guard before reaching any of the broken calls, so the suite looks green today and the three `TypeError`s land on whoever first flips a config to `true`. If record-only is the deliberate intent for the first drop, that's reasonable — but then issue 1 needs fixing before the flag is usable, otherwise enabling gating is a guaranteed error rather than a config change.

**3. The distributed suite never reads the threshold files at all**

`torchtitan_distributed.py::test_throughput` has an entirely different implementation from the single-node one: it ignores `variant_config.thresholds` and reads only the per-combo inline `result_dict`, bailing out at `if not result_dict`. No config defines one:

```
mi355_llama3_1_405b_distributed_config.json   combos=1  with result_dict=NONE
mi355_llama3_1_70b_distributed_config.json    combos=2  with result_dict=NONE
mi355_llama3_3_70b_distributed_config.json    combos=2  with result_dict=NONE
```

And one cannot be added, because `TorchTitanSweepCombo` is `_Forbid` and rejects the key outright:

```
TorchTitanSweepCombo(..., result_dict={'training.tps': 100})
  -> ValidationError: Extra inputs are not permitted [type=extra_forbidden]
```

So the four distributed `*_threshold.json` files are unreachable regardless of `enforce_thresholds`. The distributed suite should use the same `cell_key` + `variant_config.thresholds` + `evaluate_all` path as the single-node one (once that path is fixed), rather than a second, weaker inline comparison.

**4. The RDMA/NIC error check can never fire**

`verify_training_results` (`torchtitan_lib.py:549`) iterates `self.rdma_stats_dict_before`, but `run_pretraining_tasks()` — the only thing that populates it — is never called by either new suite. Only the legacy `torchtitan_deepseek_16b_single.py` calls it. So the dict is always `{}`, the loop body never executes, and `ethtool_stats_dict_before` is never compared at all.

Even once it is populated, the check only calls `log.warning` on `delta > 100`; it never calls `fail_test`, so a NIC degrading mid-run cannot fail the suite. Compare `megatron_lib.verify_training_results`, which fails on any increase. This is @amd-droy's `torchtitan_lib.py:267` question — the concrete consequence is that network validation is entirely inert.

**5. The HF token is written into world-writable files**

`torchtitan_lib.py` embeds `export HF_TOKEN={self.hf_token}` into the wrapper script (L287 and L387), which is then written out and `chmod 777`'d (L430) into a `scripts_dir` that is itself `chmod 777`'d (L252). Any local user on the node can read the token, and because the script is world-*writable* and subsequently executed, they can also rewrite it before it runs. `0600` on the scripts and `0700` on the directory fixes both. (The same pattern is present in the Megatron suite in #307.)

### Confirming existing review comments

Not re-raising, just noting I reached the same conclusions as @amd-droy:

- **`torchtitan_lib.py:499` — `verify_dmesg_for_errors` is missing.** Confirmed absent from the whole file; `megatron_lib.py` calls it at the end of `verify_training_results`, so this is a real gap relative to the sibling suite.
- **`torchtitan_distributed.py:125` — `setup_sshd()` commented out.** The comment immediately above it says distributed runs "always require inter-node MPI so sshd on 2224 is mandatory on all nodes", which directly contradicts the code being disabled, while the single-node suite (which needs it least) does run it.
- **`_shared.py` unreferenced.** Confirmed — its own docstring says "Import via `from ._shared import *` in each test file", but neither suite imports it, and its two functions take a `phdl` fixture that the new `conftest.py` does not provide.

### Non-blocking notes

- `_is_training_complete` (`torchtitan_lib.py:73`) uses `rf'step:\s+{iterations}'` with no word boundary, so `iterations=10` also matches `step: 100` and `step: 1000`. Every shipped config uses `training_iterations: 10` so steps never exceed the target and it can't misfire today. `megatron_lib._is_training_complete` gets this right with `rf'iteration\s+{n}\s*/\s*{n}\b'`.
- `poll_for_training_completion` (`torchtitan_lib.py:505-508`) has identical `if`/`else` branches — both assign `out-node0/training.log`, so in distributed mode only rank 0's log is polled and the function returns as soon as rank 0 finishes.

cvs/tests/training/torchtitan/torchtitan_single.py
amd-rthummal: Implemented

cvs/tests/training/torchtitan/torchtitan_single.py
amd-rthummal: Implemented

cvs/tests/training/torchtitan/torchtitan_single.py
amd-rthummal: Implemented

cvs/tests/training/torchtitan/_shared.py
amd-rthummal: Implemented

cvs/tests/training/torchtitan/torchtitan_distributed.py
amd-rthummal: Implemented

cvs/lib/training/torchtitan/torchtitan_lib.py
amd-rthummal: Implemented

cvs/lib/training/torchtitan/torchtitan_lib.py
amd-rthummal: Implemented

cvs/tests/training/torchtitan/torchtitan_single.py
amd-rthummal: Implemented

cvs/tests/training/torchtitan/torchtitan_single.py
amd-rthummal: Implemented

cvs/tests/training/torchtitan/torchtitan_distributed.py
amd-rthummal: Implemented

cvs/lib/training/torchtitan/torchtitan_lib.py
amd-rthummal: Implemented

cvs/input/config_file/training/torchtitan/mi355_llama3_1_8b_single_config.json
amd-rthummal: Implemented

cvs/lib/training/torchtitan/torchtitan_lib.py
amd-rthummal: Implemented

cvs/lib/training/torchtitan/torchtitan_lib.py
amd-rthummal: Implemented

amd-droy (approved):
looks good to me. thanks Rajesh.

PR #307, opened 2026-08-11
Add Megatron Single-Node & Distributed Training Suites with Configs, Thresholds, and Sweeps

sukesh-amd:
This PR introduces a complete Megatron-LM pre-training validation suite for MI325X GPUs, covering single-node and distributed (multi-node) runs. The suite drives Megatron-LM training jobs inside a container, parses training logs, and gates results against configurable per-combo performance and correctness thresholds with a linked HTML report.

### How to run

**Single-node (Llama 3.1 8B)**
```bash
cvs run megatron_single \
  --cluster_file ./cluster.json \
  --config_file cvs/input/config_file/training/megatron/mi325x_megatron_llama-3.1-8b_single.json \
  --html ./logs/megatron_single.html --self-contained-html
```

**Distributed (Llama 3.3 70B)**
```bash
cvs run megatron_distributed \
  --cluster_file ./cluster.json \
  --config_file cvs/input/config_file/training/megatron/mi325x_megatron_llama-3.3-70b_distributed.json \
  --html ./logs/megatron_distributed.html --self-contained-html
```

### What's included

- `megatron_single.py` — single-node lifecycle: container launch → tokenizer download → smoke → training sweeps → metric gating → loss curve → teardown.
- `megatron_distributed.py` — multi-node variant with RDMA networking and scaling efficiency tracking.
- Config files (`mi325x_megatron_*`) for Llama 3.1 8B (single), Llama 3.3 70B (single + distributed), and DeepSeek V2 Lite (single), with sibling threshold files.
- Sweep combos (precision, batch size) declared in the config; the suite runs each combo sequentially inside a single shared container, freeing GPU memory between combos.

### Loss curve validation

Replaced the old monotonic step-pair check with a slope-based approach. `parse_all_loss_points` extracts every `(step, lm_loss)` pair from the training log; `sample_loss_curve` reduces these to stride points, configured milestones, and always the first and last. `evaluate_loss_decreasing` fits a least-squares line across sampled points — robust against individual step noise. A PNG is rendered locally via matplotlib and linked in the HTML report at the `test_loss_curve` row.

### Security fixes

- Wrapper scripts and `scripts_dir` hardened from `chmod 777` to `chmod 700` — prevents other cluster users from reading the embedded HF token or tampering with scripts between write and execution (TOCTOU).
- HF token removed from `huggingface-cli` argv (was visible in `ps` to all users); now passed as `HF_TOKEN` environment variable scoped to the download command only.

### Bug fixes

- `NCCL_IB_HCA` was sourced from `nccl_ib_hca_list` instead of `nccl_ib_hca`, silently ignoring the narrower HCA config key.
- `training.mem_usage` threshold was `0` in all cells (always failing); updated to `0.85`.
- Trailing commas in threshold JSON files caused `JSONDecodeError` at startup.
- `scan_for_training_errors` docstring claimed it scanned all nodes; it only scans the last node.


amd-droy:
Please update the description section.

cvs/lib/training/megatron/megatron_lib.py
```diff
@@ -0,0 +1,886 @@
...
+    def scan_for_training_errors(self):
+        """Scan training logs from ALL nodes for known error patterns.
+
+        Returns:
+            bool: True if no error patterns found on any node; False otherwise.
+        """
+        log.info('Scan for training errors')
+        training_pass = True
+
+        output = self._read_last_node_log()
```
amd-droy, line 723: is it scanning all nodes or only the last node? Comments says it scans the training logs from all nodes.

cvs/tests/training/megatron/megatron_distributed.py
```diff
@@ -0,0 +1,378 @@
...
+from cvs.lib.training.megatron.megatron_lib import MegatronTrainingJob
+from cvs.lib.training.megatron.utils.loss_curve import parse_loss_at_steps, check_loss_decreasing
+from cvs.lib.training.megatron.utils.scaling import compute_scaling_efficiency
+from cvs.lib.utils.verdict import _check_one, ThresholdViolation
+from cvs.lib.utils_lib import update_test_result
+
+log = globals.log
+
+# Loss curve checkpoints: iteration numbers at which lm_loss is sampled.
+_LOSS_CURVE_STEPS = [100, 500, 1000, 5000]
```
amd-droy: test_loss_curve hardcodes _LOSS_CURVE_STEPS = [100, 500, 1000, 5000], but the sample configs run only "training_iterations": "10". With the current skip logic (len(losses) < 2), doesn't this test always skips on default configs?

cvs/input/config_file/training/megatron/mi325x_megatron_llama-3.3-70b_single_threshold.json
```diff
@@ -0,0 +1,39 @@
...
+      "value": 500
+    },
+    "training.tokens_per_gpu": {
+      "kind": "min",
+      "value": 1000
+    },
+    "training.mem_usage": {
+      "kind": "max",
+      "value": 0
+    },
```
solaiys: **Blocking — invalid JSON.** Trailing comma before the closing `}` of the cell object, so `json.load` rejects the whole file and `load_training_variant` raises `JSONDecodeError: Expecting property name enclosed in double quotes: line 20 column 3`. The suite dies in `conftest.variant_config` before any test runs.

Same problem at line **37**. Fix is to drop the comma:

```suggestion
    }
```

cvs/input/config_file/training/megatron/mi325x_megatron_llama-3.1-8b_single_threshold.json
```diff
@@ -0,0 +1,75 @@
...
+      "value": 500
+    },
+    "training.tokens_per_gpu": {
+      "kind": "min",
+      "value": 1000
+    },
+    "training.mem_usage": {
+      "kind": "max",
+      "value": 0
+    },
```
solaiys: **Blocking — invalid JSON.** Same trailing-comma problem as the other single-node threshold files; this one has it in four places: lines **19**, **37**, **55**, **73**. `load_training_variant` fails with `JSONDecodeError` at line 20 column 3.

```suggestion
    }
```

cvs/input/config_file/training/megatron/mi325x_megatron_deepseek-v2-lite_single_threshold.json
```diff
@@ -0,0 +1,39 @@
...
+      "value": 500
+    },
+    "training.tokens_per_gpu": {
+      "kind": "min",
+      "value": 1000
+    },
+    "training.mem_usage": {
+      "kind": "max",
+      "value": 0
+    },
```
solaiys: **Blocking — invalid JSON.** Trailing comma here and at line **37**; `load_training_variant` fails with `JSONDecodeError` at line 20 column 3, so this config cannot be run at all.

```suggestion
    }
```

cvs/input/config_file/training/megatron/mi325x_megatron_llama-3.3-70b_distributed_threshold.json
```diff
@@ -0,0 +1,47 @@
...
+      "kind": "max",
+      "value": 500
+    },
+    "training.tokens_per_gpu": {
+      "kind": "min",
+      "value": 1000
+    },
+    "training.mem_usage": {
+      "kind": "max",
+      "value": 0
```
solaiys: **Blocking — this threshold can never pass.** `mem_usage` is parsed from `mem usages:\s+([0-9.]+)` and is always positive, while `_check_one` for `kind: max` fails when `actual > target`. With `value: 0` every run fails:

```
actual=55.6  -> FAIL: training.mem_usage: actual 55.6 > max 0.0
```

This same `max: 0` appears in every cell of all four threshold files. Either set a real ceiling (e.g. HBM headroom for the device) or drop `training.mem_usage` from the gated set until it's calibrated — `test_metric` treats an absent spec as "don't gate".

cvs/lib/training/megatron/megatron_lib.py
```diff
@@ -0,0 +1,886 @@
...
+            + 'export EXP_NAME="megatron_training"; '
+            + 'export TORCH_NCCL_ASYNC_ERROR_HANDLING=0; '
+        )
+
+        if self.distributed_training is True:
+            # Add the backend network related environment variables ..
+            cmd = (
+                cmd
+                + f'export NCCL_IB_HCA_LIST={self.nccl_ib_hca_list}; '
+                + f'export NCCL_IB_HCA={self.nccl_ib_hca_list}; '
```
solaiys: **Blocking — wrong variable.** This sets `NCCL_IB_HCA` from `nccl_ib_hca_list` rather than `nccl_ib_hca`. `self.nccl_ib_hca` is assigned at L299 and then never read anywhere in the file, so the `nccl_ib_hca` config key has no effect at all.

Both keys share a default and are `<changeme>` in the distributed config, so nothing breaks today — but an operator who narrows `nccl_ib_hca` to a subset of HCAs will silently get the full list instead.

```suggestion
                + f'export NCCL_IB_HCA={self.nccl_ib_hca}; '
```

cvs/lib/training/megatron/megatron_lib.py
```diff
@@ -0,0 +1,886 @@
...
+                    f'nohup {self.scripts_dir}/distributed_wrapper_script_{i}.sh > '
+                    f'{self.combo_log_dir}/out-node{i}/training.log 2>&1 &'
+                    for i in range(n)
+                ]
+            )
+        else:
+            # Write single-node wrapper script on bare host
+            self.orch.all.exec(
+                f'echo {shlex.quote(self.job_cmd)} > {self.scripts_dir}/single_node_wrapper_script.sh '
+                f'&& chmod 777 {self.scripts_dir}/single_node_wrapper_script.sh'
```
solaiys: **Blocking — secret in a world-writable file.** The wrapper script body contains `export HF_TOKEN="..."` (L541), and it's written to the bare host and `chmod 777`'d here (and at L579 for the distributed scripts), inside a directory that is itself `sudo chmod 777`'d at L390.

Two separate problems on a shared cluster: any local user can read the HF token, and because the file is world-*writable* and then executed, they can rewrite it between creation and launch. `0600` on the scripts and `0700` on `scripts_dir` addresses both.

cvs/lib/training/megatron/megatron_lib.py
```diff
@@ -0,0 +1,886 @@
...
+        if not self._needs_local_tokenizer():
+            return
+
+        local_dir = f'{self.data_cache_dir}/{self.model_name}'
+        log.info('Downloading tokenizer.model for %s into %s', self.model_name, local_dir)
+        self.orch.exec(
+            f'huggingface-cli download {self.tokenizer_model} '
+            f'--include "tokenizer.model" '
+            f'--local-dir {local_dir} '
+            f'--token {self.hf_token}'
```
solaiys: **Blocking — token exposed in the process list.** Passing the token as an argv element makes it visible via `ps` to every user on the node for the duration of the download.

`huggingface-cli` already reads `HF_TOKEN` from the environment, so this argument can simply be dropped and the token exported instead.

solaiys (commented):
## Review — 4 blocking issues

Reviewed the new Megatron suite (20 files, +3099). Structure is good: `pytest_generate_tests` driving combos off `sweep.runs`, `stop_training_processes()` in a `finally` after every combo, and a load-time `validate_thresholds_cover_sweep` validator are all the right shape. Lint is clean too — `ruff format --check`, `ruff check`, and the pylint logging gate all pass on the 14 new Python files.

However, **three of the four new configs cannot be loaded at all**, and the fourth would fail every run. Details below, each verified by executing the real code paths rather than by reading.

### Blocking issues

**1. Three of the four new threshold files are invalid JSON (trailing commas)**

`json.load` rejects a comma before a closing brace. Running the real loader that `conftest.variant_config` uses:

```
load_training_variant(mi325x_megatron_deepseek-v2-lite_single.json) -> JSONDecodeError: line 20 column 3
load_training_variant(mi325x_megatron_llama-3.1-8b_single.json)     -> JSONDecodeError: line 20 column 3
load_training_variant(mi325x_megatron_llama-3.3-70b_single.json)    -> JSONDecodeError: line 20 column 3
```

So `cvs run megatron_single --config_file <any of the three>` dies in fixture setup before a single test executes. Affected lines (the comma after the `training.mem_usage` block that closes each cell):

- `mi325x_megatron_llama-3.3-70b_single_threshold.json` — lines **19**, **37**
- `mi325x_megatron_llama-3.1-8b_single_threshold.json` — lines **19**, **37**, **55**, **73**
- `mi325x_megatron_deepseek-v2-lite_single_threshold.json` — lines **19**, **37**

`mi325x_megatron_llama-3.3-70b_distributed_threshold.json` is the only valid one. Worth adding a JSON-parse check over `cvs/input/config_file/**` to `make ut` so this class of breakage can't land again.

**2. `training.mem_usage` is gated at `max: 0`, so every run fails**

Present in **all four** threshold files, in **every** cell. `mem_usage` is parsed from `mem usages:\s+([0-9.]+)`, which is always a positive number, and `_check_one` fails when `actual > target`:

```
actual=0.0    -> PASS
actual=55.6   -> FAIL: training.mem_usage: actual 55.6 > max 0.0
actual=140.2  -> FAIL: training.mem_usage: actual 140.2 > max 0.0
```

Once issue 1 is fixed, this makes `test_metric` fail unconditionally on every combo. Either set a real ceiling (device HBM headroom) or drop the metric from the gated set until it's calibrated.

**3. The HF token is written into world-writable files and passed on the command line**

`megatron_lib.py` embeds `export HF_TOKEN="{self.hf_token}"` into the wrapper script (L541), then writes that script to the bare host and `chmod 777`s it (L579 distributed, L666 single-node) inside a directory that is itself `sudo chmod 777`'d (L390). On a shared cluster any local user on the node can read the token, and because the file is world-*writable* and then executed, they can also rewrite the script between creation and launch. Separately, `download_tokenizer_model` passes `--token {self.hf_token}` as an argv element (L420), which is visible in `ps` to every user on the box.

`0600` on the script and `0700` on the directory would fix the exposure; for the CLI, `HF_TOKEN` is already read from the environment by `huggingface-cli`, so the explicit `--token` argument can just be dropped.

**4. The `nccl_ib_hca` config key is silently ignored**

`megatron_lib.py:555` sets `NCCL_IB_HCA` from `self.nccl_ib_hca_list`, not `self.nccl_ib_hca`:

```python
+ f'export NCCL_IB_HCA_LIST={self.nccl_ib_hca_list}; '
+ f'export NCCL_IB_HCA={self.nccl_ib_hca_list}; '   # <- should be self.nccl_ib_hca
```

`self.nccl_ib_hca` is assigned at L299 and then never read anywhere in the file. Both keys default to the same string and both are `<changeme>` in the distributed config, so nothing breaks today — but the schema advertises a key that has no effect, and an operator who narrows `nccl_ib_hca` to a subset of HCAs will silently get the full list instead.

### Confirming existing review comments

Not re-raising these since @amd-droy already flagged them, just adding the evidence I gathered:

- **`megatron_lib.py:720` — scans only the last node.** Confirmed, and there's an extra wrinkle worth knowing: `_read_last_node_log` calls `self.orch.exec(...)`, which fans the *same* `cat .../out-node{n-1}/training.log` out to **every** host. Since `start_training_job` creates only `out-node{i}` on host `i`, that `cat` fails on every host except the last, and the method then relies on `list(out_dict.values())[-1]` landing on the one host where the file exists. So it works only as long as result-dict ordering matches host ordering. `test_loss_curve` in the suite file already does this correctly with `orch.exec(..., hosts=[last_host])` — the same targeting in `_read_last_node_log` would fix both the correctness and the ordering dependency. Note this also means `get_training_results_dict` reports throughput from a single rank only.

- **`megatron_distributed.py:37` — loss curve always skips.** Confirmed against all four configs:

```
mi325x_megatron_deepseek-v2-lite_single.json    iterations=10  reachable checkpoints=NONE
mi325x_megatron_llama-3.1-8b_single.json        iterations=10  reachable checkpoints=NONE
mi325x_megatron_llama-3.3-70b_distributed.json  iterations=10  reachable checkpoints=NONE
mi325x_megatron_llama-3.3-70b_single.json       iterations=10  reachable checkpoints=NONE
```

None of `[100, 500, 1000, 5000]` is reachable at 10 iterations, so `len(losses) < 2` skips on every shipped config. Deriving the checkpoints from `training_iterations` (or gating the test on `iterations >= _LOSS_CURVE_STEPS[1]`) would make it meaningful.

### Non-blocking notes

- `tune_model_params` (L394-399) computes `per_gpu_batch_size = int(self.global_batch_size) / 32`, which is true division, so `global_batch_size` becomes a float and renders as `GBS=32.0` in the command. Every call site currently passes `tune_model_params=False`, so it's unreachable — but it will bite whoever turns it on. `//` fixes it.
- `cvs/lib/training/__init__.py` is absent while `cvs/lib/training/megatron/__init__.py` exists. I checked this doesn't actually break anything — `find_packages` skips the tree, but MANIFEST's `recursive-include cvs *` plus `include_package_data` still ships the files, and I confirmed `import cvs.lib.training.megatron.megatron_lib` works from a real `pip install` of the sdist via PEP 420. Adding the empty `__init__.py` would just make it consistent with every other lib subpackage.

cvs/lib/training/megatron/megatron_lib.py
```diff
@@ -0,0 +1,886 @@
...
+        self.orch.all.exec(f'mkdir -p {self.scripts_dir}')
+        self.orch.all.exec(f'sudo chmod 777 {self.scripts_dir}')
+
+        # Let us override some of the params based on number of nodes and platform
+        # if override flag set ..
+        if self.tune_model_params:
+            # Assuming the training json configs were built with 4 nodes = 32 gpus
+            if int(self.global_batch_size) > 32:
+                if int(self.global_batch_size) % 32 == 0:
+                    per_gpu_batch_size = int(self.global_batch_size) / 32
```
solaiys, line 398: Non-blocking (unreachable today): `/` is true division, so `global_batch_size` becomes a float and renders as `GBS=32.0` in the training command, which Megatron will not accept as an int. Every call site currently passes `tune_model_params=False` so this never executes, but `//` would make it safe for whoever enables it.

cvs/lib/training/megatron/megatron_lib.py
sukesh-amd: Made the required changes

cvs/tests/training/megatron/megatron_distributed.py
sukesh-amd: Fixed

cvs/input/config_file/training/megatron/mi325x_megatron_llama-3.3-70b_single_threshold.json
sukesh-amd: Done

cvs/lib/training/megatron/megatron_lib.py
sukesh-amd: Done

cvs/lib/training/megatron/megatron_lib.py
sukesh-amd: Fixed

cvs/lib/training/megatron/megatron_lib.py
sukesh-amd: Done

cvs/input/config_file/training/megatron/mi325x_megatron_llama-3.3-70b_distributed_threshold.json
sukesh-amd: Done

cvs/input/config_file/training/megatron/mi325x_megatron_llama-3.1-8b_single_threshold.json
sukesh-amd: fixed

cvs/input/config_file/training/megatron/mi325x_megatron_llama-3.3-70b_single_threshold.json
```diff
@@ -0,0 +1,39 @@
...
+      "kind": "max",
+      "value": 500
+    },
+    "training.tokens_per_gpu": {
+      "kind": "min",
+      "value": 1000
+    },
+    "training.mem_usage": {
+      "kind": "max",
+      "value": 0.85
```
solaiys, line 18: Thanks for removing the trailing commas — all four threshold files parse now, and the three single-node configs load.

On this value: `0.85` is only right if `mem usages:` is a **0-1 fraction of HBM**. I can't verify that here because the line comes from the training script inside the container, and the other three specs in this same cell are absolute (`throughput_per_gpu: 100`, `tokens_per_gpu: 1000`, `elapsed_time_per_iteration: 500` ms), which makes `0.85` look like it's in different units from its neighbours.

If the metric is GB/MB (say `140.2`), then `actual > 0.85` is still true on every run and the unconditional failure has just moved from `0` to `0.85`. Could you paste one real `mem usages: <value>` line from a run? If it's a fraction, this is settled and I'll drop it.

cvs/lib/training/megatron/megatron_lib.py
```diff
@@ -0,0 +1,886 @@
...
+        Args:
+            tail_lines (int): If > 0, only the last N lines of the log are read.
+
+        Returns:
+            str: Log text from the last node.
+        """
+        n = len(self.orch.hosts)
+        tail_suffix = f' | tail -{tail_lines}' if tail_lines > 0 else ''
+        out_dict = self.orch.exec(f'cat {self.combo_log_dir}/out-node{n - 1}/training.log{tail_suffix}')
+        return list(out_dict.values())[-1]
```
solaiys: Following up on the docstring change in `scan_for_training_errors` — updating it to "the last node" does resolve the contradiction @amd-droy flagged, so if last-node-only is deliberate for this drop, fine. Just noting the behaviour here is unchanged and what it costs:

1. `self.orch.exec(...)` fans the *same* `cat .../out-node{n-1}/training.log` out to **every** host, but `start_training_job` only creates `out-node{i}` on host `i` — so the `cat` fails everywhere except the last host, and `list(out_dict.values())[-1]` is relied upon to land on the one host where the file exists. That holds only while result-dict ordering matches host ordering.
2. Functionally, an `NCCL ERROR` or `GPU HW ERROR` on rank 0 of a 2-node run is never scanned, and `get_training_results_dict` reports throughput from a single rank.

`test_loss_curve` in the suite already does the targeted version of this (`orch.exec(..., hosts=[last_host])`), which would at least remove the ordering dependency:

```suggestion
        n = len(self.orch.hosts)
        last_host = self.orch.hosts[-1]
        tail_suffix = f' | tail -{tail_lines}' if tail_lines > 0 else ''
        out_dict = self.orch.exec(
            f'cat {self.combo_log_dir}/out-node{n - 1}/training.log{tail_suffix}', hosts=[last_host]
        )
        return out_dict.get(last_host) or ''
```

cvs/lib/training/megatron/megatron_lib.py
```diff
@@ -0,0 +1,886 @@
...
+        For deepseek/mixtral, downloads the tokenizer.model file into data_cache_dir
+        and stores the known local path in self.local_tokenizer_path.
+        """
+        if not self._needs_local_tokenizer():
+            return
+
+        local_dir = f'{self.data_cache_dir}/{self.model_name}'
+        log.info('Downloading tokenizer.model for %s into %s', self.model_name, local_dir)
+        self.orch.exec(
+            f'export HF_TOKEN={shlex.quote(self.hf_token)}; '
```
solaiys, line 417: Good change — dropping `--token` removes the token from `huggingface-cli`'s own argv, and `shlex.quote` is the right call. Two residuals for a follow-up rather than a block:

1. The token is still inside the command string handed to `orch.exec`, and that whole string becomes the argv of the remote shell, so it remains visible in `ps` on the node for the duration of the download — and it will land in the CVS command log if commands are logged. Reading it from the already-configured `hf_token_file` inside the container, instead of interpolating it into the command, would avoid both.
2. Elsewhere, `echo ... > script.sh && chmod 700 script.sh` leaves the file at the default umask (usually `0644`) for the instant between the redirect and the `chmod`. A leading `umask 077;` in the same command closes that window.

solaiys (commented):
## Re-review of `d14e7c53` — good progress, one new blocker before merge

Thanks for the quick turnaround. Re-verified everything by executing the same checks against the new head. **Four of the five points I raised are genuinely fixed**, including the two that made the suite unrunnable. One new regression came in with the fix commit, one item needs a units confirmation from you, and one is still open.

### Fixed — confirmed

**1. The invalid-JSON blocker is gone.** All four threshold files now parse, and the trailing commas are removed. The loader no longer raises `JSONDecodeError`; the only remaining load failures are the intentional `<changeme>` placeholder guard:

```
mi325x_megatron_deepseek-v2-lite_single.json  -> ValueError: unfilled placeholder '<changeme>' in: container.image
mi325x_megatron_llama-3.1-8b_single.json      -> ValueError: unfilled placeholder '<changeme>' in: container.image
mi325x_megatron_llama-3.3-70b_single.json     -> ValueError: unfilled placeholder '<changeme>' in: container.image
mi325x_megatron_llama-3.3-70b_distributed.json-> ValueError: unfilled placeholder '<changeme>' in: config.nnodes, ...
```

That's the designed operator-fills-it-in behaviour, not breakage. This was the issue that stopped all three single-node configs from running at all, so it's the important one.

**2. `nccl_ib_hca` now honoured.** `export NCCL_IB_HCA={self.nccl_ib_hca}` — the config key is no longer dead.

**3. Token handling is materially better.** `chmod 777` → `chmod 700` on `scripts_dir` and both wrapper scripts, and `--token` replaced with `export HF_TOKEN={shlex.quote(...)}`. The world-readable/world-writable script exposure is closed. Two residuals worth a follow-up rather than a block — see the inline note on the download call.

**4. @amd-droy's loss-curve point is properly fixed, not papered over.** The config-driven `loss_curve` block plus first/last-step inclusion in `sample_loss_curve` means short runs now produce a usable curve. Simulated against a realistic 10-iteration Megatron log:

```
parsed points            : 10
sampled (sample_every=10): [(1, 10.9), (10, 4.6)]
evaluate_loss_decreasing : (True, -0.70, 'loss slope -0.7/step over 2 points ...; decreasing')

rising-loss run          : (False, +0.444, '... NOT decreasing')
```

So it runs instead of skipping on the shipped `training_iterations: 10` configs, and it correctly fails a diverging run. That's a real gate now.

### New blocker introduced by the fix commit

**`ruff format --check` fails on five files, so CI `fmt-check` will be red.** All five are files `d14e7c53` touched, and I confirmed each was clean immediately before it:

```
                          pre-fix (a23e5e28)   post-fix (d14e7c53)
loss_curve.py             CLEAN                NEEDS-FMT
loss_curve_plot.py        (new file)           NEEDS-FMT
conftest.py               CLEAN                NEEDS-FMT
megatron_distributed.py   CLEAN                NEEDS-FMT
megatron_single.py        CLEAN                NEEDS-FMT
megatron_lib.py           CLEAN                CLEAN
```

At the previous head this was `14 files already formatted`; it is now `5 files would be reformatted`. `ruff check` still passes and the pylint logging gate is fine, so this is formatting only — `make fmt` (or `ruff format` on those five paths) with the pinned 0.14.8 in `.ruff_venv` should clear it.

### Needs a units confirmation from you

**`training.mem_usage` went from `max: 0` to `max: 0.85`, which is only correct if the metric is a 0-1 fraction.** I can't settle this from the repo, because `mem usages:` is emitted by the training script inside the container rather than by anything here. The concern is that its sibling thresholds in the same cell are all absolute (`throughput_per_gpu: 100`, `tokens_per_gpu: 1000`, `elapsed_time_per_iteration: 500` ms), so `0.85` reads like it's in different units from its neighbours:

- if `mem usages:` is a **fraction of HBM** (e.g. `0.62`), then `0.85` is a sensible 85% ceiling and this is fully fixed;
- if it's **GB or MB** (e.g. `140.2`), then `actual > 0.85` is still true on every run — the unconditional failure just moved from `0` to `0.85`.

One real log line showing `mem usages: <value>` would settle it. Flagging it rather than asserting it's wrong, since you have the hardware output and I don't.

### Still open

**Error scanning still covers only the last node** — the fix updated the docstring ("ALL nodes" → "the last node") but `_read_last_node_log` is unchanged. The comment now matches the code, which does resolve the contradiction @amd-droy pointed at, so if single-node scanning is a deliberate scope decision for this drop that's your call. Worth being explicit that the functional consequence remains: in a 2-node run, an `NCCL ERROR` or `GPU HW ERROR` on rank 0 is never seen by `scan_for_training_errors`, and throughput is still reported from one rank only. Details in the inline comment.

### Verdict

Not quite good to go. The formatting regression is a hard blocker but a one-command fix, and the `mem_usage` units question needs a yes/no from you. Once `fmt-check` is green and `mem_usage` is confirmed to be a fraction (or re-calibrated to real units), I'm happy with this from my side — the two issues that actually prevented the suite from running are resolved, and the loss-curve gate is now genuinely doing work.

cvs/input/config_file/training/megatron/README.md
```diff
@@ -0,0 +1,185 @@
...
+Cell keys must match the format `MBS=<mbs>,GBS=<gbs>,PRECISION=<precision>`.
+
+Example cell:
+
+```json
+"MBS=4,GBS=128,PRECISION=FP8": {
+  "training.throughput_per_gpu":        { "kind": "min", "value": 100 },
+  "training.tokens_per_gpu":            { "kind": "min", "value": 1000 },
+  "training.elapsed_time_per_iteration":{ "kind": "max", "value": 500 },
+  "training.mem_usage":                 { "kind": "max", "value": 0 }
```
solaiys: Minor, but this example still carries the value that was just fixed in the real threshold files:

```
"training.mem_usage": { "kind": "max", "value": 0 }
```

Since `max: 0` fails for any positive reading, anyone copying this block out of the README reintroduces the always-fail gate. Worth syncing it to whatever `mem_usage` value you settle on.

cvs/input/config_file/training/megatron/mi325x_megatron_llama-3.3-70b_single_threshold.json
sukesh-amd: It is in the percentage the values lies between 0 and 1

cvs/input/config_file/training/megatron/README.md
sukesh-amd: Fixed

cvs/lib/training/megatron/megatron_lib.py
sukesh-amd: Fixed

cvs/lib/training/megatron/megatron_lib.py
sukesh-amd: Done

solaiys (commented):
## Re-review @ `aca6fb2b` — no blockers left from my side

All seven blockers I raised are resolved, and the `fmt-check` regression the previous fix commit introduced is gone. Verified locally at `aca6fb2b`.

### The blocker from my last pass is fixed

`0762a700` reformatted exactly the five files that were failing:

```
$ .ruff_venv/bin/ruff format --check cvs/lib/training/megatron cvs/tests/training/megatron
15 files already formatted
$ .ruff_venv/bin/ruff check ...   -> All checks passed!
$ pylint --enable=logging-too-many-args,logging-too-few-args -> 10.00/10
```

I checked the diff is behaviour-neutral — a comprehension collapsed onto one line and an import split across parens, no logic touched. Repo-wide `ruff format --check` still reports 21 files, but I cross-checked every one against this PR's changed-file list and the overlap is **zero**; that's the pre-existing `dev/dtni` debt, so this PR contributes nothing to it.

### `training.mem_usage` — resolved

Thanks for confirming the metric is a 0–1 fraction. That settles it: `max: 0.85` is a meaningful 85%-of-HBM ceiling rather than the unconditional failure the old `0` produced. All **10 cells across all 4 threshold files** are now at `0.85`, none left at `0`, and the README example in `aca6fb2b` matches. All 8 config + threshold files parse.

### `umask 077` and the targeted log read

Both landed the way I'd hoped:

- `umask 077` before the two `echo >` redirects closes the window where the wrapper script existed at the default `0644` with the token in it before `chmod 700` ran. The `chmod` was never the problem; the creation mode was.
- `_read_last_node_log` now passes `hosts=[last_host]` and returns `out_dict.get(last_host) or ''`. This drops the fan-out of a `cat` that could only ever succeed on one host, and removes the reliance on `list(out_dict.values())[-1]` happening to land on it — that was correctness-by-coincidence on dict ordering. I confirmed `hosts` is a real parameter of `ContainerOrchestrator.exec` (`container.py:630`), so this is valid.

### Loss curve still works after the reformat

Re-ran the 10-iteration simulation against the config's own `loss_curve` block (`sample_every: 10`, `max_slope: 0.0`):

| scenario | parsed | sampled | result |
|---|---|---|---|
| decreasing | 10 | 2 | PASS, slope −0.30 |
| rising | 10 | 2 | FAIL, slope +0.40 |
| flat | 10 | 2 | FAIL, slope +0.00 |

Two points on a 10-iteration run, so the test actually executes instead of silently skipping, and it discriminates in the right direction.

### Non-blocking, for whenever

- `megatron_lib.py:650` still has a commented-out `chmod 777` line — worth deleting so it doesn't get uncommented later.
- `megatron_lib.py:541` interpolates the token as `export HF_TOKEN="{self.hf_token}"` without `shlex.quote`, unlike line 417. Practically safe since HF tokens are alphanumeric, but the asymmetry is easy to fix.
- Worth a line in the README saying `mem_usage` is a 0–1 fraction. It reads as absolute next to its neighbours (`tokens_per_gpu: 1000`, `elapsed_time_per_iteration: 500` ms) — it tripped me up, so it'll likely trip up whoever calibrates these next.
- The token is still embedded in the wrapper script body, and error scanning still covers only the last rank. Both are the acknowledged follow-ups, not merge gates.

One thing I could not verify locally: full test collection fails in my scratch venv on `ModuleNotFoundError: No module named 'nodescraper'`. That's in `cvs/lib/node_scraper_adapter.py`, which this PR doesn't touch, so it's my environment rather than anything here — but it does mean I'm relying on CI for the collection gate.

Nice turnaround on these. Nothing further blocking from me — happy for this to go in.

solaiys (approved):
LGTM

amd-droy (approved):
looks good to me. Thanks Sukesh.

PR #308, opened 2026-08-11
fix(preflight): disable MI4XX fabric checks by default

speriaswamy-amd:
## Summary
- disable MI4XX IFoE fabric admission checks in the default preflight configuration
- document these checks as an MI4XX-only opt-in and keep MI3XX defaults compatible

## Test plan
- [x] `python -m json.tool cvs/input/config_file/preflight/preflight_config.json`
- [ ] `python -m unittest cvs.lib.preflight.unittests.test_scaleup_fabric` (blocked: `nodescraper` is not installed in this checkout)


urtiwari (approved):
LGTM

PR #309, opened 2026-08-12
fix(docker): use sudo_prefix() for pull_image, repair two failing UTs

atnair-amd:
## Summary

`DockerRuntime.pull_image` was the only docker invocation in the class that
hardcoded `sudo` instead of `orchestrator.sudo_prefix()`. Because
`setup_containers()` now pulls whenever `check_image_exists()` reports the image
missing, that call sits on the container-orchestrator startup path for every
suite, including the RVS and AGFHC health suites.

## Change

- `pull_image` uses `sudo_prefix()`. This removes two failure modes on clusters
  without passwordless sudo: `registry_login()` already honors `sudo_prefix()`,
  so the login authenticated as the SSH user while the pull ran as root against
  an empty `/root/.docker/config.json` — a private image then fails with "pull
  access denied" despite a clean login. Bare `sudo` also drops `-n`, letting a
  password prompt block for the full 600s timeout instead of failing fast.
- `test_pulls_image_when_missing_before_run` asserted against a `MagicMock`
  orchestrator whose `sudo_prefix()` returned a Mock rather than a string.
- `test_run_test_omits_log_file_when_not_set` did not patch
  `_validate_json_config`, so the new pre-flight check called `sys.exit` on the
  mock config path.
- Out of scope: `check_image_exists` false-negatives on untagged images (causes
  a spurious but harmless re-pull); the 19 `fmt-check` and 17 `lint` failures
  already present on `dev/dtni`.

## Tests

- `test_pull_uses_sudo_prefix_not_hardcoded_sudo` (new), pinning the rendered
  command for both the sudo and no-sudo cluster shapes.
- Gate: `.test_venv/bin/python run_all_unittests.py` — 1151 tests, 2 failures,
  both pre-existing in `cvs/lib/inference/unittests/test_inferencing_config_loader.py`
  and untouched by this branch.

### DIFF
--- a/cvs/cli_plugins/unittests/test_run_plugin.py
+++ b/cvs/cli_plugins/unittests/test_run_plugin.py
@@ -102,7 +102,8 @@ class TestRunPlugin(unittest.TestCase):
         mock_pytest_main.return_value = 0
 
         with patch.object(self.plugin, "get_test_file", return_value="/mock/path/test.py"):
-            self.plugin.run(args)
+            with patch.object(self.plugin, "_validate_json_config"):
+                self.plugin.run(args)
 
         expected_args = [
             "/mock/path/test.py",
--- a/cvs/core/runtimes/docker.py
+++ b/cvs/core/runtimes/docker.py
@@ -352,9 +352,17 @@ class DockerRuntime:
         return args
 
     def pull_image(self, image_name, timeout=None):
-        """Pull container image on all hosts."""
+        """Pull container image on all hosts.
+
+        Uses sudo_prefix() like every other docker call in this class. A
+        hardcoded `sudo` would pull as root while registry_login() -- which
+        does honor sudo_prefix() -- authenticated as the SSH user, so a private
+        image would fail with "pull access denied" on any cluster without
+        passwordless sudo. Bare `sudo` also drops the `-n`, letting a password
+        prompt block until the timeout instead of failing fast.
+        """
         timeout = timeout or 600
-        cmd = f"sudo docker pull {shlex.quote(image_name)}"
+        cmd = f"{self.orchestrator.sudo_prefix()}docker pull {shlex.quote(image_name)}"
         self.log.info(f"Pulling image on all hosts: {image_name}")
         return self.orchestrator.all.exec(cmd, timeout=timeout, detailed=True)
 
--- a/cvs/core/runtimes/unittests/test_docker.py
+++ b/cvs/core/runtimes/unittests/test_docker.py
@@ -20,6 +20,7 @@ All code contained here is Property of Advanced Micro Devices, Inc.
 #     retry, which double-runs the caller's payload whenever it fails for any
 #     reason (not just permission-denied).
 
+import shlex
 import unittest
 from unittest.mock import MagicMock, patch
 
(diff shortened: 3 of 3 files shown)


solaiys:
## Review — no blocking issues, verified locally

The `sudo_prefix()` change is correct and the two UT repairs do exactly what the description claims. Details of what I checked:

### The `pull_image` fix is the right one

`pull_image` was genuinely the odd one out. On this branch every other docker invocation in `DockerRuntime` already renders as `f"{self.orchestrator.sudo_prefix()}docker ..."` — `check_image_exists` (L20), the `sh -c` wrapper (L52), `docker run` (L173), `docker rm -f` (L179/L229), `docker ps` (L208), `docker exec` (L248/L292), and `docker load` (L371). L365 now matches all nine. `sudo_prefix()` is documented to return `'sudo -n '` or `''` (trailing space included), so the f-string interpolation is correct for both cluster shapes.

The reasoning in the description holds up: since `registry_login()` honors `sudo_prefix()`, a hardcoded `sudo docker pull` really would authenticate as the SSH user and then pull as root against a different `~/.docker/config.json`, which is a silent "pull access denied" on a private image. Worth fixing given `setup_containers()` now pulls on the startup path for every suite.

### The two UT repairs are real, not cosmetic

Confirmed against the actual merge base (`1d2641f3`):

| | tests | result |
|---|---|---|
| base `1d2641f3` | 28 | **2 failures** — `test_pulls_image_when_missing_before_run`, `test_run_test_omits_log_file_when_not_set` |
| this PR | 29 | **OK** |

Both failures are exactly the two named in the description, and both are genuine pre-existing breakage rather than fallout from this change:

- `test_pulls_image_when_missing_before_run` failed because the `MagicMock` orchestrator's `sudo_prefix()` returned a `Mock`, so the rendered command never matched. Setting `orchestrator.sudo_prefix.return_value` is the right fix.
- `test_run_test_omits_log_file_when_not_set` failed with `Expected 'exit' to be called once. Called 7 times. Calls: [call(1) x6, call(0)]` — the new `_validate_json_config` pre-flight was calling `sys.exit(1)` on the mock cluster/config paths. Patching it out is the right scope.

`test_pull_uses_sudo_prefix_not_hardcoded_sudo` is the one test added, and pinning the fully rendered command across both the `""` and `"sudo -n "` shapes is a good regression guard for precisely the class of bug being fixed here.

### Blocking issues

✅ None found. No line-level findings either — nothing in the diff needed pinning to a specific line.

I did look specifically for a concurrency angle, since `sudo_prefix()` lazily caches `_needs_sudo` and now sits on the container-startup path: a concurrent first call could double-probe, but the probe is an idempotent read and the cached value is identical either way, so there's no harmful interleaving. It's also pre-existing behaviour, not something this PR introduces.

One pre-existing wrinkle this inherits (not a change request, just noting it's now on the pull path too): `sudo_prefix()` takes the head node's answer fleet-wide, so on a mixed cluster where the head has passwordless sudo and a worker doesn't, the worker's pull renders `sudo -n docker pull` and fails. That's the documented design of `sudo_prefix()` and already true of `docker run`/`docker exec`, so it isn't newly introduced here.

Agreed with leaving `check_image_exists` false-negatives on untagged images and the pre-existing `fmt-check`/`lint` debt out of scope.

solaiys (approved):
LGTM

PR #184, opened 2026-05-21
Improve PyTorch XDit benchmark workflow

speriaswamy-amd:
## Summary
- tighten PyTorch XDit benchmark preflight and output validation
- clarify the public README and sample configs for model staging and threshold placeholders
- add/update unit tests and rebase follow-up fixes needed to keep the branch green

## Test plan
- [x] `make fmt`
- [x] `make lint`
- [x] `.test_venv/bin/python run_all_unittests.py`
- [x] `CVS=\".test_venv/bin/cvs\" ./test_cli.sh`

Made with [Cursor](https://cursor.com)

PR #314, opened 2026-08-14
chore(packaging): add missing __init__.py for cvs.tests subpackages

speriaswamy-amd:
Part 1 of 12 in a stack that replaces #184. Base: `main`.

### Why
`setup.py` builds its package list with `find_packages(where="cvs")`, which only walks directories containing `__init__.py`. Most `cvs/tests/*` suite directories have none, so they are silently dropped from the sdist and from `pip install` / `pip install -e .`.

### What changed
- Added 15 empty `__init__.py` files under `cvs/tests/`.

This makes 17 subpackages discoverable. Two of them — `tests.inference.pytorch_xdit` and `tests.training.jaxmaxtext` — already had their own `__init__.py` and were excluded only because a parent directory did not.

No behaviour change for anyone running from a source checkout.

### DIFF
--- a/cvs/tests/ibperf/ib_perf_bw_test.py
+++ b/cvs/tests/ibperf/ib_perf_bw_test.py
@@ -262,7 +262,7 @@ def test_ib_bw_perf(shdl, phdl, bw_test, config_dict):
                     config_dict['expected_results'],
                 )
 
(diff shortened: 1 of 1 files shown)


amd-droy (approved):
looks good to me. Thanks Surya.

PR #315, opened 2026-08-14
fix(tests): make missing-file test self-contained

speriaswamy-amd:
Part 2 of 12 in a stack that replaces #184. Base: #314.

### Why
`test_missing_file_raises` asserted `FileNotFoundError` against the hardcoded absolute path `/nonexistent/path/cluster.json`. That is environment-dependent (AGENTS.md forbids it) and would stop testing what it claims to test if such a path ever existed.

### What changed
- `cvs/monitors/unittests/test_check_cluster_health.py` — build the missing path inside a `TemporaryDirectory`, matching the sibling `test_invalid_json_raises_value_error`.
- `cvs/tests/ibperf/ib_perf_bw_test.py` — one log separator `%%%%%%%%%` → `=========`.

The ibperf line is unrelated drive-by cleanup. The same `%%%` style appears elsewhere in `ib_perf_bw_test.py`, `rccl_lib.py` and `mori_lib.py` and is untouched, so this leaves the file internally inconsistent — happy to drop that hunk if reviewers prefer.

### DIFF
--- a/cvs/monitors/unittests/test_check_cluster_health.py
+++ b/cvs/monitors/unittests/test_check_cluster_health.py
@@ -61,8 +61,10 @@ class TestLoadClusterFile(unittest.TestCase):
         self.assertEqual(pkey, "/home/bob/.ssh/id_rsa")
 
     def test_missing_file_raises(self):
-        with self.assertRaises(FileNotFoundError):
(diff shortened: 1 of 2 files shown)


amd-droy (approved):
lgtm.

PR #316, opened 2026-08-14
fix(utils_lib): detect MI350 by device ID when rocm-smi omits the model name

speriaswamy-amd:
Part 3 of 12 in a stack that replaces #184. Base: #315.

### Why
On some driver / rocm-smi versions an MI350 node reports no marketing-name string anywhere in `rocm-smi -a` — only `Device ID: 0x75a0` and `GFX Version: gfx950`. `get_model_from_rocm_smi_output` then fell through to its conservative `'mi300x'` default, so MI350 nodes silently picked the MI300X entry in `expected_results` and were graded against the wrong performance threshold.

### What changed
- `cvs/lib/utils_lib.py` — added a `Device ID: 0x75a0` fallback *after* the existing marketing-name checks, so the name-based path is unchanged whenever a name is present.
- `cvs/lib/unittests/test_utils_lib.py` — tests for all four marketing names, the device-ID fallback, and the unrecognized-input default.

### DIFF
--- a/cvs/lib/utils_lib.py
+++ b/cvs/lib/utils_lib.py
@@ -205,9 +205,14 @@ def get_model_from_rocm_smi_output(smi_output):
       - Some driver/rocm-smi versions emit no marketing-name string at all (observed on
         MI350: only "Device ID: 0x75a0" and "GFX Version: gfx950", no "MI350" substring
         anywhere in `rocm-smi -a`); in that case, falls back to matching the known Device ID.
+        MI355X is affected by the same omission (ROCM-21360), hence the second entry below.
       - Falls back to 'mi300x' if nothing matches (conservative default).
 
     """
+    device_id_to_model = {
+        '0x75a0': 'mi350',
+        '0x75a3': 'mi355',
+    }
     if re.search('MI300X', smi_output, re.I):
         model = 'mi300x'
     elif re.search('MI325', smi_output, re.I):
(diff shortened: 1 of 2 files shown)


cvs/lib/utils_lib.py
```diff
@@ -213,6 +216,8 @@ def get_model_from_rocm_smi_output(smi_output):
         model = 'mi350'
     elif re.search('MI355', smi_output, re.I):
         model = 'mi355'
```
amd-droy, line 223: MI355 might have the same issue, might need this snippet -

elif re.search(r'Device ID:\s*0x75a3', smi_output, re.I):
    model = 'mi355'

Same marketing-name omission (ROCM-21360) also affects MI355X

cvs/lib/utils_lib.py
```diff
@@ -213,6 +216,8 @@ def get_model_from_rocm_smi_output(smi_output):
...
+    elif re.search(r'Device ID:\s*0x75a0', smi_output, re.I):
+        model = 'mi350'
```
amd-droy: If the series will add more device-ID mappings, a small dict (DEVICE_ID_TO_MODEL = {"0x75a0": "mi350", ...}) might be more helpful.

cvs/lib/utils_lib.py
speriaswamy-amd: Addressed, with latest commit, thank you for bringing this up

amd-droy (approved):
lgtm. Thank you.

PR #317, opened 2026-08-14
fix(pssh): retry once with a fresh client on SessionError

speriaswamy-amd:
Part 4 of 12 in a stack that replaces #184. Base: #316.

### Why
Long-running suites reuse a single `ParallelSSHClient` for the whole module. If the underlying session goes stale — common after a benchmark step that leaves the connection idle for tens of minutes — the next `run_command` raises `SessionError` and the test fails on a trivial follow-up command rather than on anything it was validating.

### What changed
- `_recreate_parallel_client()` — the client-construction block lifted out of `__init__` verbatim; `__init__` now calls it, so there is one construction path.
- `_run_command_with_session_retry()` — logs, rebuilds the client, retries **once**. A second failure propagates unchanged, so a genuinely broken host still fails fast.
- `exec()` and `exec_cmd_list()` route their `run_command` calls through it.

### Tests
The two `setUp` methods move from a decorator-scoped `@patch` to `patcher.start()` + `addCleanup`, because the retry constructs a client during the test body, not just during `setUp`. New coverage: retry-then-succeed (with and without a timeout) and retry-then-fail, on both exec paths.

### DIFF
--- a/cvs/lib/parallel/pssh.py
+++ b/cvs/lib/parallel/pssh.py
@@ -105,6 +105,7 @@ class Pssh:
                 e,
             )
             self.log.debug("ParallelSSH session retry detail", exc_info=True)
+            self.destroy_clients()
             self._recreate_parallel_client()
             return self.client.run_command(*args, **kwargs)
 
--- a/cvs/lib/parallel/pssh.py
+++ b/cvs/lib/parallel/pssh.py
@@ -80,6 +80,10 @@ class Pssh:
             self.log.info("%s", self.reachable_hosts)
             self.log.info("%s", self.user)
             self.log.info("%s", self.pkey)
+        self._recreate_parallel_client()
+
+    def _recreate_parallel_client(self):
+        if self.password is None:
             self.client = ParallelSSHClient(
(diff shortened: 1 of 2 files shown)


cvs/lib/parallel/pssh.py
```diff
@@ -92,6 +96,18 @@ def __init__(
...
+    def _run_command_with_session_retry(self, *args, **kwargs):
+        try:
+            return self.client.run_command(*args, **kwargs)
+        except SessionError as e:
+            self.log.info(
+                "ParallelSSH: SessionError on first attempt; recreating client and retrying once (%s).",
+                e,
+            )
+            self.log.debug("ParallelSSH session retry detail", exc_info=True)
+            self._recreate_parallel_client()
```
amd-droy, line 109: Consider calling cleanup on the old client before recreating, so a stale session retry doesn’t leave orphaned SSH connections on long-running module-scoped Pssh instances - function destroy_clients() already exists.

amd-droy (approved):
lgtm. thanks

PR #318, opened 2026-08-14
feat(pssh): support inactivity_timeout in exec_cmd_list

speriaswamy-amd:
Part 5 of 12 in a stack that replaces #184. Base: #317.

### Why
`exec()` already accepts `inactivity_timeout` (a per-line silence timer). `timeout` is parallel-ssh's `read_timeout` — a **total** wall-clock cap that activity does not reset. `exec_cmd_list` only had the latter, so callers running long per-host commands had no way to distinguish "still working" from "hung": the cap has to be set to the worst plausible runtime, and a genuine hang then burns that entire budget.

### What changed
- `Pssh.exec_cmd_list` accepts `inactivity_timeout` and disables `read_timeout` when it is set.
- `MultiProcessPssh.exec_cmd_list` threads it through both the direct and the sharded worker path, so behaviour is identical regardless of host count.
- Docstring spells out the difference between the two parameters.

Defaults to `None`; existing `timeout` callers are unaffected.

### Known gap
No unit test covers the `exec_cmd_list` inactivity path specifically — the underlying per-line timer in `_process_output` is already exercised via `exec`. Happy to add one here if reviewers want it before merge.

### DIFF
--- a/cvs/lib/parallel/pssh.py
+++ b/cvs/lib/parallel/pssh.py
@@ -326,6 +326,9 @@ class Pssh:
         inactivity_timeout is set, the underlying read_timeout is disabled so there
         is no total cap. Pass at most one of the two.
         """
+        if timeout is not None and inactivity_timeout is not None:
+            raise ValueError("Pass at most one of timeout and inactivity_timeout, not both.")
+
         if self.env_prefix:
             full_cmd = f"{self.env_prefix} ; {cmd}"
         else:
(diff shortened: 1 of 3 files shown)


cvs/lib/parallel/pssh.py
```diff
@@ -371,11 +371,19 @@ def exec(self, cmd, timeout=None, print_console=True, detailed=False, inactivity
...
         which runs the same command on all hosts.
         Returns a dictionary of host as key and command output as values
+
+        timeout is parallel-ssh's read_timeout: a TOTAL wall-clock cap on reading
+        the command's output (it is NOT reset by activity). inactivity_timeout is
+        an alternative that measures the gap between output lines instead: the
+        command may run arbitrarily long as long as it keeps producing output, and
+        is aborted only after inactivity_timeout seconds of silence. When
+        inactivity_timeout is set, the underlying read_timeout is disabled so there
+        is no total cap. Pass at most one of the two.
```
amd-droy, line 390: Docstring says “Pass at most one of the two,” but if a caller passes both, inactivity_timeout silently wins and timeout is ignored. Consider either validating and raising ValueError, or adding a one-line note that inactivity_timeout takes precedence when both are set.

amd-droy (approved):
lgtm. thanks

PR #319, opened 2026-08-14
feat(docker_lib): add image-presence preflight and a prune opt-out

speriaswamy-amd:
Part 6 of 12 in a stack that replaces #184. Base: #318.

### Why
Two small library gaps the pytorch_xdit benchmarks hit, both usable by any suite.

### What changed
**`nodes_missing_docker_image(phdl, image_ref)`** — runs `docker image inspect` on every target and returns the hosts where the image is absent. Without it, a missing image surfaces as a `docker run` failure minutes into a test, buried in container output. The image ref is `shlex`-quoted.

**`delete_all_containers_and_volumes()`** now honours `CVS_PYTORCH_XDIT_SKIP_DOCKER_SYSTEM_PRUNE` (`1`/`true`/`yes`/`on`). The function runs `docker system prune --force`, which removes stopped containers and unused networks belonging to *other users* on shared cluster nodes. The opt-out lets a run stop its own named container and skip the cluster-wide prune. Default behaviour is unchanged.

Unit tests cover both, including the missing/present cases and the skip path.

### DIFF
--- a/cvs/lib/docker_lib.py
+++ b/cvs/lib/docker_lib.py
@@ -5,8 +5,10 @@ The year included in the foregoing notice is the year of creation of the work.
 All code contained here is Property of Advanced Micro Devices, Inc.
 '''
 
-import re
 import json
+import os
+import re
+import shlex
 import time
 
 from cvs.lib import globals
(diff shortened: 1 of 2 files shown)


cvs/lib/docker_lib.py
```diff
@@ -45,6 +63,22 @@ def kill_docker_container(phdl, container_name):
 
 
 def delete_all_containers_and_volumes(phdl):
+    """
+    Prune unused Docker objects on all target hosts.
+
+    This runs ``docker system prune --force`` (not ``-a``), but it can still remove
+    stopped containers and unused networks on shared nodes. Set environment variable
+    ``CVS_PYTORCH_XDIT_SKIP_DOCKER_SYSTEM_PRUNE`` to ``1``/``true``/``yes`` to skip this
```
amd-droy: isn't this function used by most of the frameworks, can this 'CVS_PYTORCH_XDIT_SKIP_DOCKER_SYSTEM_PRUNE' be generalized to 'CVS_SKIP_DOCKER_SYSTEM_PRUNE'

amd-droy (approved):
lgtm. thanks.

PR #320, opened 2026-08-14
fix(pytorch_xdit): key benchmark output dirs on the cluster target, not hostname

speriaswamy-amd:
Part 7 of 12 in a stack that replaces #184. Base: #319.

### Why
Both benchmarks named their per-node output directory from the remote `hostname` command (`wan_22_${hostname}_outputs`, `flux_${hostname}_outputs`), and the run step and the parse step each resolved it *independently*. When a node reports a short name in one context and an FQDN in another, the two steps disagree and parse fails with "no results found" even though the run wrote them — and results scatter across several directory trees for what is one node.

The cluster `node_dict` key — the SSH target string already in `cluster.json` — is stable by construction and needs no remote call.

### What changed
- `utils_lib.cluster_target_output_label()` — normalizes the key (strips whitespace, `/` → `_`, falls back to `unknown_node`). Unit tested.
- Both modules derive output directories from that label in the run step and the parse step, including WAN's multi-node aggregation filter, which previously ran a second `hostname` probe to build its expected-directory list.
- Pins `_test_output_dir` for single-node WAN runs, matching what FLUX already did, so parse does not re-derive the path at all in the common case.

`hostname` is still collected, but only for the log line.

### DIFF
--- a/cvs/lib/unittests/test_utils_lib.py
+++ b/cvs/lib/unittests/test_utils_lib.py
@@ -22,6 +22,11 @@ class TestUtilsLib(unittest.TestCase):
         utils_lib.scan_test_results(out_dict)
         mock_fail_test.assert_not_called()
 
+    def test_cluster_target_output_label_strips_and_sanitizes(self):
+        self.assertEqual(utils_lib.cluster_target_output_label("  node1.example.com  "), "node1.example.com")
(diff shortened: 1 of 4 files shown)


amd-droy (approved):
lgtm. thanks.

PR #321, opened 2026-08-14
fix(pytorch_xdit): fail fast on missing image and on incomplete WAN caches

speriaswamy-amd:
Part 8 of 12 in a stack that replaces #184. Base: #320.

### Why
Two preflight gaps, both of which turned a clear setup problem into a confusing downstream failure.

### What changed

**Missing container image.** Neither benchmark checked that `config.container_image` was present locally before `docker run`. Both now call `docker_lib.nodes_missing_docker_image()` and fail with the node list and the `docker pull` command to fix it. WAN also gains the `/dev/kfd` check FLUX already had.

**Incomplete WAN snapshot.** Verification only ran `test -d` on the snapshot directory — which exists as soon as `hf download` *starts*, so a half-downloaded cache passed and the failure surfaced ~20 minutes later inside the container. New `utils_lib.wan_hf_snapshot_offline_check_commands()` returns per-file OK/MISSING shell checks for the Wan2.2-I2V-A14B layout: three `config.json` files, six >500MiB diffusion shards in each of the low/high-noise dirs, and size floors on the VAE and T5 weights so LFS pointer stubs do not count as present.

The checks use `find -L`. A `hf_hub` cache stores weights as symlinks into `blobs/`, so `find` without `-L` stats the link (tiny, type `l`) and reports **zero** shards for a fully correct cache. Verified against a real ~118GB download: without `-L` the low-noise shard count is 0, with `-L` it is 6.

They run in both staging modes (explicit host path and `hf_home` cache) and only when the repo id or path names `Wan2.2-I2V-A14B`, so other WAN forks and custom layouts are unaffected.

Unit test asserts the generated commands quote paths containing spaces and emit both branches.


cvs/lib/utils_lib.py
```diff
@@ -448,6 +449,50 @@ def cluster_target_output_label(cluster_node_key: str) -> str:
...
+    """
+    Remote-shell checks for a Wan2.2 I2V-A14B-style Hugging Face snapshot (host path).
+
+    Each command prints OK on success and MISSING on failure. This catches the common
+    false positive where the snapshot directory exists while `hf download` is still
+    running or LFS pointer files were not fully materialized.
+
+    Intended for snapshots of `Wan-AI/Wan2.2-I2V-A14B` (and forks with the same tree).
+    Callers should only run these when `model_repo` is known to use this layout.
+    """
```
amd-droy, line 494: Hardcoded layout is brittle by design. Exactly 6 shards, specific filenames, and fixed size thresholds are fine for this model, but worth a short comment that these constants mirror Wan-AI/Wan2.2-I2V-A14B as of the tested revision - if Hugging Face changes the repo layout later, e.g. splits weights into 8 shards instead of 6, renames files, or changes sizes — a fully valid download could start failing these checks until someone updates the helper.

cvs/tests/inference/pytorch_xdit/pytorch_xdit_wan22_14b_single.py
```diff
@@ -361,6 +362,20 @@ def test_verify_hf_cache_or_download(s_phdl, inference_dict, hf_token):
             update_test_result()
             return
 
+        # Align with HF snapshot checks for Wan2.2 I2V-A14B when staging via absolute path.
+        if "Wan2.2-I2V-A14B" in model_repo or "Wan2.2-I2V-A14B" in host_model_path:
+            for label, cmd in wan_hf_snapshot_offline_check_commands(host_model_path).items():
```
amd-droy: This loop (lines 367–377) is nearly identical to the HF-cache branch at 413–423 — same exec/bad-node/fail/return pattern, only the path and error prefix differ. Consider a local helper like _assert_wan_snapshot_complete(s_phdl, snapshot_path, failure_prefix) to avoid the two copies drifting apart.

amd-droy (approved):
lgtm. thanks.

PR #323, opened 2026-08-14
fix(pytorch_xdit): stop leaking the HF token from the WAN benchmark

speriaswamy-amd:
Part 10 of 12 in a stack that replaces #184. Base: #322.

### Why
Two ways the HF token could reach a log or a test report from the WAN module. Both are already handled in the FLUX module; this brings WAN in line.

### What changed

**`_redact_secrets` was a no-op.** The pattern was `r"(HF_TOKEN=)\\S+"` — inside a raw string that is a literal backslash followed by `S`, not the `\S` character class. It never matched, so every logged docker command carried the token in clear text. Corrected to `r"(HF_TOKEN=)\S+"` (FLUX uses the equivalent `[^\s]+`).

**The `hf_token` fixture returned a bare `str`.** Pytest includes fixture values in failure reports, so any failure in a test taking `hf_token` printed the token. It now returns `_SecretValue`, the wrapper FLUX already defines: truthy, stringifies to the token for command assembly, but its `repr` is `<redacted>`. Call sites unchanged.


amd-droy (approved):
lgtm. thanks.

PR #324, opened 2026-08-14
fix(pytorch_xdit): replace placeholder thresholds with measured values; fix cache paths

speriaswamy-amd:
Part 11 of 12 in a stack that replaces #184. Base: #323.

### Thresholds (AIMVT-88)
The shipped `mi300x` thresholds were unvalidated placeholders and one was wrong by more than an order of magnitude: `wan22_i2v_a14b` set `max_avg_total_time_s: 12.0`, while a real MI300X run with `compile: true` averages **161.03 s/step** (5 clean steps: 160.98, 161.03, 160.98, 161.06, 161.11). Every real run on that hardware failed the threshold.

| benchmark | gpu | before | after | basis |
|---|---|---|---|---|
| `wan22_i2v_a14b` | mi300x | 12.0 | **200.0** | 161.03 s/step measured, ~25% margin |
| `wan22_i2v_a14b` | mi350 | — | **120.0** | 86.10–90.99 s/step across two nodes |
| `flux1_dev_t2i` | mi300x | 8.0 | **3.0** | 1.39 s/pipe over 25 reps, ~2x margin |
| `flux1_dev_t2i` | mi350 | — | **2.0** | 0.92 s/pipe, consistent across two runs |

`auto` and `mi355` are unchanged and remain unvalidated. Each config gains an `_expected_results_comment` recording which numbers are measured, on what hardware, with what margin — and which are not.

### Schema
`PytorchXditWan22Benchmarks` and `PytorchXditFlux1DevBenchmarks` use `extra="forbid"`, so the comment field would make every test in the module fail validation before running. Both switch to `extra="allow"`, matching the other schemas in the same file that already permit comment fields.

### Cache paths
`hf_home` was `/home/{user-id}`, but `HF_HOME` must point at the root containing `hub/` — the test builds `hub/models--*/snapshots/<rev>` underneath it, which does not exist directly in a home directory. Corrected to `{home}/.cache/huggingface`, the `hf` CLI default. `hf_token_file` and `output_base_dir` move from `/home/{user-id}/...` to `{home}/...`, which resolves correctly where homes are not under `/home`.


amd-droy (approved):
lgtm. thank you.

PR #325, opened 2026-08-14
docs(pytorch_xdit): add a first-time setup guide and correct stale instructions

speriaswamy-amd:
Part 12 of 12 in a stack that replaces #184. Base: #324.

Docs only. The README described a workflow that did not reproduce.

### What changed
Reworked around a numbered first-run path — connect to the node, install the `hf` CLI, set `HF_HOME`, `hf download` each model, `docker pull` the image, run sanity checks, write a minimal single-node `cluster.json`, run the module — with the exact commands.

**Corrections:**
- WAN cache size was "~40GB"; a real `hf download` of `Wan-AI/Wan2.2-I2V-A14B` is ~118GB.
- Config examples used `hf_home: "/home/{user-id}"`, which is not a valid HF cache root; they now match the shipped configs.
- Output paths were documented as `${output_base_dir}/wan_22_${hostname}_outputs`; directories are now named from the cluster SSH target (#320).
- Site-specific SLURM commands (a named reservation and account) were presented as *the* way to get a node. Replaced with the actual requirement: a target with Docker, the image pulled, and ROCm device nodes.

**New sections:** the multi-node `hf_home` caveat (staged per node, so an unshared `$HOME` means one full copy per node — AIMVT-81/AIMVT-130), the `CVS_PYTORCH_XDIT_SKIP_DOCKER_SYSTEM_PRUNE` opt-out, which placeholders resolve in test configs vs cluster files, and troubleshooting entries for the new missing-image and incomplete-snapshot failures.

### DIFF
--- a/cvs/tests/inference/pytorch_xdit/README.md
+++ b/cvs/tests/inference/pytorch_xdit/README.md
@@ -2,7 +2,169 @@
 
 This directory contains inference microbenchmark tests for PyTorch XDit models (WAN, Flux, etc.).
 
-**IMPORTANT**: Always run these tests on SLURM compute nodes, never on login nodes (shared resources).
+**IMPORTANT**: Use GPU-capable targets with the drivers and device nodes this workload expects (for example ROCm and `/dev/kfd` for these tests). Do not run GPU benchmarks on machines that are not set up for them.
+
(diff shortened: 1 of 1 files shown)


amd-droy (approved):
lgtm.

PR #171, opened 2026-05-14
feat(aorta): multi-node disaggregated launch via single cluster.json (AIMVT-173)

speriaswamy-amd:
## Summary

Implements [AIMVT-173](https://ontrack-internal.amd.com/browse/AIMVT-173): run the CVS Aorta benchmark across N nodes from a single `cluster.json`, mirroring the disaggregated launch pattern used by the existing PyTorch xDiT and SGLang multi-node test suites.

The `AortaRunner` now orchestrates `torchrun` on every node in parallel (rendezvousing on the head), and consolidates per-node `torch_profiler` trees into `<aorta_path>/combined_traces/node_<rank>/` so the host parser sees a single unified set. Single-node behavior is unchanged: `multi_node.master_launch_mode='auto'` keeps the legacy `experiment_script` path for 1-node clusters, and pre-existing yamls without a `multi_node:` block still validate (Pydantic supplies sensible defaults).

## What's in the diff

- `cvs/runners/aorta.py` — `AortaMultiNodeConfig` dataclass; `_resolve_launch_mode`, `_pick_master_port`, `_build_torchrun_command`, `_run_single_node`, `_collect_multi_node_traces` + local/remote copy helpers; refactored `run()`
- `cvs/parsers/schemas.py` — `AortaMultiNodeConfigFile` Pydantic schema + `train_script` existence check
- `cvs/input/config_file/aorta/aorta_benchmark.yaml` — new `multi_node:` block with inline docs
- `docs/reference/configuration-files/aorta.rst` — new "Multi-node disaggregated launch" section + parameter table
- `cvs/runners/unittests/test_aorta_multinode.py` — 24 unit tests (launch-mode resolution, port selection, command construction, env merging, trace-tree copy, schema validation, single `--override` group invariant)
- `cvs/tests/benchmark/test_aorta.py` — wire `multi_node` block through the runner-config fixture

## Validation

End-to-end `cvs run test_aorta` against a real 2-node cluster (`g17u19` head + `f16u13` worker, 16xMI300X total) — **5/5 pytest cases pass in 148s**, traces collected from both nodes, host parser produced metrics for all 16 ranks. Four runtime bugs surfaced and were fixed during this validation:

1. Container ran as `jenkins` UID and couldn't open `/dev/kfd` despite `--privileged` → now passes `user="root"` and `group_add=["video","render"]`
2. `UnboundLocalError: trace_mtime` in the freshest-trace selector → initialised upfront
[...]

PR #326, opened 2026-08-14
fix(runner): tear down resources after a partially-successful setup

speriaswamy-amd:
**Stack 1/6** — splits #171 into reviewable pieces. Base: `main`.

### Why

`BaseRunner.execute()` only set `_setup_complete` after `setup()` returned `True`, so the `finally` block's `teardown()` was skipped whenever setup failed. A multi-node setup that launched containers on nodes 0..N-1 before failing on node N leaked every container it had already created.

### What changed

- `cvs/runners/_base_runner.py` — set `_setup_complete` as soon as `setup()` has been *attempted*, so teardown runs for partial setups. Existing `teardown()` implementations already tolerate partially-populated state.
- `cvs/runners/unittests/` — new package (`cvs/runners/` had no unit tests); 4 cases pinning the lifecycle.
- `.gitignore` — add `.venv/`.

### Test

`ruff check` / `ruff format --check` clean. Unit tests 585 → 589.

### DIFF
--- a/cvs/runners/_base_runner.py
+++ b/cvs/runners/_base_runner.py
@@ -184,11 +184,14 @@ class BaseRunner(ABC):
         try:
             # Setup phase
             log.info(f"Setting up {self.__class__.__name__}...")
-            if not self.setup():
+            setup_ok = self.setup()
+            # Set even on partial/failed setup so the finally block below still
+            # tears down resources a partially-successful setup already created.
+            self._setup_complete = True
+            if not setup_ok:
                 return RunResult(
                     status=RunStatus.FAILED, start_time=start_time, end_time=time.time(), error_message="Setup failed"
                 )
-            self._setup_complete = True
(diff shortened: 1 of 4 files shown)


cvs/runners/_base_runner.py
```diff
@@ -184,11 +184,14 @@ def execute(self, **kwargs) -> RunResult:
         try:
             # Setup phase
             log.info(f"Setting up {self.__class__.__name__}...")
-            if not self.setup():
+            setup_ok = self.setup()
```
amd-droy: If setup() raises after creating resources, _setup_complete is never set and finally still skips teardown, so you can leak the same way as before. Set the flag in a finally around setup() (or always call teardown()), and add a unit test where setup() raises.

amd-droy (approved):
lgtm.thanks

PR #327, opened 2026-08-14
fix(aorta): container /dev/kfd access and --override argument packing

speriaswamy-amd:
**Stack 2/6** — splits #171. Base: #326.

### Why

Two defects on the **existing single-node path**, both surfaced by live 2-node validation but neither multi-node specific.

1. `_launch_container()` never passed `user`, so the container ran as the image's default UID (`jenkins` on the validation cluster) and could not open `/dev/kfd` despite `--privileged`. The `render` group was also missing. The function's own comment already claimed *"Run as root so the container can access GPUs"* — the code now matches it.
2. Training overrides were emitted as `--override k=v --override k2=v2`. Aorta's `train.py` declares `--override` with `nargs="*"`, so argparse keeps only the **last** group — every preceding override was silently dropped.

### What changed

- `cvs/runners/aorta.py` — `user="root"`, `group_add=["video", "render"]`; all `key=value` tokens packed behind a single `--override`.
- Extracted the env-building block out of `run()` into `_build_base_env()` so the override packing is unit-testable without a container.

### Test

`ruff` clean. Unit tests 589 → 595 (6 new: env construction, single-`--override` invariant, container user/groups).

### DIFF
--- a/cvs/runners/aorta.py
+++ b/cvs/runners/aorta.py
@@ -308,7 +308,8 @@ class AortaRunner(BaseRunner):
             volumes=volumes,
             devices=devices,
             working_dir=self.config.container_mount_path,
-            group_add=["video"],
+            user="root",
+            group_add=["video", "render"],
             cap_add=["SYS_PTRACE"],
             security_opt=["seccomp=unconfined"],
             ulimits=[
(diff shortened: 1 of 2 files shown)


cvs/runners/aorta.py
```diff
@@ -308,7 +308,8 @@ def _launch_container(self, client: docker.DockerClient, node: str) -> Container
             volumes=volumes,
             devices=devices,
             working_dir=self.config.container_mount_path,
-            group_add=["video"],
+            user="root",
+            group_add=["video", "render"],
```
amd-droy: I think hardcoding render will fail containers.run on hosts that have no such group. With user="root" and privileged=True, /dev/kfd is usually already accessible, you can keep render only if you still need it, and skip or resolve the group when it is missing.

cvs/runners/aorta.py
```diff
@@ -553,6 +554,27 @@ def setup(self) -> bool:
...
+            f"/opt/openmpi/lib:/opt/rccl-tests/build:$LD_LIBRARY_PATH"
+        )
+        env["rccl_path"] = rccl_path
+
+        if self.config.training_overrides:
+            # aorta train.py exposes `--override` with `nargs="*"`; multiple
+            # `--override` groups collapse to the last group's values. Emit a
+            # single `--override` followed by all key=value tokens so that
+            # downstream legacy launch scripts also forward them correctly.
+            tokens = " ".join(f'{key}="{value}"' for key, value in self.config.training_overrides.items())
```
amd-droy: Those embedded quotes become literal characters in AORTA_OVERRIDE_ARGS, so Hydra/argparse often sees training.max_steps="15" instead of training.max_steps=15, and values with spaces still split. Emit key=value tokens (quote only when the value needs it) and assert the full env string in the unit test.

amd-droy (approved):
lgtm. thanks.

PR #328, opened 2026-08-14
feat(aorta): multi_node configuration surface (schema, yaml, docs)

speriaswamy-amd:
**Stack 3/6** — splits #171. Base: #327.

### Why

Declares the `multi_node:` block that the disaggregated torchrun launch consumes. **Config plumbing only — nothing reads these values yet, so behavior is unchanged.** Split out so the launch logic in #329 is reviewable on its own; review this one for names, defaults, and docs.

### What changed

- `cvs/parsers/schemas.py` — `AortaMultiNodeConfigFile` (`extra="forbid"`, `master_launch_mode` restricted to `auto`/`script`/`torchrun`), wired in with a default factory so yamls predating the block still validate. `validate_paths_exist()` checks `train_script` only when the mode is explicitly `torchrun`.
- `cvs/runners/aorta.py` — matching `AortaMultiNodeConfig` dataclass, plus `AortaConfig.node_vpc_ips` so `master_addr` can later resolve to the head node's RDMA-fabric address rather than its mgmt/SSH address (the same `node_dict` vs `vpc_ip` split `rccl_perf.py` already uses).
- `aorta_benchmark.yaml` / `docs/.../aorta.rst` — the block with inline docs and a parameter table.
- `cvs/tests/benchmark/test_aorta.py` — fixture maps the validated block onto the dataclass and populates `node_vpc_ips` from the cluster file.

### Test

`ruff` clean. Unit tests 595 → 603. Adds `cvs/parsers/unittests/` (per AGENTS.md, the package had none): schema defaults, unknown-key and bad-mode rejection, conditional `train_script` check, and backward compat for configs with no `multi_node:` block.

### DIFF
--- a/cvs/parsers/schemas.py
+++ b/cvs/parsers/schemas.py
@@ -466,6 +466,96 @@ class AortaAnalysisConfigFile(BaseModel):
     )
 
 
+class AortaMultiNodeConfigFile(BaseModel):
+    """
+    Schema for the optional ``multi_node`` section in aorta_benchmark.yaml.
+
+    When the cluster file contains more than one node, the runner launches a
+    disaggregated ``torchrun`` invocation on every node (one per ``node_dict``
+    entry), rendezvous-ing on the head node. This block tunes that path.
+
+    Single-node clusters ignore this block; ``master_launch_mode`` defaults to
+    ``auto`` which means: ``script`` (current behavior, single-node) when the
+    cluster has one node, ``torchrun`` (multi-node disaggregated) when it has
+    more than one.
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    master_launch_mode: str = Field(
+        default="auto",
+        description=(
+            "How the experiment is launched on each node. 'auto' picks 'script' "
+            "for single-node clusters and 'torchrun' for multi-node clusters. "
+            "'script' always uses the configured experiment_script (single-node only). "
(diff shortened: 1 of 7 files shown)


speriaswamy-amd:
Closing/reopening to force GitHub to recompute mergeable state (stuck as 'dirty' despite a verified clean fast-forward merge locally).

cvs/parsers/schemas.py
```diff
@@ -466,6 +466,96 @@ class AortaAnalysisConfigFile(BaseModel):
...
+
+    When the cluster file contains more than one node, the runner launches a
+    disaggregated ``torchrun`` invocation on every node (one per ``node_dict``
+    entry), rendezvous-ing on the head node. This block tunes that path.
+
+    Single-node clusters ignore this block; ``master_launch_mode`` defaults to
+    ``auto`` which means: ``script`` (current behavior, single-node) when the
+    cluster has one node, ``torchrun`` (multi-node disaggregated) when it has
+    more than one.
+    """
```
amd-droy, line 482: master_launch_mode: torchrun on one node still uses train_script / extra_env / master_addr. The master_addr text also says it defaults to the cluster-file hostname, but node_vpc_ips is there specifically to prefer vpc_ip. People might pin the SSH address and break fabric rendezvous if not documented.

amd-droy (approved):
lgtm.

PR #330, opened 2026-08-14
fix(aorta): bound setup and run by timeout_seconds on daemon threads

speriaswamy-amd:
**Stack 5/6** — splits #171. Base: #390.

### Why

Neither `setup()` nor `run()` put a deadline on their parallel per-node work, so a single stalled node hung the entire CVS invocation with no result. Hit live: an NCCL collective stalled on one node of a 2-node run and the process sat there indefinitely instead of reporting `TIMEOUT`.

`ThreadPoolExecutor` cannot fix this on its own — its workers are non-daemon, and CPython's atexit hook joins every one of them at interpreter shutdown regardless of `shutdown(wait=False)`. A node stuck in a blocking Docker or SSH call would still wedge the process on the way out.

### What changed

- `_run_bounded_parallel()` — runs each task on a **daemon** thread against a shared deadline, returning per-key results / errors / timeouts. Abandoned threads cannot block exit. Replaces `ThreadPoolExecutor` in both `setup()` and `run()`.
- `_setup_single_node()` takes a cancel `Event`: if it finishes launching its container after `setup()` has given up, it tears that container down itself. Otherwise it would register into `self._containers` *after* `teardown()`'s one-time snapshot had already run, orphaning the container on the node. (Builds on #326.)

### Test

`ruff` clean. Unit tests 623 → 631 (8 new, including a case asserting the worker thread is actually a daemon, and end-to-end `setup()`/`run()` timeout cases that assert the call returns promptly).

### DIFF
--- a/cvs/runners/aorta.py
+++ b/cvs/runners/aorta.py
@@ -14,12 +14,11 @@ import logging
 import shlex
 import subprocess
 import time
-from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass, field
 from functools import partial
 from pathlib import Path
-from threading import Lock
-from typing import Any, Dict, List, Optional, Tuple, TYPE_CHECKING
+from threading import Event, Lock, Thread
+from typing import Any, Callable, Dict, List, Optional, Tuple, TYPE_CHECKING
 
 if TYPE_CHECKING:
     import docker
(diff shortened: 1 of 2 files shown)


amd-droy (approved):
looks good. thanks.

amd-droy (approved):

amd-droy (approved):

PR #331, opened 2026-08-14
feat(aorta): collect per-node traces, survive partial node failure

speriaswamy-amd:
**Stack 6/6** — splits #171. Base: #330. Final PR; the tip of this stack is identical to #171's tree.

### Why

Two problems with profiler artifacts on a multi-node run.

1. Each node wrote its `torch_profiler/` tree to its own filesystem, so the host parser only ever saw the head node's ranks.
2. `run()` returned as soon as any node failed, **before collecting anything**. One flaky node in a multi-hour run discarded every surviving node's traces, forcing a full rerun or a manual `TraceLensParser` salvage.

### What changed

- `_collect_multi_node_traces()` — consolidates every node's trees into `<aorta_path>/combined_traces/node_<rank>/`: local copy when the orchestrator shares the head's filesystem, rsync over SSH otherwise, `scp -r` where rsync is absent. Per-node failures are logged and skipped rather than aborting the collection.
- Trace discovery skips anything under `combined_traces/` so the stale per-node originals cannot shadow the consolidated set, and seeds `trace_mtime` from the collected tree — without that seed the existing freshest-trace comparison raises `UnboundLocalError` once `trace_dir` can be pre-set.
- Collection and artifact discovery now run unconditionally. The run is still reported `FAILED`/`TIMEOUT` with the offending nodes named in `error_message`, but artifacts come from whatever the survivors produced.

### Test

`ruff` clean. Unit tests 631 → 638 (7 new: `combined_traces` path predicate, local copy-tree with recursion guard, collection layout, and partial-failure-still-collects).

Live validation from #171 still applies unchanged — this stack's tip is byte-identical to that branch for every non-test file: `cvs run test_aorta` on a real 2-node cluster (g17u19 + f16u13, 16×MI300X), 5/5 pytest cases in 148s, traces from both nodes, host parser produced metrics for all 16 ranks.

### DIFF
--- a/cvs/runners/aorta.py
+++ b/cvs/runners/aorta.py
@@ -39,6 +39,19 @@ from cvs.runners._base_runner import BaseRunner, RunConfig, RunResult, RunStatus
 log = logging.getLogger(__name__)
 
 
+def combined_traces_in(path: Path, root: Path) -> bool:
+    """Return True if ``path`` lives under ``root/combined_traces``.
+
+    Used to skip already-collected traces when rescanning the head node so
+    repeated runs do not nest combined_traces inside itself.
+    """
+    try:
+        rel = path.relative_to(root)
+    except ValueError:
+        return False
+    return rel.parts and rel.parts[0] == "combined_traces"
+
+
 @dataclass
 class RcclConfig:
     """RCCL build and runtime configuration."""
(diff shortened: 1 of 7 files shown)


cvs/runners/aorta.py
```diff
@@ -952,52 +1119,76 @@ def run(self, **kwargs) -> RunResult:
...
-                        trace_dir = candidate
-                        output_dir = candidate.parent
-                        trace_mtime = candidate_mtime
-                    elif candidate_mtime > trace_mtime:
-                        trace_dir = candidate
-                        output_dir = candidate.parent
-                        trace_mtime = candidate_mtime
+                if not candidate.is_dir():
+                    continue
+                if combined_traces_in(candidate, combined_root):
```
amd-droy: combined_traces_in is true when path is under root/combined_traces, but combined_root is already aorta_path/combined_traces, so rel.parts[0] is node_0 and the skip never runs. Pass self.config.aorta_path (as the unit tests do), or a fresher head-node torch_profiler/ can overwrite artifacts["torch_traces"] and drop every other node.

cvs/runners/aorta.py
```diff
@@ -1055,13 +1246,14 @@ def run(self, **kwargs) -> RunResult:
                     break
 
             return RunResult(
-                status=RunStatus.COMPLETED,
+                status=partial_failure_status or RunStatus.COMPLETED,
```
amd-droy, line 1381: Collecting traces after a node failure is right, but TraceLensParser.parse / AortaReportParser.parse still return immediately when not run_result.succeeded, and test_aorta still requires COMPLETED. The files land on disk; automated salvage still does not run.

amd-droy (approved):
lgtm.

PR #350, opened 2026-08-18
Added nodesmoke tier 3 support and fixed few issues inpreflight confi…

urtiwari:
Added Tier 3 preflight info (tier3_info.py, test_tier3_info.py, test_tier3_info in preflight_checks.py): runs primus-cli direct -- preflight --host --gpu --network across nodes via parallel SSH (no Slurm), mirroring the node_smoke pattern.

Config & schema: new tier3_info block in preflight_config.json (opt-in, default skip), PreflightTier3InfoConfig in schemas.py, Primus paths fall back to node_smoke, and gpus_per_node set to 8 for node health.

Reporting & setup: report.py adds Tier 3 summary/HTML; primus_setup.py supports tier3_info config section; default artifacts root moved to /home/{user-id}/preflight in node_smoke.py, rdma_connectivity.py, and docs.

Scope: Tier 3 runs Host/GPU/Network info checks only — not --perf-test (no intra/inter/ring bandwidth tests); unit tests added for flag building, output parsing, and parallel SSH launch.

Fixed the config file which was having transferbench and l2 ping default enabled

### DIFF
--- a/cvs/input/config_file/preflight/preflight_config.json
+++ b/cvs/input/config_file/preflight/preflight_config.json
@@ -11,7 +11,7 @@
       "_comment_gpus_per_node": "Exact number of AMD GPUs expected on every node.",
 
       "_example_expected_rocm_version": "6.2.0",
-      "expected_rocm_version": "<changemea>",
+      "expected_rocm_version": "10.0.0",
       "_comment_expected_rocm_version": "Expected ROCm version across all cluster nodes. Must match the output of 'amd-smi version' on all nodes. Format: 'major.minor.patch' (e.g., '6.2.0', '5.7.1')."
     },
 
@@ -168,7 +168,7 @@
       "_comment_dump_path": "Directory for per-node smoke/*.json output. Leave empty to use <reporting.artifacts_root_dir>/node_smoke.",
 
       "expected_rdma_nics": null,
-      "_comment_expected_rdma_nics": "Hard-fail when training RDMA NIC count differs. null defaults to len(node_check.rdma_interfaces). Example: 8.",
+      "_comment_expected_rdma_nics": "Hard-fail when training RDMA NIC count differs. null defaults to len(connectivity_check.rdma.interfaces). Example: 8.",
 
       "ulimit_l_min_gb": 32,
       "_comment_ulimit_l_min_gb": "FAIL when RLIMIT_MEMLOCK is below this many GiB. 0 disables.",
@@ -195,10 +195,10 @@
       "_comment_gloo_socket_ifname": "Optional GLOO_SOCKET_IFNAME override (defaults to nccl_socket_ifname when empty).",
 
       "nccl_ib_hca": "",
-      "_comment_nccl_ib_hca": "Optional NCCL_IB_HCA override. Defaults to comma-joined node_check.rdma_interfaces.",
+      "_comment_nccl_ib_hca": "Optional NCCL_IB_HCA override. Defaults to comma-joined connectivity_check.rdma.interfaces.",
 
       "nccl_ib_gid_index": null,
-      "_comment_nccl_ib_gid_index": "Optional NCCL_IB_GID_INDEX override. Defaults to node_check.gid_index.",
+      "_comment_nccl_ib_gid_index": "Optional NCCL_IB_GID_INDEX override. Defaults to connectivity_check.rdma.gid_index.",
 
       "ssh_timeout": 300,
       "_comment_ssh_timeout": "SSH timeout in seconds for each node's node_smoke invocation (~30s Tier 1; use 600+ with tier2_perf).",
@@ -229,7 +229,7 @@
       "_comment": "Primus Tier 3 Host/GPU/Network info via primus-cli direct -- preflight --host --gpu --network (opt-in; default skip). No Slurm.",
 
       "connectivity_mode": "skip",
-      "_comment_connectivity_mode": "Options: 'run' (cluster-wide preflight --host --gpu --network) or 'skip' (default).",
+      "_comment_connectivity_mode": "Options: 'run' or 'skip' (default). Independent of node_smoke.connectivity_mode — enabling node_smoke does not enable Tier 3.",
 
       "auto_setup": true,
       "_comment_auto_setup": "When true, clone/update Primus and create venv before Tier 3 (falls back to node_smoke paths).",
@@ -265,10 +265,10 @@
       "_comment_gloo_socket_ifname": "Optional GLOO_SOCKET_IFNAME override.",
 
       "nccl_ib_hca": "",
-      "_comment_nccl_ib_hca": "Optional NCCL_IB_HCA override.",
+      "_comment_nccl_ib_hca": "Optional NCCL_IB_HCA override. Defaults to comma-joined connectivity_check.rdma.interfaces when empty.",
 
       "nccl_ib_gid_index": null,
-      "_comment_nccl_ib_gid_index": "Optional NCCL_IB_GID_INDEX override.",
+      "_comment_nccl_ib_gid_index": "Optional NCCL_IB_GID_INDEX override. Defaults to connectivity_check.rdma.gid_index when null.",
 
       "ssh_timeout": 600,
       "_comment_ssh_timeout": "SSH timeout in seconds for the parallel Tier 3 cluster run.",
--- a/cvs/input/config_file/preflight/preflight_config.json
+++ b/cvs/input/config_file/preflight/preflight_config.json
@@ -11,7 +11,7 @@
       "_comment_gpus_per_node": "Exact number of AMD GPUs expected on every node.",
 
       "_example_expected_rocm_version": "6.2.0",
-      "expected_rocm_version": "10.0.0",
+      "expected_rocm_version": "<changeme>",
       "_comment_expected_rocm_version": "Expected ROCm version across all cluster nodes. Must match the output of 'amd-smi version' on all nodes. Format: 'major.minor.patch' (e.g., '6.2.0', '5.7.1')."
     },
 
--- a/cvs/input/config_file/preflight/preflight_config.json
+++ b/cvs/input/config_file/preflight/preflight_config.json
@@ -6,12 +6,12 @@
       "enabled": true,
       "_comment_enabled": "Enable GPU visibility, AMDGPU/KFD, kernel-health, and ROCm validation.",
 
-      "_example_gpus_per_node": 4,
-      "gpus_per_node": 4,
+      "_example_gpus_per_node": 8,
+      "gpus_per_node": 8,
       "_comment_gpus_per_node": "Exact number of AMD GPUs expected on every node.",
 
       "_example_expected_rocm_version": "6.2.0",
-      "expected_rocm_version": "<changeme>",
+      "expected_rocm_version": "<changemea>",
       "_comment_expected_rocm_version": "Expected ROCm version across all cluster nodes. Must match the output of 'amd-smi version' on all nodes. Format: 'major.minor.patch' (e.g., '6.2.0', '5.7.1')."
     },
 
@@ -25,11 +25,11 @@
         "_comment_connectivity_mode": "RDMA connectivity testing mode. Options: 'basic' (test adjacent node pairs, fast, ~14% coverage for 8 nodes), 'full_mesh' (test all possible node pairs, comprehensive, 100% coverage), 'skip' (skip RDMA connectivity testing entirely).",
 
         "_example_gid_index": "3",
-        "gid_index": "3",
+        "gid_index": "<changeme>",
         "_comment_gid_index": "GID index to check on all RDMA interfaces. Typically '3' for RoCE (RDMA over Converged Ethernet). Must be a valid GID index for your InfiniBand/RoCE setup.",
 
         "_example_interfaces": ["rocep28s0", "rocep62s0", "rocep79s0", "rocep96s0", "rocep158s0", "rocep190s0", "rocep206s0", "rocep222s0"],
-        "interfaces": ["enp4s0np0"],
+        "interfaces": ["<changeme>"],
         "_comment_interfaces": "List of specific RDMA interface names that should be present on all cluster nodes. Examples: ['rocep28s0', 'rocep62s0'] for standard setup, ['mlx5_0', 'mlx5_1'] for Mellanox, ['ib0', 'ib1'] for generic InfiniBand.",
 
         "nodes_per_full_mesh_group": 32,
@@ -82,7 +82,7 @@
         "l2ping": {
           "_comment": "Strict afmctl L2 connectivity gate executed before TransferBench and RDMA",
 
-          "enabled": true,
+          "enabled": false,
           "_comment_enabled": "Enable IFoE L2 connectivity validation.",
 
           "pings_per_port": 3,
@@ -92,7 +92,7 @@
         "transferbench": {
           "_comment": "TransferBench IFoE data-path validation executed before RDMA",
 
-          "enabled": true,
+          "enabled": false,
           "_comment_enabled": "Enable the TransferBench preflight gate.",
 
           "scope": "node",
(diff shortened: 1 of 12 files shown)


solaiys (commented):
## Review: 4 blocking issues

Reviewed the Tier 3 addition plus the config changes. CI is green, so everything below is logic/config behaviour that the current tests don't cover. I verified each item by checking out the PR head and executing the affected code paths, not just reading the diff.

| # | Blocking issue | Where |
|---|---|---|
| 1 | `<changemea>` typo defeats the `<changeme>` placeholder guard, so ROCm validation runs against a bogus expected version | `preflight_config.json:14` |
| 2 | Tier 3 inherits **every** setting from `node_smoke`, so it silently auto-runs cluster-wide on existing configs | `tier3_info.py:43` |
| 3 | `NCCL_IB_HCA` / `NCCL_IB_GID_INDEX` are derived from `node_check`, which can never hold those keys | `tier3_info.py:240-252` |
| 4 | `artifacts_root_dir` in-code default is now an unresolvable literal `/home/{user-id}/preflight`, which crashes report generation | `report.py:206` |

Issue 2 is the one I'd most want fixed before merge: it changes behaviour for configs that predate this PR, and the effect is an unrequested distributed launch across the whole cluster.

For context on what I did *not* flag: the `l2ping`/`transferbench` `enabled: false` flips, `gpus_per_node` 4 -> 8, and turning `gid_index`/`interfaces` back into `<changeme>` all match the stated intent of the PR and the repo's template-config convention, so I've left them alone.

cvs/input/config_file/preflight/preflight_config.json
```diff
@@ -6,12 +6,12 @@
...
 
-      "_example_gpus_per_node": 4,
-      "gpus_per_node": 4,
+      "_example_gpus_per_node": 8,
+      "gpus_per_node": 8,
       "_comment_gpus_per_node": "Exact number of AMD GPUs expected on every node.",
 
       "_example_expected_rocm_version": "6.2.0",
-      "expected_rocm_version": "<changeme>",
+      "expected_rocm_version": "<changemea>",
```
solaiys: **Blocking — typo silently disables the placeholder guard.**

`<changeme>` became `<changemea>` here. That looks like a stray keystroke, but it has real consequences, because the guard in `utils_lib._resolve_placeholders_in_dict` is an exact substring test:

```python
has_changeme = '<changeme>' in value.lower() or '<changeme>' in path.lower()
```

`'<changeme>' in '<changemea>'` is `False` (the `>` no longer follows `changeme`), so the placeholder sails through config load instead of triggering the intended `sys.exit(1)`. It is then passed straight into the ROCm check:

```python
expected_version = get_nested_config(config_dict, 'node_check', 'expected_rocm_version', '6.2.0')
version_checker = RocmVersionCheck(phdl, expected_version, config_dict)
```

Since `node_check.enabled` is `true` by default, every node's `amd-smi` version gets compared against the literal string `<changemea>` and mismatches, so `rocm_versions` reports FAIL cluster-wide and drags `overall_status` to FAIL. Confirmed on this branch:

```
shipped value               : '<changemea>'
caught by '<changeme>' guard: False
```

**Fix:** restore `"expected_rocm_version": "<changeme>"`.

cvs/lib/preflight/tier3_info.py
```diff
@@ -0,0 +1,404 @@
...
+    re.IGNORECASE,
+)
+
+
+def resolve_tier3_setting(cfg: dict, key: str, default=None):
+    """Read ``tier3_info.<key>``, falling back to ``node_smoke.<key>``."""
+    value = get_nested_config(cfg, "tier3_info", key, None)
+    if value not in (None, ""):
+        return value
+    return get_nested_config(cfg, "node_smoke", key, default)
```
solaiys: **Blocking — the `node_smoke` fallback is applied to every key, so Tier 3 auto-enables itself on existing configs.**

The PR description says "Primus paths fall back to node_smoke", but `resolve_tier3_setting` falls back for *all* keys, including `connectivity_mode`. Any config that predates this PR has no `tier3_info` block at all, so `get_nested_config(cfg, "tier3_info", "connectivity_mode", None)` returns `None` and the mode is inherited from `node_smoke`. For the documented `node_smoke` usage (`connectivity_mode: "run"`), Tier 3 turns itself on.

Ran this against the PR head with a config containing only a `node_smoke` block (`connectivity_mode: "run"`, `dump_path: "/shared/node_smoke_out"`, `extra_args: ["--expected-gpus", "8"]`):

```
tier3_info.connectivity_mode in config : ABSENT
resolved self.mode                     : 'run'
resolved dump_path                     : '/shared/node_smoke_out'
resolved extra_args                    : ['--expected-gpus', '8']
exec_cmd_list called?                  : True
emitted flags: --host --gpu --network --dump-path /shared/node_smoke_out \
               --report-file-name tier3_info --dist-timeout-sec 120 --disable-pdf --expected-gpus 8
```

Three distinct consequences, none of them opt-in:

1. A torchrun-based distributed run is launched across the whole cluster even though the feature is documented as opt-in and both the JSON and `PreflightTier3InfoConfig` default to `skip`. `auto_setup` is inherited the same way, so it also re-runs the Primus clone/pip install pass.
2. `dump_path` is inherited, so Tier 3 writes into the `node_smoke` output directory rather than the documented `<artifacts_root_dir>/tier3_info`.
3. `extra_args` is inherited, so `node_smoke`-specific flags such as `--expected-gpus 8` get appended to a `preflight` command line that does not accept them.

Note this is a real divergence from `node_smoke`, which reads each of its own keys with `get_nested_config(cfg, "node_smoke", ...)` and never inherits.

**Fix:** restrict the fallback to the Primus location/setup keys it was meant for (`primus_dir`, `venv_activate`, `primus_git_*`, `force_reclone`, `pip_install_mode`, `torch_pip_index_url`, `setup_timeout`, `shared_install`) and read the Tier 3-owned keys (`connectivity_mode`, `auto_setup`, `dump_path`, `extra_args`, `report_file_name`, `master_port`, `gpus_per_node`, `ssh_timeout`, `dist_timeout_sec`, `save_pdf`) strictly from `tier3_info`. A key allowlist passed to `resolve_tier3_setting` would be enough.

cvs/lib/preflight/tier3_info.py
```diff
@@ -0,0 +1,404 @@
...
+
+        nccl_ib_hca = resolve_tier3_setting(cfg, "nccl_ib_hca", None)
+        if not nccl_ib_hca and rdma_ifaces:
+            nccl_ib_hca = ",".join(rdma_ifaces)
+        self.nccl_ib_hca = nccl_ib_hca or None
+
+        gid_index = resolve_tier3_setting(cfg, "nccl_ib_gid_index", None)
+        if gid_index is None:
+            gid_index = node_check.get("gid_index")
+        self.nccl_ib_gid_index = int(gid_index) if gid_index not in (None, "") else None
```
solaiys: **Blocking — these two settings are read from a section that cannot contain them, so the derivation is dead code.**

`rdma_interfaces` and `gid_index` are read off `cfg["node_check"]`, but they live under `connectivity_check.rdma` in the shipped config, and `node_check` is validated as a closed set:

```python
unknown = sorted(
    key for key in set(config) - {'enabled', 'gpus_per_node', 'expected_rocm_version'} if not key.startswith('_')
)
if unknown:
    raise ValueError("Unsupported preflight.node_check option(s): " + ', '.join(unknown))
```

So `node_check` can never legally hold either key — putting them there is a hard config error. `rdma_ifaces` is therefore always `[]` and `gid_index` always `None`, and both auto-derivations are unreachable. Verified on this branch with `connectivity_check.rdma` populated:

```
connectivity_check.rdma.interfaces : ['rocep28s0', 'rocep62s0']
connectivity_check.rdma.gid_index  : '3'
-> self.nccl_ib_hca       : None
-> self.nccl_ib_gid_index : None
```

Net effect: Tier 3's `--network` checks always run without `NCCL_IB_HCA` / `NCCL_IB_GID_INDEX` pinned unless the operator duplicates the values into `tier3_info`, which is exactly what the `_comment_nccl_ib_hca` "optional override" wording tells them they don't have to do. On multi-NIC hosts that means NCCL picks its own HCA, so the check validates a different path than the rest of preflight.

Heads up that this is copied from `node_smoke.py:265,301`, which has the identical bug, so it's pre-existing rather than introduced here — but it's still wrong in the new code, and fixing both would be ideal.

**Fix:** read from the real location, e.g. `get_nested_config(cfg, "connectivity_check.rdma", "interfaces", [])` and `... "gid_index", None)`. Please also guard the `int()` on line 252: `gid_index` is a *string* in the config, and this PR sets it to `"<changeme>"`, so once the section is corrected a non-numeric value would raise `ValueError` inside `__init__`. Wrap it in `try/except (TypeError, ValueError)` and fall back to `None`.

cvs/lib/preflight/report.py
```diff
@@ -191,7 +202,9 @@ def _generate_html_report(self, output_path=None):
 
         if output_path is None:
             timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
-            output_dir = get_nested_config(self.config_dict, 'reporting', 'artifacts_root_dir', '/tmp/preflight')
+            output_dir = get_nested_config(
+                self.config_dict, 'reporting', 'artifacts_root_dir', '/home/{user-id}/preflight'
```
solaiys: **Blocking — the new in-code default is an unresolvable placeholder and crashes report generation.**

`{user-id}` is only substituted for values that actually appear in the config file: the `config_dict` fixture calls `resolve_test_config_placeholders(config_dict, cluster_dict)` on the loaded JSON. An in-code default like this one is never seen by that pass, so when a config omits `reporting.artifacts_root_dir` the literal string survives and reaches the very next line, `Path(output_dir).mkdir(parents=True, exist_ok=True)`.

On a normal host `/home` is root-owned (`drwxr-xr-x`, uid 0), so creating `/home/{user-id}` raises `PermissionError` and HTML report generation dies. The old `/tmp/preflight` default was always writable, so this is a new crash on the fallback path. Verified on this branch:

```
node_smoke dump w/o reporting : '/home/{user-id}/preflight/node_smoke'
tier3      dump w/o reporting : '/home/{user-id}/preflight/tier3_info'
/home writable by me          : False   -> mkdir raises PermissionError
```

Same change needs attention in three more places, one of which builds a remote path that would be `mkdir`'d on every cluster node:

- `cvs/lib/preflight/node_smoke.py:61`
- `cvs/lib/preflight/rdma_connectivity.py:106` (feeds `<artifacts_root_dir>/rdma_connectivity_workspace` on each node)
- `cvs/parsers/schemas.py:1331` (`PreflightReportingConfig.artifacts_root_dir` default is now a placeholder too, so validation hands callers a path that cannot be used as-is)

**Fix:** keep the in-code fallbacks literal and writable (`/tmp/preflight`), or resolve the placeholder at the point of use with `getpass.getuser()`. Changing the *config file* value to `/home/{user-id}/preflight` is fine — it's only the code-level defaults and the schema default that can't carry an unresolved placeholder.

cvs/input/config_file/preflight/preflight_config.json
solaiys: **Still blocking — `<changemea>` was not reverted.**

`2b9559de` did not touch this file. Shipped value is still `'<changemea>'`, so `utils_lib._resolve_placeholders_in_dict` still does not trip (`'<changeme>' in '<changemea>'` is `False`) and `RocmVersionCheck` still compares every node against the literal string. Restore `"expected_rocm_version": "<changeme>"`.

cvs/lib/preflight/rdma_connectivity.py
```diff
@@ -10,7 +10,8 @@
 from cvs.lib import globals
 from collections import defaultdict
 
-from cvs.lib.preflight.base import PreflightCheck, partition_nodes_into_groups
+from cvs.lib.preflight.base import PreflightCheck
+from cvs.lib.preflight.node_smoke import DEFAULT_ARTIFACTS_ROOT_DIR
```
solaiys, line 14: **Blocking — this import edit dropped `partition_nodes_into_groups` and broke CI.**

When adding `DEFAULT_ARTIFACTS_ROOT_DIR`, the previous line

```python
from cvs.lib.preflight.base import PreflightCheck, partition_nodes_into_groups
```

was replaced with `PreflightCheck` only. `_execute_full_mesh` still calls `partition_nodes_into_groups(...)` at line 399, so CI `test` fails with three `NameError`s (`test_execute_full_mesh_*`).

**Fix:** restore the name on the `base` import:

```python
from cvs.lib.preflight.base import PreflightCheck, partition_nodes_into_groups
from cvs.lib.preflight.node_smoke import DEFAULT_ARTIFACTS_ROOT_DIR
```

cvs/lib/preflight/node_smoke.py
```diff
@@ -298,7 +331,7 @@ def _load_settings(self):
 
         gid_index = get_nested_config(cfg, "node_smoke", "nccl_ib_gid_index", None)
         if gid_index is None:
-            gid_index = node_check.get("gid_index")
+            gid_index = resolve_rdma_gid_index(cfg)
         self.nccl_ib_gid_index = int(gid_index) if gid_index not in (None, "") else None
```
solaiys, line 335: **Blocking — the section fix is right, but `int()` now explodes on the shipped config.**

`resolve_rdma_gid_index()` correctly reads `connectivity_check.rdma.gid_index`. This PR also changed that field to `"<changeme>"`. `_load_settings()` always runs in `__init__` (including when `connectivity_mode` is `skip`), so:

```python
int("<changeme>")  # ValueError
```

hits as soon as `NodeSmokeCheck` (and the same pattern at `tier3_info.py:263`) is constructed. Previously this `int()` was dead code because `node_check` never held `gid_index`; wiring it to the real key without a guard is a new crash on the default preflight config.

Same for `interfaces: ["<changeme>"]` → `nccl_ib_hca = "<changeme>"` if a check actually runs.

**Fix:** treat non-numeric / placeholder values as unset:

```python
try:
    self.nccl_ib_gid_index = int(gid_index) if gid_index not in (None, "") else None
except (TypeError, ValueError):
    self.nccl_ib_gid_index = None
```

Apply the same guard in `tier3_info.py`. Optionally skip interface names that contain `<changeme>` in `resolve_rdma_interfaces`.

solaiys (commented):
## Re-review of `2b9559de` — cannot approve yet

Thanks for addressing the earlier comments. Two of the four original blockers are fixed:

- Tier 3 no longer inherits `connectivity_mode` / `dump_path` / `extra_args` from `node_smoke` (fallback is now only `primus_dir` / `venv_activate`). Covered by `test_tier3_skipped_when_only_node_smoke_enabled`.
- In-code / schema `artifacts_root_dir` fallback is writable `/tmp/preflight` again.

Still blocking merge:

| # | Issue |
|---|---|
| 1 | `<changemea>` is still in `preflight_config.json` — the placeholder guard is still bypassed |
| 2 | `rdma_connectivity.py` dropped `partition_nodes_into_groups` — CI `test` is red (`NameError`, 3 errors) |
| 3 | Reading `connectivity_check.rdma.gid_index` is correct, but `int("<changeme>")` now runs on the **shipped** default config and will crash `NodeSmokeCheck` / `Tier3InfoCheck` construction even when those checks are skipped |

Please restore the import, restore `"<changeme>"`, and wrap the `int()` (treat non-numeric as `None`). Happy to approve once CI is green and those three are in.

cvs/lib/preflight/node_smoke.py
solaiys: **Retracting this one — my analysis was wrong, please disregard.**

I claimed `int("<changeme>")` is reachable on the shipped config. It is not. The placeholder guard runs during config load, before any check object is constructed:

```python
# cvs/tests/preflight/preflight_checks.py:289
config_dict = resolve_test_config_placeholders(config_dict, cluster_dict)
```

Since `connectivity_check.rdma.gid_index` ships as the correct `<changeme>` sentinel, that call exits first. Verified by running the real function:

```
gid_index='<changeme>'  -> SystemExit(1), "ACTION REQUIRED: replace all the '<changeme>' placeholders"
```

So `NodeSmokeCheck.__init__` is never reached with a placeholder value, and `<changeme>` here is working exactly as designed. The `int()` conversion is only reachable for a non-numeric, non-placeholder value such as `"auto"` — a robustness nit at most, not blocking. Sorry for the noise.

The `partition_nodes_into_groups` import issue is unaffected and still stands.

solaiys (approved):
**Approving — all previously flagged blockers are resolved.** Verified against head `ccf59078`.

Resolution status of my earlier comments:

| Finding | Status |
| --- | --- |
| `tier3_info`: `node_smoke` fallback applied to every key, auto-enabling Tier 3 | Fixed in `2b9559de` — fallback narrowed to `_TIER3_NODE_SMOKE_FALLBACK_KEYS = {primus_dir, venv_activate}`, covered by `test_tier3_skipped_when_only_node_smoke_enabled` |
| RDMA settings read from `node_check` (a closed key set that cannot hold them) | Fixed in `2b9559de` — `resolve_rdma_interfaces()` / `resolve_rdma_gid_index()` now read `connectivity_check.rdma.*`, with legacy normalization for back-compat |
| `/home/{user-id}/preflight` in-code default not writable and never placeholder-resolved | Fixed in `2b9559de` — `DEFAULT_ARTIFACTS_ROOT_DIR = "/tmp/preflight"` shared across `node_smoke`, `rdma_connectivity`, `report`, `tier3_info` |
| Dropped `partition_nodes_into_groups` import breaking CI | Fixed in `ccf59078` |
| `int(gid_index)` on a placeholder value | Retracted — the placeholder guard in `preflight_checks.py:289` aborts before any check is constructed, so this was never reachable |

Also confirmed the `_comment_*` strings in `preflight_config.json` were updated to point at `connectivity_check.rdma.interfaces` / `.gid_index` instead of the stale `node_check.rdma_interfaces`, so the documentation now matches where the code actually reads from. The `connectivity_mode` comment calling out that Tier 3 is independent of `node_smoke.connectivity_mode` is a good clarification of the fallback change.

CI is green on `ccf59078`: both `test` and `container-smoke` pass.

solaiys:
LGTM..

PR #351, opened 2026-08-18
Adding subtests for performance sweeps and collapsible metrics view

amd-droy:
**Updates:**

**SGLang core — shared logic & lib refactor**
cvs/lib/inference/sglang/sglang_common.py — shared helpers for benchmark log parsing, polling, threshold verification, subtest verification, and OpenAI/lm-eval probes
cvs/lib/inference/sglang/sglang_single_lib.py — delegates parsing/verification to sglang_common; adds verify_inference_results_subtests()
cvs/lib/inference/sglang/sglang_distributed_lib.py — same refactor as single
cvs/lib/inference/sglang/sglang_disagg_lib.py — same refactor; removes duplicated parsing/verify code
**SGLang tests**
cvs/tests/inference/sglang/sglang_single.py — perf benchmark uses pytest subtests; records metric rows for HTML report; adds test_verify_dmesg_after_benchmark
cvs/tests/inference/sglang/sglang_distributed.py — same subtest + dmesg changes
cvs/tests/inference/sglang/sglang_disagg_distributed.py — same subtest + dmesg changes
cvs/tests/inference/sglang/conftest.py — lifecycle perf_metric_rows; pytest-html hooks for collapsible per-metric pass/fail rows
cvs/tests/inference/sglang/_shared.py — updated test ordering to include dmesg stage
**pytest-html reporting**
cvs/lib/report/benchmark_metric_registry.py (new) — stores per-metric pass/fail rows; patches pytest-html JSON after session
cvs/lib/report/render/perf_metric_table.py (new) — renders metric table HTML for pytest-html rows
cvs/lib/report_plugins.py — CSS/JS for expand/collapse benchmark metric rows; subtest count in filter bar
cvs/lib/report/presets/sglang_single.py — clears row_card_test_names for perf test (metrics shown via subtest HTML panel)
cvs/lib/report/presets/sglang_distributed.py — same preset change
cvs/lib/report/presets/sglang_disagg_distributed.py — same preset change
**Sample configs**
cvs/input/config_file/inference/sglang/mi30x_sglang_llama_70b_single.json (new) — single-node MI30x Llama 3.1 70B template
cvs/input/config_file/inference/sglang/mi30x_sglang_llama_70b_distributed.json (new) — multi-node unified-server template
cvs/input/config_file/inference/sglang/mi30x_sglang_llama_70b_disaggregated.json (new) — prefill/decode disaggregated template
**Unit tests & dependency**
cvs/lib/inference/sglang/unittests/test_sglang_common.py (new) — parsing, polling, threshold/subtest verification
cvs/lib/report/unittests/test_benchmark_metric_registry.py (new) — metric row storage and HTML patching
cvs/lib/report/unittests/test_perf_metric_table.py (new) — metric table rendering
requirements.txt — adds pytest-subtests


cvs/lib/inference/sglang/unittests/test_sglang_common.py
```diff
@@ -0,0 +1,155 @@
...
+"""
+
+
+class _FakeSubtests:
+    class _Ctx:
+        def __enter__(self):
+            return self
+
+        def __exit__(self, exc_type, exc, tb):
+            return False
```
solaiys, line 49: **Blocking (half 1 of 2).** Returning `False` here is the opposite of what real pytest-subtests does: its context manager *captures* the assertion, records a subtest failure, and returns `True` so the caller's loop continues to the next metric.

Because this stub propagates instead, the fail path can't be expressed with it — a test with a violated threshold would raise out of `verify_inference_results_subtests` at the first bad metric, so you'd never reach the `return all_passed` or see the remaining rows. That makes the current suite structurally unable to catch a gating regression.

Suggest mirroring the plugin:

```python
class _FakeSubtests:
    def __init__(self):
        self.failures = []

    class _Ctx:
        def __init__(self, outer):
            self._outer = outer

        def __enter__(self):
            return self

        def __exit__(self, exc_type, exc, tb):
            if exc_type is not None:
                self._outer.failures.append(exc)
                return True  # suppress, like pytest-subtests
            return False

    def test(self, **kwargs):
        return self._Ctx(self)
```

That also gives you `failures` to assert one subtest failure was raised per violated metric.

cvs/lib/inference/sglang/unittests/test_sglang_common.py
```diff
@@ -0,0 +1,155 @@
...
+        host_exec = mock.Mock(return_value={'head': 'time'})
+        with mock.patch.object(sglang_common.time, 'sleep'):
+            end = sglang_common.verify_inference_results(
+                {'node1': {'mean_ttft_ms': '50.0'}},
+                {'mean_ttft_ms': 100.0},
+                host_exec,
+            )
+        self.assertEqual(end, {'head': 'time'})
+
+    def test_verify_inference_results_subtests(self):
```
solaiys, line 130: **Blocking (half 2 of 2).** This only covers the passing case — actual 50 against a max of 100, asserting `passed` is True. There's no test anywhere that a *violated* threshold is gated, which is the invariant the whole PR rests on.

I confirmed the behaviour is correct today (pytest exits 1, parent test reports `contains 1 failed subtest`, `all_passed` is `False`, and both metrics still get evaluated), so this is purely a coverage gap. But since enforcement now depends on a third-party plugin's exception-suppression semantics and `pytest-subtests` is unpinned in `requirements.txt`, a version bump could silently turn threshold violations green with nothing to catch it.

Once the stub above suppresses correctly, a companion test would lock it in:

```python
def test_verify_inference_results_subtests_gates_violation(self):
    host_exec = mock.Mock(return_value={'head': 'time'})
    lifecycle = mock.Mock()
    lifecycle.perf_metric_rows = {}
    subtests = _FakeSubtests()
    with mock.patch.object(sglang_common.time, 'sleep'):
        passed, _end = sglang_common.verify_inference_results_subtests(
            {'node1': {'mean_ttft_ms': '500.0', 'p99_ttft_ms': '10.0'}},
            {'mean_ttft_ms': 100.0, 'p99_ttft_ms': 100.0},
            host_exec, subtests, 'bench_serv',
            lifecycle=lifecycle, report_nodeid='nid',
        )
    self.assertFalse(passed)
    rows = {r['metric']: r['status'] for r in lifecycle.perf_metric_rows['nid']}
    self.assertEqual(rows, {'mean_ttft_ms': 'fail', 'p99_ttft_ms': 'pass'})
    self.assertEqual(len(subtests.failures), 1)
```

The last two assertions are the important ones: they prove evaluation continues past the first violation and that exactly one subtest failure was raised.

cvs/lib/inference/sglang/sglang_common.py
```diff
@@ -175,6 +180,372 @@ def coerce_sglang_actual(value: Any) -> float | None:
...
+    (r'Benchmark duration\s+\(s\):\s+([0-9]+)', 'benchmark_duration'),
+    (r'Total input tokens:\s+([0-9\.]+)', 'total_input_tokens'),
+    (r'Total generated tokens:\s+([0-9\.]+)', 'total_generated_tokens'),
+    (r'Request throughput \(req/s\):\s+([0-9\.]+)', 'request_throughput_per_sec'),
+    (r'Output token throughput \(tok/s\):\s+([0-9\.]+)', 'output_throughput_per_sec'),
+    (r'Mean TTFT \(ms\):\s+([0-9\.]+)', 'mean_ttft_ms'),
+    (r'Median TTFT \(ms\):\s+([0-9\.]+)', 'median_ttft_ms'),
+    (r'P99 TTFT \(ms\):\s+([0-9\.]+)', 'p99_ttft_ms'),
+    (r'Mean TPOT \(ms\):\s+([0-9\.]+)', 'mean_tpot_ms'),
+    (r'Median TPOT \(ms\):\s+([0-9]+)', 'median_tpot_ms'),
```
solaiys: **Not blocking — pre-existing, but this is the moment to fix it.** This pattern and `Benchmark duration` on line 292 capture `([0-9]+)` while every sibling uses `([0-9\.]+)`, so a float is truncated at the decimal point: a `median_tpot_ms` of `10.9` parses as `10` and would *pass* a `max: 10` threshold.

I checked git history — both are identical on `main` in all three libs and were moved here verbatim, so this PR doesn't introduce it and I'm not treating it as a blocker. Worth folding in while the code is being consolidated into one place, though, since it directly affects gating precision.

One thing that hides it: the new fixture uses integer values (`Median TPOT (ms): 12`, `Benchmark duration (s): 120`), so `test_parse_inference_bench_results_basic` passes either way. A fractional value in the sample log would catch it.

cvs/tests/inference/sglang/sglang_single.py
```diff
@@ -144,12 +134,20 @@ def test_run_performance_benchmark_test(im_obj, inf_res_dict, lifecycle, request
...
     )
     inf_res_dict[key] = dict(im_obj.inference_results_dict or {})
     lifecycle.complete_stage(request, f"bench_serv_random[{perf_cell['isl']}/{perf_cell['osl']}]", t0)
 
 
+def test_verify_dmesg_after_benchmark(im_obj, lifecycle, request):
+    globals.error_list = []
+    t0 = time.monotonic()
+    time.sleep(2)
+    verify_dmesg_for_errors(im_obj.orch.all, im_obj.inference_start_time, im_obj.inference_end_time)
```
solaiys, line 151: **Not blocking**, but worth knowing about this new call. `inference_end_time` initialises to `None` and is only assigned inside the verify path, so if the perf test raises before verification is reached (or is deselected), this passes `None` through as the end of the dmesg window.

The default node-scraper path is safe — `till_end_flag` defaults to `True`, so `end_time_dict` is never dereferenced. But the legacy path reads `end_time_dict[node0]` unconditionally, regardless of `till_end_flag`, so under `CVS_DMESG_PARSER=legacy` this surfaces as `TypeError: 'NoneType' object is not subscriptable` instead of a clean failure.

Cheapest guard is to skip or fall back when the timestamp is missing, e.g. `if not im_obj.inference_end_time: pytest.skip("benchmark did not complete; no dmesg window")`, or pass `till_end_flag=True` explicitly and let the end bound go unused.

solaiys (commented):
## Review: one blocking issue

Reviewed against `main` (noting `dev/dtni` is retired). The refactor itself holds up well — my main worry going in was that collapsing three libs into `sglang_common` would quietly drop metrics, and it doesn't. The one blocker is a test-coverage gap on the new gating path rather than a live defect.

### Blocking: the new per-metric gating path has no failing-case test, and the test double can't express one

This PR rewires how perf thresholds fail a run. Previously `benchserv_test_random` called `verify_inference_results`, where a violation went `fail_test` -> `globals.error_list` -> `update_test_result()` -> `pytest.fail` — entirely in-repo and deterministic. Now the tests call `benchserv_test_random(verify=False)` and enforcement lives in `assert violation is None` inside `with subtests.test(...)`, which depends on pytest-subtests suppressing the assertion and emitting its own failure report.

**I verified this works today.** With one violated and one satisfied threshold, pytest exits `1`, the parent test is marked `FAILED - contains 1 failed subtest`, `all_passed` returns `False`, and both metrics are still evaluated — so the plugin does suppress the assert and the loop continues. Gating is functional on pytest-subtests 0.15.0.

The problem is that nothing tests it. `test_verify_inference_results_subtests` only exercises the passing path (`mean_ttft_ms` 50 against a max of 100, asserting `passed` is True); there is no assertion anywhere that a violated threshold yields `all_passed is False` or a `'fail'` row. And the stub cannot currently express that case — `_FakeSubtests._Ctx.__exit__` returns `False`, the opposite of the real plugin's suppress-and-record behaviour, so a fail-path test written against it would abort the function at the first violation instead of continuing through the remaining metrics.

Combined with `pytest-subtests` being unpinned, a version change could silently stop threshold violations from failing runs, and the failure mode is invisible: perf regressions would simply start reporting green. That's why I'd rather see it pinned down by a test now. Inline comments on the two exact spots.

### Checked and cleared

**The parsing refactor is faithful.** I diffed each variant's old metric set against the new flag-driven parser: single and distributed previously had no ITL and no per-GPU throughput and now pass neither flag, while disagg passes `include_itl=True, include_extended_e2e_percentiles=True`, preserving its richer p90/p95 set. `verify=True` is the keyword-only default in all three libs and all three call sites explicitly opt out, so no caller silently loses verification.

**Concurrency (mandated scan).** No threads, async, multiprocessing or background jobs are introduced, and there's no xdist in `requirements.txt` or `pytest.ini`, so the registry's module-level `_ROWS_BY_NODEID` / `_SUBTEST_SUMMARY` state is process-local with a single writer. `_SUBTEST_SUMMARY_COUNTED` guards double-counting if patching runs more than once.

Also clear:
- `pytest_sessionfinish` guards `if not htmlpath: return`, so no `Path(None)` crash when `--html` is omitted.
- The new report JS bails when `.cvs-benchmark-metrics-wrap` is absent, so this shared-file change leaves other suites' reports untouched.
- The three new configs are valid JSON with placeholders only on site-specific values, and their `threshold_file` references resolve.
- Perf cells derive from the threshold file via `perf_cells_from_thresholds`, so sweep/threshold coverage is automatic rather than something that can drift.
- `test_verify_dmesg_after_benchmark` is defined in all three suites and ordered consistently after the perf test.
- `setup.py` reads `requirements.txt` directly, so the new `pytest-subtests` dependency propagates to installs.
- Lint clean (`ruff format`, `ruff check`, pylint 10.00/10); full unit suite 1152 -> 1172 tests, OK on both the PR and the merge base.

### Two non-blocking notes

Left inline: a pre-existing float-truncation in two metric regexes (moved verbatim, so not introduced here, but this consolidation is the natural moment to fix it), and `inference_end_time` being `None`-able into the new dmesg test under the legacy parser.

cvs/lib/report/render/perf_metric_table.py
```diff
@@ -0,0 +1,63 @@
...
+from __future__ import annotations
+
+import html
+from typing import Any, Mapping, Sequence
+
+_BENCHMARK_METRICS_WRAP = 'cvs-benchmark-metrics-wrap'
+
+
+def _metric_display_label(metric_key: str) -> str:
+    from cvs.lib.inference.sglang.sglang_parsing import SGLANG_RESULTS_COLUMNS
```
solaiys: **Not blocking — a layering suggestion, entirely your call.**

This is the only module in the generic report layer that imports a single named suite. Everything else outside `presets/` is either suite-agnostic or imports the shared `cvs.lib.inference.utils.*` layer; `presets/` is where per-suite coupling is expected to live, and that's exactly what `presets/sglang_single.py` and friends do. The function-local import here reads like it was already felt to be awkward.

The coupling is only these two lines, and only to turn a metric key into a display label — the other 60 lines of this file are fully generic. So I'd suggest injecting the label source rather than relocating the file, which keeps it correctly placed in `render/` alongside `cell_card.py`, `gate_matrix.py` and `sweep_charts.py`:

```python
def _metric_display_label(metric_key: str, columns: Sequence[tuple[str, str]] = ()) -> str:
    for label, key in columns:
        if key == metric_key:
            return label
    if not metric_key:
        return ''
    return metric_key[0].upper() + metric_key[1:]


def render_benchmark_metrics_html(
    rows: Sequence[Mapping[str, Any]],
    *,
    columns: Sequence[tuple[str, str]] = (),
) -> str:
    ...
    label = html.escape(_metric_display_label(str(row.get('metric') or ''), columns))
```

Defaulting `columns` to `()` falls through to the prettify path that's already there, so behaviour is unchanged for any caller that passes nothing and the existing unit tests keep passing as-is.

Threading it is small, because both entry points already start in sglang-land:

- `conftest.py:569` calls `render_benchmark_metrics_html(rows)` directly — pass `columns=SGLANG_RESULTS_COLUMNS` there.
- `conftest.py:645` calls `patch_benchmark_metrics_into_html(...)`, which reaches the renderer via `_apply_benchmark_entry_patch` (`benchmark_metric_registry.py:144`) and `benchmark_metrics_extra` (`:89`), so those three signatures take the same keyword and hand it down.

Worth noting `benchmark_metric_registry.py` itself needs no move at all — it has zero sglang references in code (only two docstring mentions), `benchmark_test_name` is already parameterised, and the pytest-html JSON patching is the kind of thing vLLM/ATOM will want for the same threshold-per-metric shape. The one sglang-flavoured leftover there is `DEFAULT_BENCHMARK_TEST_NAME`, since `test_run_performance_benchmark_test` only exists in the three sglang suites — optional, but moving that default to the sglang call site would leave the registry free of suite assumptions.

To be explicit about scope: I'm deliberately *not* asking for these files to move to `cvs/lib/inference/sglang/`. The CSS/JS for this feature already lives in the shared `cvs/lib/report_plugins.py` and is injected into every report, so moving only the Python would split one feature across two layers. Happy for you to take the injection tweak, defer it to a follow-up, or leave it as-is if you'd rather not grow the diff — your decision, not a merge gate.

cvs/lib/inference/sglang/sglang_common.py
amd-droy: resolved.

cvs/lib/inference/sglang/unittests/test_sglang_common.py
amd-droy: Fixed _FakeSubtests to suppress and record exceptions like pytest-subtests, and added test_verify_inference_results_subtests_gates_violation asserting all_passed is False, both metrics are evaluated (fail + pass rows), and exactly one subtest failure is recorded.

cvs/lib/inference/sglang/unittests/test_sglang_common.py
amd-droy: Added test_verify_inference_results_subtests_gates_violation with a fixed _FakeSubtests stub that suppresses like pytest-subtests. Asserts all_passed is False, both metrics are evaluated (fail + pass rows), and exactly one subtest failure is recorded.

cvs/tests/inference/sglang/sglang_single.py
amd-droy: Added a guard in test_verify_dmesg_after_benchmark to pytest.skip when inference_end_time is unset (perf stage failed/skipped before verification). Avoids a legacy-parser TypeError and makes “no dmesg window” explicit.

cvs/lib/report/render/perf_metric_table.py
amd-droy: Resolved. Removed the SGLang import from the generic report layer and inject metric labels at the suite boundary instead.

**What changed -**
perf_metric_table.py — metric_display_label(..., columns=()) and render_benchmark_metrics_html(..., columns=()); no suite imports. Empty columns falls back to prettified keys.
benchmark_metric_registry.py — record_benchmark_metric_rows(..., columns=()) stores columns per nodeid; benchmark_metric_columns_for_nodeid() feeds patch/render at session finish.
SGLang — the three perf tests pass columns=SGLANG_RESULTS_COLUMNS at record time; conftest.py uses stored columns when attaching pytest-html extras and passes benchmark_test_name= to patch_benchmark_metrics_into_html.

**How other frameworks will use it**

Each suite owns its metric vocabulary; the report layer stays shared. When vLLM/ATOM (or others) adopt per-metric subtest rows, they wire up the same three touchpoints SGLang uses today:

Define columns — use the existing tuple in that framework's parsing module (e.g. VLLM_RESULTS_COLUMNS in cvs/lib/inference/utils/vllm_parsing.py, ATOM_RESULTS_COLUMNS in cvs/lib/inference/utils/inference_suite_results_table.py). Same shape: (display_label, metric_key).

Record rows + columns — in the perf test, after building pass/fail rows:

record_benchmark_metric_rows(request.node, rows, columns=VLLM_RESULTS_COLUMNS)
Suite conftest hooks — in that framework's conftest.py:

pass columns=... (or benchmark_metric_columns_for_nodeid(nodeid)) to render_benchmark_metrics_html when attaching extras
call patch_benchmark_metrics_into_html(htmlpath, benchmark_test_name="<that suite's perf test name>") in pytest_sessionfinish

solaiys (approved):
## Re-review @ `fa74b41a` — all four points addressed, approving

Verified locally at `fa74b41a`. Every item from my last pass is closed, including the two I'd marked optional.

### The blocker is properly fixed

Both halves landed. `_FakeSubtests` now mirrors the real plugin — it captures the exception, records it in `failures`, and returns `True` to suppress — so the stub can finally express the failure case instead of encoding the opposite semantics. And `test_verify_inference_results_subtests_gates_violation` locks the invariant down with exactly the assertions that matter: `assertFalse(passed)`, `rows == {'mean_ttft_ms': 'fail', 'p99_ttft_ms': 'pass'}`, and `len(subtests.failures) == 1`. The middle one is the important one — it can only hold if suppression works and the loop continues past the first violation, so a regression in that behaviour now breaks a test rather than silently turning perf gating green.

I also re-confirmed the end-to-end path with real pytest-subtests at this head, since `sglang_common.py` changed: one violated and one satisfied threshold gives pytest **exit code 1**, `all_passed=False`, statuses `{'mean_ttft_ms': 'fail', 'p99_ttft_ms': 'pass'}`, and a `SUBFAILED` report. Gating is intact and now covered.

### Both non-blocking notes taken

The float truncation is gone — `Benchmark duration` and `Median TPOT` both use `([0-9\.]+)` now. I checked the behaviour that was actually at risk rather than just the diff: a log carrying `Median TPOT (ms): 10.9` and `Benchmark duration (s): 60.15` now parses to `'10.9'` and `'60.15'` instead of `'10'` and `'60'`, so a `median_tpot_ms` of 10.9 correctly fails a `max: 10` threshold where it previously passed. Thanks for picking that up even though it predated the PR.

The dmesg window guard is in all three suites, with `pytest` already imported in each, so a benchmark that never reaches verification now skips cleanly instead of handing `None` to the legacy parser.

### The layering suggestion went further than I asked

`perf_metric_table.py` has no sglang references at all now, and I checked the whole generic report layer: **zero named-suite imports remain anywhere outside `presets/`**, which is the invariant I was really after. Two things I liked beyond the suggestion:

- Rather than threading `columns` down through `patch_benchmark_metrics_into_html`, you store them per nodeid at record time (`_COLUMNS_BY_NODEID`) and look them up at patch time, so the columns travel with the rows and the patch call site stays clean. Better than what I proposed.
- `benchmark_test_name=SGLANG_PERF_BENCHMARK_TEST` is now passed explicitly at the sglang call site, which closes the `DEFAULT_BENCHMARK_TEST_NAME` point too. I confirmed the constant is the same string, so behaviour is unchanged.

The rename of `_metric_display_label` to public `metric_display_label` is clean — no stale references to the old private name, and the default `columns=()` still falls through to the prettify path, so any caller passing nothing behaves exactly as before.

### Gates

- Unit tests: 1152 → **1176**, `OK` on both the PR and the merge base. The four added tests cover the gating violation path, column storage, and both label branches.
- `ruff format --check` and `ruff check` clean across all 19 changed Python files; pylint 10.00/10 on the logging gate.
- Concurrency re-scan on the three new commits: no threads, async, multiprocessing, or background jobs introduced. The new `_COLUMNS_BY_NODEID` has the same profile as the existing `_ROWS_BY_NODEID` — process-local, single writer, and there's still no xdist in `requirements.txt` or `pytest.ini`.

Nice work on this one, particularly taking the two optional items and improving on the injection design. LGTM.

PR #353, opened 2026-08-19
AIMVT-298: Detect scheduler type (SPUR/SLURM/bare-metal)

speriaswamy-amd:
## Summary
- Adds `cvs/core/scheduler.py` with `detect_scheduler()` / `is_managed_compute()` to classify the cluster as SPUR, SLURM, or bare-metal, by probing for `spur version` / `scontrol version` on `PATH` (SPUR checked first, since it ships SLURM-compatible shims). Detection can be forced via the `CVS_SCHEDULER` env var.
- This is scoped to scheduler-type detection only (per ticket discussion) — no in-job rank/worker detection (`SLURM_PROCID`/`SLURM_NTASKS`/etc.), since those envs are only available once inside an active job/step.
- Adds unit tests under `cvs/core/unittests/test_scheduler.py` covering detection ordering, fallback to bare-metal, timeout/hang handling, and the `CVS_SCHEDULER` override (including normalization and invalid-value errors).

Ticket: AIMVT-298 (subtask of AIMVT-297, CVS in scheduler(SLURM/SPUR) environment)

## Test plan
- [x] `python -m unittest cvs.core.unittests.test_scheduler -v` — 13 tests pass
- [x] `ruff format --check` / `ruff check` clean

### DIFF
--- a/cvs/core/scheduler.py
+++ b/cvs/core/scheduler.py
@@ -1,0 +1,60 @@
+'''
+Copyright 2025 Advanced Micro Devices, Inc.
+All rights reserved. This notice is intended as a precaution against inadvertent publication and does not imply publication or any waiver of confidentiality.
+The year included in the foregoing notice is the year of creation of the work.
+All code contained here is Property of Advanced Micro Devices, Inc.
+'''
+
+import shutil
+import subprocess
+from enum import Enum
+import os
+
+CHECK_TIMEOUT = 5
+SCHEDULER_ENV_VAR = "CVS_SCHEDULER"
+
+
+class Scheduler(Enum):
+    SPUR = 'spur'
+    SLURM = 'slurm'
+    BARE_METAL = 'bare_metal'
+
+
+# SPUR must be checked before SLURM or a spur cluster gets misclassified.
+SCHEDULER_CHECK_COMMANDS = {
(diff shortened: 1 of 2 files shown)


cvs/core/scheduler.py
```diff
@@ -0,0 +1,60 @@
...
+
+class Scheduler(Enum):
+    SPUR = 'spur'
+    SLURM = 'slurm'
+    BARE_METAL = 'bare_metal'
+
+
+# SPUR must be checked before SLURM or a spur cluster gets misclassified.
+SCHEDULER_CHECK_COMMANDS = {
+    Scheduler.SPUR: [["spur", "version"]],
```
cijohnson, line 25: This doesn't mean the cvs run process is invoked using srun right?
For example if a managed cluster that did not block ssh connectivity,
1. user can login to head node (one of the managed compute)
2. install cvs and invoke cvs run , in this case as well scontrol cli will work and declare SLRUM env

So we should check 

os.environ.get("SLURM_STEP_ID") is not None and os.environ.get("SLURM_PROCID") is not None  along with SLURM_JOB_ID.

cijohnson (changes requested):

cijohnson (approved):

PR #354, opened 2026-08-20
Deterministic run layout and {run_dir} placeholder

atnair-amd:
## Summary

Every rank of a Slurm/SPUR job step has to derive the same run directory without coordinating, because those paths are the rendezvous for the agent control plane: worker ranks never enter pytest and read rank0's port out of `agent_dir`. `RunLayout` resolves workspace / run_dir / agent_dir once in `cvs run`, before pytest or agent bootstrap, and publishes it as `CVS_RUN_DIR`.

Ticket: Fixes AIMVT-300 (epic AIMVT-297)

## Change

- `--workspace`, falling back to `$CVS_WORKSPACE` then to a `cvs_runs` directory beside the venv. Run directory is `<workspace>/cvs/runs/<run_id>`.
- run_id is `<job>.<step>[.r<restart>]` under a scheduler, else `local-<timestamp>`. The job id alone is not enough: concurrent steps in one allocation share it and would share an `agent_dir`, and a requeued job repeats it.
- Keyed on the `SLURM_*` environment rather than `is_managed_compute()`, which also probes for the `scontrol`/`spur` binaries. Those are absent from the CVS container image even when srun exported the variables, so ranks would each fall back to their own wall clock and hang at the rendezvous instead of erroring.
- `{run_dir}` placeholder for test configs. Resolved from the environment because `cvs/lib/utils_lib.py` cannot import `cvs.core` at module level — the orchestrator factory imports it back. Using it outside `cvs run` exits with a diagnostic rather than substituting an empty path.
- The container orchestrator bind-mounts the run directory at the same path in scheduler-managed runs, so a config value substituted on the host resolves inside the container. Not mounted otherwise: nothing guarantees the workspace exists on that node, and a bind mount of a missing path quietly creates a root-owned empty directory.
- Out of scope: the agent bootstrap and port rendezvous that consume `agent_dir` (AIMVT-297).

## Tests

50 unit tests across `cvs/core/unittests/test_run_layout.py`, `test_container.py`, `test_run_plugin.py` and `test_utils_lib.py`, covering rank agreement, requeue and concurrent-step collisions, unwritable workspaces, and the no-venv case.

Gate: `make fmt-check`, `make lint` (pylint 10.00/10), `make ut` — 1245 tests, all passing.


atnair-amd:
Split into two stacked PRs: #355 (RunLayout and the {run_dir} placeholder) and #356 (the container mount, stacked on #355).

PR #355, opened 2026-08-20
Deterministic run layout and {run_dir} placeholder

atnair-amd:
## Summary

Every rank of a Slurm/SPUR job step has to derive the same run directory without coordinating, because those paths are the rendezvous for the agent control plane: worker ranks never enter pytest and read rank0's port out of `agent_dir`. `RunLayout` resolves workspace / run_dir / agent_dir once in `cvs run`, before pytest or agent bootstrap, and is the single source of truth for those paths thereafter.

Ticket: Fixes [AIMVT-300](https://amd-hub.atlassian.net/browse/AIMVT-300) (story [AIMVT-297](https://amd-hub.atlassian.net/browse/AIMVT-297))

## Change

- `--workspace`, falling back to `$CVS_WORKSPACE` then to the venv's parent directory. Run directory is `<workspace>/cvs_runs/<run_id>`.
- run_id is the scheduler's job id, else `local-<timestamp>`. Taken as-is: it is the one name every rank already agrees on, and how finely a job subdivides into runs is a question for whatever needs that granularity, not for the module that holds the paths.
- Managed-or-not comes from `is_managed_compute()` (AIMVT-298), so the layout cannot disagree with the rest of CVS about what kind of run this is. Where the `scontrol`/`spur` probe cannot see the scheduler — the CVS container image ships neither — `CVS_SCHEDULER` is the existing override; without it the run is unmanaged and each rank would take its own timestamp.
- `{run_dir}` placeholder for test configs, read off the layout. `cvs/lib/utils_lib.py` imports `RunLayout` inside the function rather than at module level: importing `cvs.core.run_layout` pulls in `cvs/core/__init__.py`, whose orchestrator factory reaches `cvs/core/orchestrators/baremetal.py`, which imports `utils_lib` back. Deferring to call time is enough — `pytest.main()` runs in-process, so the resolver sees the same layout object `cvs run` built. The resolver only asks for a layout when the token is actually present: `RunLayout.get()` creates directories, and roughly half the test modules call it with configs that never mention `{run_dir}`.
- Out of scope: passing the run directory through to the container orchestrator, which stacks on this branch as #356; and the agent bootstrap that consumes `agent_dir` ([AIMVT-297](https://amd-hub.atlassian.net/browse/AIMVT-297)).

## Tests

31 unit tests: `cvs/core/unittests/test_run_layout.py` (18), `cvs/cli_plugins/unittests/test_run_plugin.py` (9), `cvs/lib/unittests/test_utils_lib.py` (4). They cover workspace precedence, rank agreement, unwritable workspaces, the no-venv case, that the derived default composes the same path shape as an explicit workspace, that an environment disagreeing with the layout cannot redirect where artifacts land, and that a mistyped suite name exits without creating a run directory. Run-id tests pin `CVS_SCHEDULER` so they do not depend on whether the host has scheduler binaries installed. What makes a run "managed" is `cvs.core.scheduler`'s contract and is tested there, not re-tested through this module.

Gate: `make fmt-check`, `make lint` (pylint 10.00/10), `make ut` — 1226 tests, all passing.


cvs/cli_plugins/run_plugin.py
```diff
@@ -64,6 +78,7 @@ def run(self, args):
             args.log_level,
             args.capture,
             getattr(args, "extra_pytest_args", []),
+            workspace=args.workspace,
```
cijohnson: Let us move this after line 74(args.config_file), we no need to maintain default value in run_test for workspace argument because arg parser at line 23 already set the default value to None

cvs/cli_plugins/run_plugin.py
```diff
@@ -98,6 +113,7 @@ def run_test(
         log_level,
         capture,
         extra_pytest_args,
+        workspace=None,
```
cijohnson: we can move it after config_file arg

cvs/cli_plugins/run_plugin.py
```diff
@@ -111,6 +127,17 @@ def run_test(
 
         test_file = self.get_test_file(module_path)
 
+        # Resolve the run layout before pytest, but only once the run is going to
+        # happen: worker ranks in a Slurm/Spur job never enter pytest at all, and
+        # config placeholder resolution needs the layout resolved, so this cannot
+        # wait for a fixture -- while creating directories for a mistyped suite
+        # name would litter shared storage.
+        try:
+            RunLayout.initialize(workspace)
```
cijohnson: [3] we should do RunLayout initialization in all the ranks, so we should move to def run method

`
def run(self, args):
    try:
        RunLayout.initialize(args.workspace)   # ALL ranks, before any rank branching
    except RuntimeError as e:
        print(f"Error: {e}")
        sys.exit(1)

    self.run_test(...) 
`

cvs/cli_plugins/run_plugin.py
cijohnson: when comment [3] is taken care this comment will become invalid

cvs/cli_plugins/run_plugin.py
cijohnson: when comment [3] is taken care this comment will become invalid

cvs/core/run_layout.py
```diff
@@ -0,0 +1,128 @@
...
+
+class RunLayout:
+    '''The run's directory layout, resolved once per process.'''
+
+    _instance = None
+
+    def __init__(self, workspace, run_id):
+        self.workspace = workspace
+        self.run_id = run_id
+        self.run_dir = workspace / "cvs" / "runs" / run_id
```
cijohnson: lets keep it         self.run_dir = workspace / "cvs_runs" / run_id

jira was incorrectly using cvs/runs

cvs/core/run_layout.py
```diff
@@ -0,0 +1,128 @@
...
+        self.agent_dir = self.run_dir / "agent"
+
+    @classmethod
+    def initialize(cls, workspace=None):
```
cijohnson: Consider collapsing initialize() / instance() / instance_or_none() into a single RunLayout.get(workspace=...) — workspace passed → resolve and init idempotently; no args → return existing or None. Same behavior, less API surface.

you can have _initialize but call it from get method.
users of the RunLayout now only need to do one these

RunLayout.get(workspace)  # init or return existing (workspace=None → env/default)
RunLayout.get()            # existing or None

cijohnson (changes requested):

cijohnson (approved):

PR #360, opened 2026-08-21
feat(JAX): jaxmaxtext model configs + smoke & checkpoint-resume tests + incremental log streaming

solaiys:
## Summary
Extends the `jaxmaxtext_single` / `jaxmaxtext_distributed` training suites with new
model configs, a smoke test, a checkpoint save+resume (with I/O timing) test, and
fixes noisy poll logging. 6 commits on top of `origin/main`.
### 1. `2e31e64` feat(JAX) Added more model configs and threshold files for jaxmaxtext
- New config + threshold pairs for **llama-3.1-8b**, **deepseek-v2-lite**, and
  **llama-3.1-405b** (MI300X/MI325X/MI355X as applicable).
- Config files aligned with the MAD repo reference (env vars, XLA flags, parallelism,
  MoE knobs).
### 2. `089c0d9` fix(JAX) Updated llama3.1-405b model config with BS and xla_autotune=0
- Tuned `per_device_batch_size` and set `xla_gpu_autotune_level=0` for the 405B config
  (+ matching threshold tweaks).
### 3. `ab6fe09` feat(JAX) Added Smoke Test for jaxmaxtext training
- New `test_smoke` stage (both suites): loads the model and runs 10 steps with a small
  fixed batch/seqlen in BF16. Pass = completes the steps with no error signature; no
  metric/threshold checks. A failure sets `lifecycle.failed` so downstream stages skip
  (early gate before the full sweeps).
### 4. `ec65b2d` feat(JAX) Added Checkpoint save and resume I/O timing metrics test
- New opt-in `test_checkpoint_resume` stage (`training.checkpoint_resume.enabled`):
  - **Phase 1** trains `steps_before_ckpt` with checkpointing on (saved at
    `checkpoint_period`); **Phase 2** resumes and trains `steps_after_resume` more.
  - Pass = Phase 2 restarts from the checkpoint step (non-zero first step) **and** the
    boundary loss matches Phase 1 within `loss_tolerance` (state restored, not
    reinitialized).
  - Benchmarks **checkpoint_save_seconds / checkpoint_load_seconds** (parsed from orbax
    logs), gated against `max_save_seconds` / `max_load_seconds` when > 0 (else
    record-only), rendered as a separate "Checkpoint I/O" section in the results.
  - `delete_ckpt_dir` (default true) cleans the checkpoint dir after the test.
- `checkpoint_resume` schema block added to all jaxmaxtext configs (default `enabled=false`).
> ⚠️ **Validation scope:** the checkpoint save+resume test has been validated **only with
> the small `llama3.1-8b` config on MI325X**. It has **not** yet been validated on the
> larger models (70B/405B/DeepSeek). Treat those as untested for this stage.
### 5. `94c1ffe` fix(exec_cmd_list) Exposed print_console arg to enable/disable
- Threaded a `print_console` argument through `ContainerOrchestrator.exec_cmd_list` and
  the Docker runtime (default `True`; pass `False` to fetch bulk output without
  re-echoing it to the console). Pssh already supported it.
### 6. `3522040` fix(JAX) Added logic to print training log incrementally
- `poll_for_completion()` no longer re-dumps a `tail -N` window every poll (which bloated
  `--log-file` with repeated content). It now **streams only the new node-0 log lines
  once** per poll (via a per-node line cursor), scans **all** nodes' new chunks for error
  signatures, and prints a compact `[poll]` heartbeat so polling remains visible.
- Fixed ruff fmt & lint.
## Test plan
- `make fmt-check` / `make lint` — clean on the changed files.
- Unit tests: `test_jaxmaxtext_training_lib`, `test_maxtext_parsing`,
  `test_training_config_loader` pass (incl. new smoke/checkpoint/drain coverage).
- Smoke + training + metric + loss-curve stages exercised on jaxmaxtext_distributed.
- Checkpoint save+resume validated end-to-end on **MI325X llama3.1-8b** (save/load
  timings captured, loss continuity within tolerance, resume from checkpoint step).

### DIFF
--- a/cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
+++ b/cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
@@ -123,6 +123,12 @@ class MaxTextTrainingJob:
         self._scratch_dir = None  # resolved lazily to /tmp/<user>/jax
         self._train_script = None  # resolved lazily to the first existing candidate
 
+        # Per-node cursor (lines already surfaced) so polling STREAMS only the
+        # new training-log lines to the console once, instead of re-dumping a
+        # `tail -N` window every iteration (which bloated --log-file with
+        # repeated content).
+        self._log_line_cursor = [0] * self.num_nodes
+
     def _get_scratch_dir(self):
         """User-namespaced in-container scratch base (``/tmp/<user>/jax``).
 
@@ -394,6 +400,9 @@ class MaxTextTrainingJob:
 
         self.orch.exec_cmd_list(launch_cmds)
 
+        # Fresh run -> stream the log from the top.
+        self._log_line_cursor = [0] * self.num_nodes
+
         time.sleep(self._initial_wait_s)
 
     # ---------- polling ----------
@@ -414,7 +423,7 @@ class MaxTextTrainingJob:
             f"{shlex.quote(f'{self.out_dir}/out-node{i}/training.log')} 2>/dev/null || true"
             for i in range(self.num_nodes)
         ]
-        out = self.orch.exec_cmd_list(cmd_list)
+        out = self.orch.exec_cmd_list(cmd_list, print_console=False)
         if not out or len(out) < self.num_nodes:
             return False
         for _host, result in out.items():
@@ -424,32 +433,56 @@ class MaxTextTrainingJob:
                 return False
         return True
 
-    def _scan_for_errors(self):
-        """Scan each node's own training log for known error patterns.
+    def _scan_chunk_for_errors(self, host, i, text):
+        """Raise on the first known error signature (or NaN/Inf) in `text`."""
+        if not text:
+            return
+        if _NAN_INF_RE.search(text):
+            raise RuntimeError(f"NaN/Inf in training metrics on {host} (node {i}): {text[-500:]}")
+        for err_name, err_pattern in self.error_patterns.items():
+            if not err_pattern:
+                continue
+            if re.search(err_pattern, text, re.I):
+                raise RuntimeError(f"Training error '{err_name}' on {host} (node {i}): {text[-500:]}")
 
-        Reads all nodes' logs in one parallel ``orch.exec_cmd_list`` call
-        (``cmd_list[i]`` runs on ``hosts[i]``). Raises on the first match.
+    def _drain_new_log_lines(self):
+        """Fetch training-log lines written since the last poll, on every node,
+        in one parallel call, and advance each node's cursor.
+
+        Returns ``{node_index: new_text}``. ``print_console=False`` so the
+        orchestrator does NOT re-echo the bulk output -- the caller decides what
+        to surface (we stream node 0 and scan every node for errors). This is
+        what makes the console/--log-file carry each log line ONCE instead of a
+        repeated ``tail`` window per poll.
         """
         cmd_list = [
-            f"tail -2000 {shlex.quote(f'{self.out_dir}/out-node{i}/training.log')} 2>/dev/null"
+            f"tail -n +{self._log_line_cursor[i] + 1} "
+            f"{shlex.quote(f'{self.out_dir}/out-node{i}/training.log')} 2>/dev/null || true"
             for i in range(self.num_nodes)
         ]
-        out = self.orch.exec_cmd_list(cmd_list)
+        out = self.orch.exec_cmd_list(cmd_list, print_console=False)
         node_of = {h: i for i, h in enumerate(self.orch.hosts)}
-        for host, text in (out or {}).items():
-            text = text if isinstance(text, str) else (text or {}).get("output", "")
+        new_by_node = {}
+        for host, result in (out or {}).items():
+            text = result if isinstance(result, str) else (result or {}).get("output", "")
             text = text or ""
-            i = node_of.get(host, "?")
-            if _NAN_INF_RE.search(text):
-                raise RuntimeError(f"NaN/Inf in training metrics on {host} (node {i}): {text[-500:]}")
-            for err_name, err_pattern in self.error_patterns.items():
-                if not err_pattern:
-                    continue
-                if re.search(err_pattern, text, re.I):
-                    raise RuntimeError(f"Training error '{err_name}' on {host} (node {i}): {text[-500:]}")
+            i = node_of.get(host)
+            if i is None or not text:
+                continue
+            # Advance the cursor by the number of newly read lines so the next
+            # poll starts right after them.
+            self._log_line_cursor[i] += len(text.splitlines())
+            new_by_node[i] = text
+        return new_by_node
 
     def poll_for_completion(self, timeout_s=None):
-        """Poll is_complete() with error scanning until training finishes or times out."""
+        """Poll until training finishes or times out.
+
+        Each iteration streams only the NEW training-log lines (node 0 to the
+        console; all nodes scanned for error signatures), then checks for the
+        completion marker. A concise ``[poll]`` heartbeat makes the internal
+        polling visible without re-dumping the log.
+        """
         if timeout_s is None:
             timeout_s = self._poll_count * self._poll_wait_s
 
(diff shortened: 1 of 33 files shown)


cvs/tests/training/jaxmaxtext/_common.py
```diff
@@ -220,6 +222,73 @@ def setup_tokenizer(orch, variant_config, hf_token, lifecycle, request):
...
+    job = MaxTextTrainingJob(orch, smoke_variant, hf_token, sweep=smoke_sweep)
+    t = time.monotonic()
+    try:
+        job.setup_training_env()
+        job.build_training_cmd()
+        job.start_training()
+        # poll scans each node's log for error_patterns/NaN every iteration and
+        # raises on the first match or on timeout; returns cleanly once the run
+        # reaches step _SMOKE_STEPS-1.
+        job.poll_for_completion()
```
solaiys, line 271: **Self-review — leftover SMOKE logs can make this stage pass without training.**

Smoke's only pass criterion is `poll_for_completion()` → `is_complete()`, which greps the on-disk `training.log` for `completed step: 9,`. Unlike `checkpoint_resume` (which `rm -rf`s `out_dir` before Phase 1), smoke never clears `{log_dir}/jaxmaxtext/SMOKE/`. A previous successful smoke run leaves that marker in place.

`start_training()` does `nohup bash launcher &` and sleeps 60s. The launcher only truncates the log when it reaches `python ... | tee training.log`. If this run dies before `tee` opens the file (env `source` fails, launcher not actually started, etc.), the **old** log is still there, `is_complete()` is true on the first poll, and smoke PASSES. There is no `parse_results()` / step-count check that would catch "this log is from yesterday".

Because a smoke failure sets `lifecycle.failed` and skips the rest of the suite, the inverse (false pass) is the fail-open case: the early gate does not actually gate.

**Fix:** delete or truncate the per-node `training.log` (or the SMOKE `out_dir`) on every node *before* launch — same idea as the Phase 1 `rm -rf` in `checkpoint_resume`. Even better, do it in `MaxTextTrainingJob.start_training()` so `training_run` gets the same protection. A cheap second check: after poll, `parse_results()` and require `len(step_metrics) >= _SMOKE_STEPS` (or a mtime newer than `training_start_time`).

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
```diff
@@ -452,16 +492,28 @@ def poll_for_completion(self, timeout_s=None):
...
+                log.info("[train node0]\n%s", node0_new)
 
             if self.is_complete():
+                # Flush any tail lines written since the drain above (e.g. the
+                # final "completed step" marker) so the stream ends cleanly.
+                tail_new = (self._drain_new_log_lines().get(0) or "").rstrip()
+                if tail_new:
+                    log.info("[train node0]\n%s", tail_new)
                 log.info("training complete (poll iter=%d, %.0fs elapsed)", it, elapsed)
                 return
```
solaiys, line 535: **Self-review — the completion-path drain is not error-scanned, and non-0 nodes are dropped.**

The new poller contract is: every newly read chunk on **every** node is passed through `_scan_chunk_for_errors` before we declare success. That holds for the loop body, but not here:

```python
if self.is_complete():
    tail_new = (self._drain_new_log_lines().get(0) or "").rstrip()
    if tail_new:
        log.info("[train node0]\n%s", tail_new)
    ...
    return
```

Two gaps:

1. `tail_new` is the lines written in the SSH round-trip between the already-scanned drain and `is_complete()`. That is exactly where the final `completed step: N` line (and anything after it — shutdown traceback, last-step NaN) lands. We print node 0's chunk and return success without scanning it.
2. `_drain_new_log_lines()` returns `{node_index: text}` for **all** nodes; `.get(0)` throws away every other node's final tail, so a worker-only error in that window is never scanned and never logged.

Smoke and `training_run` both treat "poll returned" as "no error signature". This path can return success with an error sitting in the last chunk.

**Fix:**

```python
if self.is_complete():
    tail = self._drain_new_log_lines()
    for i, text in tail.items():
        self._scan_chunk_for_errors(self.orch.hosts[i], i, text)
    node0 = (tail.get(0) or "").rstrip()
    if node0:
        log.info("[train node0]\n%s", node0)
    ...
    return
```

Worth a unit test that stubs `is_complete() -> True` and a second drain containing `TFLOP/s/device: NaN` and asserts `poll_for_completion` raises.

solaiys (commented):
## Self-review

Looked through this PR as the author, focusing on the new smoke / checkpoint-resume stages, the incremental log poller, and whether leftover artifacts can make a stage pass without this run actually doing the work. CI is green; the items below are logic holes the unit tests do not cover.

| # | Issue | Where |
|---|---|---|
| 1 | Smoke (and `start_training` in general) never clears `training.log`, so `is_complete()` can succeed from a **previous** SMOKE run | `_common.py` smoke / `start_training` |
| 2 | Completing poll fetches the final log tail and then **does not scan it for errors** (and drops every node except 0) | `poll_for_completion` |

I did not re-raise config-label nits (GBS in sweep names, `log_period` comments vs values) or the untested-on-405B caveat already in the PR body.

cvs/tests/training/jaxmaxtext/_common.py
solaiys: Fixed.

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
solaiys: Fixed.

cvs/tests/training/jaxmaxtext/_common.py
```diff
@@ -288,6 +357,247 @@ def training_run(orch, variant_config, hf_token, sweep_name, training_res_dict,
...
+    # A fresh (non-resumed) run logs step 0 first; a restored run starts at the
+    # checkpoint step. So a non-zero first step means the checkpoint (step +
+    # weights + optimizer) was restored.
+    resumed_step = p2_steps[0]["step"] if p2_steps else None
+    resume_ok = resumed_step is not None and resumed_step > 0
+
+    p1_last_loss = next((s["loss"] for s in reversed(p1_steps) if isinstance(s.get("loss"), (int, float))), None)
+    p2_first_loss = next((s["loss"] for s in p2_steps if isinstance(s.get("loss"), (int, float))), None)
+    loss_delta = abs(p2_first_loss - p1_last_loss) if (p1_last_loss is not None and p2_first_loss is not None) else None
+    loss_ok = loss_delta is not None and loss_delta <= cfg.loss_tolerance
```
amd-droy, line 550: loss continuity may compare different steps. The resume check compares Phase 1’s last loss to Phase 2’s first loss without requiring the step numbers to match.
If MaxText resumes and the first logged Phase 2 step is checkpoint_period + 1 (rather than re-emitting the checkpoint step), you can get a false failure even with a correct restore. Safer approach: find the checkpoint step from Phase 1 (or _latest_checkpoint_step_path), then compare loss at that same step in both phases.

cvs/tests/training/jaxmaxtext/_common.py
```diff
@@ -288,6 +357,247 @@ def training_run(orch, variant_config, hf_token, sweep_name, training_res_dict,
...
+    # push checkpoint_period beyond the total step count so no NEW periodic
+    # checkpoint is written while we train the extra steps.
+    total_steps = cfg.steps_before_ckpt + cfg.steps_after_resume
+    if not _latest_checkpoint_step_path(orch, ckpt_dir):
+        log.warning(
+            "[checkpoint] no saved checkpoint found under %s before Phase 2; "
+            "resume/loss-continuity checks will likely fail",
+            ckpt_dir,
+        )
+    job2 = _mk_job(total_steps, {"checkpoint_period": total_steps + 1000}, True)
```
amd-droy, line 485: If steps_before_ckpt < checkpoint_period, Phase 1 can finish without saving anything; this only warns and continues into Phase 2.

amd-droy (approved):
looks good to me. Thanks Saravanan.

PR #361, opened 2026-08-24
feat(Megatron): Megatron checkpoint save and resume test, convergence tracking, and Primus factory dispatch

sukesh-amd:
## Summary

  **Commit 1 — Convergence tracking and steps/time-to-target thresholds**
  - Add `ConvergenceConfig` (`target_metric`, `target_value`) to `MegatronVariantConfig`
  - Add `convergence.py` utility (`parse_step_metrics`, `compute_convergence`)
  - Wire convergence computation into `test_training` in both single and distributed
  - Add convergence block and `steps_to_target` / `time_to_target_seconds` info
    thresholds to `mi325x_megatron_llama-3.1-8b` config and threshold file

**Commit 2 — Primus factory dispatch and training skip propagation fix**
  - Add `primus_lib.py` with `PrimusTrainingJob`, training completion detection,
    result/iteration/NaN patterns, and `_parse_step_losses` for checkpoint tests
  - Add `_make_training_job()` factory in both test files: returns `PrimusTrainingJob`
    when image contains `primus`, otherwise `MegatronTrainingJob`
  - Fix per-combo failure isolation: set `train_res_dict[combo_key] = None` on exception
    so `test_metric` / `test_loss_curve` skip correctly without blocking other combos

  **Commit 3 — Checkpoint save+resume test with I/O timing**
  - Add `CheckpointConfig` pydantic model (`enforce`, `save_interval`, `save_iters`,
    `resume_iters`, `loss_rtol`, `checkpoint_dir`) wired into `MegatronVariantConfig`
  - Add `test_checkpoint` to single and distributed: two-phase test that saves a
    checkpoint, resumes from it, and verifies step counter restoration and loss continuity
  - Add `checkpoint_io.py`: parses save/load I/O times from Primus logs using embedded
    microsecond timestamps; strips ANSI escape codes before matching
  - Distributed: reads node-0 log for timing (rank-0 only emits checkpoint lines),
    last-node log for losses; adds `--auto_detect_ckpt_format` on load; requires
    `checkpoint_dir` (shared NFS path) in config with volume mount entry
  - Add `_read_node_log(node_idx)` to `PrimusTrainingJob`
  - Add checkpoint block to `mi325x_megatron_llama-3.1-8b_single.json`,
    `mi325x_megatron_deepseek-v2-lite_single.json`, and
    `mi325x_megatron_llama-3.3-70b_distributed.json`
  - Insert `test_checkpoint` at stage 3 in conftest test ordering

  ## Test plan
  - [ ] Single-node: run `test_checkpoint` with `enforce=true` on a Primus image; verify save/load I/O times
  - Add `CheckpointConfig` pydantic model (`enforce`, `save_interval`, `save_iters`,
    `resume_iters`, `loss_rtol`, `checkpoint_dir`) wired into `MegatronVariantConfig`
  - Add `test_checkpoint` to single and distributed: two-phase test that saves a
    checkpoint, resumes from it, and verifies step counter restoration and loss continuity
  - Add `checkpoint_io.py`: parses save/load I/O times from Primus logs using embedded
    microsecond timestamps; strips ANSI escape codes before matching
  - Distributed: reads node-0 log for timing (rank-0 only emits checkpoint lines),
    last-node log for losses; adds `--auto_detect_ckpt_format` on load; requires
    `checkpoint_dir` (shared NFS path) in config with volume mount entry
  - Add `_read_node_log(node_idx)` to `PrimusTrainingJob`
  - Add checkpoint block to `mi325x_megatron_llama-3.1-8b_single.json`,
    `mi325x_megatron_deepseek-v2-lite_single.json`, and
    `mi325x_megatron_llama-3.3-70b_distributed.json`
  - Insert `test_checkpoint` at stage 3 in conftest test ordering

### DIFF
--- a/cvs/lib/training/megatron/primus_lib.py
+++ b/cvs/lib/training/megatron/primus_lib.py
@@ -396,8 +396,8 @@ class PrimusTrainingJob:
             f'--global_batch_size {self.global_batch_size} '
             f'--train_iters {self.iterations}'
         )
-        if self.checkpoint_dir:
-            batch_args += f' --save {self.checkpoint_dir} --save_interval {self.save_interval or self.iterations}'
+        if self.checkpoint_dir and not self.load_checkpoint:
+            batch_args += f' --save_interval {self.save_interval or self.iterations}'
         if self.load_checkpoint and self.checkpoint_dir:
             batch_args += f' --load {self.checkpoint_dir}/checkpoints'
             if self.distributed_training:
@@ -421,9 +421,9 @@ class PrimusTrainingJob:
                     + f'-- train pretrain --config {exp_path} {batch_args} &'
                 )
                 script_cmd = (
-                    f'echo {shlex.quote(full_cmd)} > '
+                    f'umask 077; echo {shlex.quote(full_cmd)} > '
                     f'{self.scripts_dir}/distributed_wrapper_script_{i}.sh '
-                    f'&& chmod 777 {self.scripts_dir}/distributed_wrapper_script_{i}.sh'
+                    f'&& chmod 700 {self.scripts_dir}/distributed_wrapper_script_{i}.sh'
                 )
                 self.job_cmd_list.append(script_cmd)
         else:
@@ -481,8 +481,8 @@ class PrimusTrainingJob:
             )
         else:
             self.orch.all.exec(
-                f'echo {shlex.quote(self.job_cmd)} > {self.scripts_dir}/single_node_wrapper_script.sh '
-                f'&& chmod 777 {self.scripts_dir}/single_node_wrapper_script.sh'
+                f'umask 077; echo {shlex.quote(self.job_cmd)} > {self.scripts_dir}/single_node_wrapper_script.sh '
+                f'&& chmod 700 {self.scripts_dir}/single_node_wrapper_script.sh'
             )
             self.orch.exec(
                 f'nohup {self.scripts_dir}/single_node_wrapper_script.sh '
--- a/cvs/lib/training/megatron/primus_lib.py
+++ b/cvs/lib/training/megatron/primus_lib.py
@@ -385,6 +385,17 @@ class PrimusTrainingJob:
                 'export NVTE_CK_IS_V3_ATOMIC_FP32=1; '
             )
 
+        # PRIMUS_WORKSPACE controls where Primus writes checkpoints.
+        # With PRIMUS_TEAM/USER/EXP_NAME all empty the checkpoint lands at:
+        #   {PRIMUS_WORKSPACE}/checkpoints
+        if self.checkpoint_dir:
+            env_exports += (
+                f'export PRIMUS_WORKSPACE={self.checkpoint_dir}; '
+                f'export PRIMUS_TEAM=""; '
+                f'export PRIMUS_USER=""; '
+                f'export PRIMUS_EXP_NAME=""; '
+            )
+
         batch_args = (
             f'--micro_batch_size {self.micro_batch_size} '
             f'--global_batch_size {self.global_batch_size} '
@@ -396,7 +407,9 @@ class PrimusTrainingJob:
                 f' --save_interval {self.save_interval or self.iterations}'
             )
         if self.load_checkpoint and self.checkpoint_dir:
-            batch_args += f' --load {self.checkpoint_dir}'
+            batch_args += f' --load {self.checkpoint_dir}/checkpoints'
+            if self.distributed_training:
+                batch_args += ' --auto_detect_ckpt_format'
 
         if self.distributed_training:
             env_exports += (
@@ -482,16 +495,19 @@ class PrimusTrainingJob:
             )
         time.sleep(50)
 
-    def _read_last_node_log(self, tail_lines=0):
-        """Read training log from the last node."""
-        n = len(self.orch.hosts)
-        last_host = self.orch.hosts[-1]
+    def _read_node_log(self, node_idx: int, tail_lines: int = 0) -> str:
+        """Read training log from a specific node by index."""
+        host = self.orch.hosts[node_idx]
         tail_suffix = f' | tail -{tail_lines}' if tail_lines > 0 else ''
         out_dict = self.orch.exec(
-            f'cat {self.combo_log_dir}/out-node{n - 1}/training.log{tail_suffix}',
-            hosts=[last_host],
+            f'cat {self.combo_log_dir}/out-node{node_idx}/training.log{tail_suffix}',
+            hosts=[host],
         )
-        return out_dict.get(last_host) or ''
+        return out_dict.get(host) or ''
+
+    def _read_last_node_log(self, tail_lines=0):
+        """Read training log from the last node."""
+        return self._read_node_log(len(self.orch.hosts) - 1, tail_lines)
 
     def _get_training_results_dict(self):
         """Parse training log and extract Primus metrics."""
--- a/cvs/lib/training/megatron/primus_lib.py
+++ b/cvs/lib/training/megatron/primus_lib.py
@@ -64,6 +64,7 @@ PRIMUS_NAN_PATTERNS = [
     r'lm loss:\s+(?:NaN|Inf)',
 ]
 
+
 def _is_training_complete(output, total_iters):
     """Return True only when the final Primus iteration line is present.
 
@@ -119,6 +120,7 @@ def _parse_step_losses(log_text):
 # ---------------------------------------------------------------------------
 # PrimusTrainingJob
 # ---------------------------------------------------------------------------
+
 
 class PrimusTrainingJob:
     """Manages a single Primus-Megatron training run (single-node or distributed).
@@ -183,7 +185,8 @@ class PrimusTrainingJob:
         if int(self.nnodes) != len(orch.hosts):
             log.warning(
                 'config nnodes=%s does not match cluster host count=%d; using cluster host count',
-                self.nnodes, len(orch.hosts),
+                self.nnodes,
+                len(orch.hosts),
             )
             self.nnodes = str(len(orch.hosts))
 
@@ -307,8 +310,7 @@ class PrimusTrainingJob:
 
         if not self.training_results_dict:
             fail_test(
-                'Failed to populate training results, training_results_dict is empty'
-                ' - please check logs for failures'
+                'Failed to populate training results, training_results_dict is empty - please check logs for failures'
             )
             return
 
@@ -316,8 +318,7 @@ class PrimusTrainingJob:
             for val in result_vals:
                 if re.search('nan|inf', val, re.I):
                     fail_test(
-                        f'Failures seen in training_result dict for {result_key},'
-                        f' numbers are either NaN or Inf - {val}'
+                        f'Failures seen in training_result dict for {result_key}, numbers are either NaN or Inf - {val}'
                     )
 
         if self.distributed_training and self.verify_network_errors.lower() == 'true':
@@ -362,10 +363,7 @@ class PrimusTrainingJob:
 
     def _exp_config_path(self):
         """Return the Primus EXP YAML path relative to primus_root."""
-        return (
-            f'examples/megatron/configs/'
-            f'{self.gpu_arch}/{self.model_name}-{self.precision}-pretrain.yaml'
-        )
+        return f'examples/megatron/configs/{self.gpu_arch}/{self.model_name}-{self.precision}-pretrain.yaml'
 
     def _build_primus_cmd(self):
         """Build the primus-cli launch command per node."""
@@ -380,10 +378,7 @@ class PrimusTrainingJob:
         )
 
         if re.search(r'MI3(00|25)X', self.gpu_arch, re.I):
-            env_exports += (
-                'export PRIMUS_TURBO_ATTN_V3_ATOMIC_FP32=1; '
-                'export NVTE_CK_IS_V3_ATOMIC_FP32=1; '
-            )
+            env_exports += 'export PRIMUS_TURBO_ATTN_V3_ATOMIC_FP32=1; export NVTE_CK_IS_V3_ATOMIC_FP32=1; '
 
         # PRIMUS_WORKSPACE controls where Primus writes checkpoints.
         # With PRIMUS_TEAM/USER/EXP_NAME all empty the checkpoint lands at:
@@ -402,10 +397,7 @@ class PrimusTrainingJob:
             f'--train_iters {self.iterations}'
         )
         if self.checkpoint_dir:
-            batch_args += (
-                f' --save {self.checkpoint_dir}'
-                f' --save_interval {self.save_interval or self.iterations}'
-            )
+            batch_args += f' --save {self.checkpoint_dir} --save_interval {self.save_interval or self.iterations}'
         if self.load_checkpoint and self.checkpoint_dir:
             batch_args += f' --load {self.checkpoint_dir}/checkpoints'
             if self.distributed_training:
@@ -458,7 +450,8 @@ class PrimusTrainingJob:
             fallback_arch = 'MI300X'
             log.warning(
                 'EXP config not found for %s, retrying with fallback arch %s',
-                self.gpu_arch, fallback_arch,
+                self.gpu_arch,
+                fallback_arch,
             )
             self.gpu_arch = fallback_arch
             self.job_cmd = ''
@@ -479,11 +472,13 @@ class PrimusTrainingJob:
 
         if self.distributed_training:
             self.orch.all.exec_cmd_list(self.job_cmd_list)
-            self.orch.exec_cmd_list([
-                f'nohup {self.scripts_dir}/distributed_wrapper_script_{i}.sh '
-                f'>> {self.combo_log_dir}/out-node{i}/training.log 2>&1 &'
-                for i in range(n)
-            ])
+            self.orch.exec_cmd_list(
+                [
+                    f'nohup {self.scripts_dir}/distributed_wrapper_script_{i}.sh '
+                    f'>> {self.combo_log_dir}/out-node{i}/training.log 2>&1 &'
+                    for i in range(n)
+                ]
+            )
         else:
             self.orch.all.exec(
                 f'echo {shlex.quote(self.job_cmd)} > {self.scripts_dir}/single_node_wrapper_script.sh '
--- a/cvs/lib/training/megatron/primus_lib.py
+++ b/cvs/lib/training/megatron/primus_lib.py
@@ -1,0 +1,545 @@
+'''
+Copyright 2025 Advanced Micro Devices, Inc.
+All rights reserved. This notice is intended as a precaution against inadvertent publication and does not imply publication or any waiver of confidentiality.
+The year included in the foregoing notice is the year of creation of the work.
+All code contained here is Property of Advanced Micro Devices, Inc.
+
+Primus-Megatron training job — standalone module for Primus-specific logic.
+
+PrimusTrainingJob has the same public interface as MegatronTrainingJob and can
+be used as a drop-in replacement when the container image is Primus-based:
+  build_training_job_cmd()
+  start_training_job()
+  poll_for_training_completion()
+  verify_training_results()
+  stop_training_processes()
+  download_tokenizer_model()   (always a no-op; Primus uses HF repo IDs)
+  _needs_local_tokenizer()     (always False)
+  _read_last_node_log()
+'''
+
+import os
+import re
+import shlex
+import time
+
+from cvs.lib import globals
+from cvs.lib import linux_utils
+from cvs.lib.utils_lib import fail_test
+from cvs.lib.verify_lib import verify_dmesg_for_errors
+
+log = globals.log
+
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


cvs/lib/training/megatron/primus_lib.py
```diff
@@ -0,0 +1,556 @@
...
+
+        batch_args = (
+            f'--micro_batch_size {self.micro_batch_size} '
+            f'--global_batch_size {self.global_batch_size} '
+            f'--train_iters {self.iterations}'
+        )
+        if self.checkpoint_dir:
+            batch_args += f' --save {self.checkpoint_dir} --save_interval {self.save_interval or self.iterations}'
+        if self.load_checkpoint and self.checkpoint_dir:
+            batch_args += f' --load {self.checkpoint_dir}/checkpoints'
```
solaiys, line 402: **Blocking — the save target and the load/verify target are different directories, so the save phase can never validate.**

These two lines disagree by one path component:

```python
# :400  save phase
batch_args += f' --save {self.checkpoint_dir} --save_interval {...}'
# :402  load phase
batch_args += f' --load {self.checkpoint_dir}/checkpoints'
```

Rendering them with the shipped single-node defaults (`log_dir=/home/tester/LOGS`, so `ckpt_dir=$log_dir/ckpt_primus`):

```
--save  /home/tester/LOGS/ckpt_primus
--load  /home/tester/LOGS/ckpt_primus/checkpoints
test requires  /home/tester/LOGS/ckpt_primus/checkpoints/latest_checkpointed_iteration.txt
```

`megatron_single.py:271` and `megatron_distributed.py:276` both gate on that `checkpoints/latest_checkpointed_iteration.txt` file, and `--load` reads the same nested dir — but `--save` writes one level up. Whichever way Primus resolves this, one of the two is wrong:

- If the explicit `--save` wins (it is passed after `--`, so it lands in the trainer's own argv and would override any `--save` Primus derives from `PRIMUS_WORKSPACE`), the checkpoint is written to `{ckpt_dir}/` and the existence check at `{ckpt_dir}/checkpoints/...` fails — `test_checkpoint` fails at the save-verification gate every time, before resume is ever attempted.
- If `PRIMUS_WORKSPACE` (`:388`) wins and the checkpoint really does land at `{ckpt_dir}/checkpoints` as the comment on `:383-385` says, then the `--save {self.checkpoint_dir}` on `:400` is at best dead and at worst actively fighting the workspace mechanism.

Please make the save target and the load/verify target the same path — either drop `--save` and let `PRIMUS_WORKSPACE` own the location, or pass `--save {self.checkpoint_dir}/checkpoints` so all three agree.

Worth noting why this is easy to miss: the only hardware checkbox in the PR description ("Single-node: run `test_checkpoint` with `enforce=true` on a Primus image") is still unchecked, and there are no unit tests over the new checkpoint parsers, so nothing currently exercises this path. `checkpoint.enforce=false` in all three shipped configs means CI stays green regardless.

solaiys (commented):
## Review — 2 blocking issues

Reviewed against `main` (merge-base `dddd9143`, so the branch is current). No other reviewer has commented yet, so nothing here is a duplicate.

**Ruff, as requested — both clean.** Using the repo-pinned `ruff==0.14.8`:

| Check | Scope | Result |
| --- | --- | --- |
| `ruff format --check` | 7 changed `.py` files | `7 files already formatted` |
| `ruff format --check .` | repo-wide (`make fmt-check`) | `436 files already formatted` |
| `ruff check` | 7 changed `.py` files | `All checks passed!` |
| `ruff check . --unsafe-fixes` | repo-wide (`make lint`) | `All checks passed!` |

Unit suite is green (`Ran 1255 tests ... OK`), and the concurrency scan found nothing: no threads, `async`, or multiprocessing are introduced, `pytest-xdist` is not in play so `train_res_dict` has a single writer at a time, and the new code adds no shared mutable state.

The two blocking issues are inline. Both are in the new checkpoint path.

### Also verified as fine (so you know these were checked, not missed)

- **`_make_training_job` dispatch is interface-safe.** I diffed every member the two test files touch off the job object against both classes. Nothing the tests use is missing from `PrimusTrainingJob`, and the three `MegatronTrainingJob` methods it lacks (`exec_nic_setup_scripts`, `get_training_results_dict`, `scan_for_training_errors`) are only ever called as `self.<method>()` inside their own class, never on the job from the test files. `orch.container_config` is a plain dict (`container.py:102`), so `.get("image", "")` is safe.
- **Dropping `lifecycle.failed = True` in `test_training` is a fix, not a regression.** `README.md:75` already documents the intended behaviour — "A training failure is isolated to that combo's `test_training` row; other combos still run" — which the old code contradicted by halting every later combo.
- **The new `<changeme>` values are consistent with existing practice.** I loaded all three shipped configs through `load_training_variant`; every one of them already fails closed on `<changeme>` before this PR (`container.image` in the single-node configs, and eight fields in the 70B distributed config), so `checkpoint.checkpoint_dir` and the new volume entry follow the established fill-before-use pattern rather than breaking anything.

### Non-blocking, mentioning once and not gating on it

`save_iters % save_interval != 0` and `resume_iters > last_ckpt_step` are load-bearing invariants stated only in a comment (`megatron_single.py:256-257`). The shipped `21/5` satisfies both, but a config of `save_iters=20, save_interval=5` would make `last_ckpt_step` equal the final step, which `disable_last_saving=true` suppresses — surfacing as `FAILED step counter check` and reading like a product bug rather than a config mistake. A `model_validator` on `CheckpointConfig` would turn that into a load-time error.

cvs/lib/training/megatron/primus_lib.py
```diff
@@ -0,0 +1,556 @@
...
+                [
+                    f'nohup {self.scripts_dir}/distributed_wrapper_script_{i}.sh '
+                    f'>> {self.combo_log_dir}/out-node{i}/training.log 2>&1 &'
+                    for i in range(n)
+                ]
+            )
+        else:
+            self.orch.all.exec(
+                f'echo {shlex.quote(self.job_cmd)} > {self.scripts_dir}/single_node_wrapper_script.sh '
+                f'&& chmod 777 {self.scripts_dir}/single_node_wrapper_script.sh'
```
solaiys: **Blocking — this makes a file containing a live `HF_TOKEN` world-readable and world-writable.**

The script written here embeds the token in cleartext (`:374`):

```python
f'export HF_TOKEN="{self.hf_token}"; '
```

and is then created with mode `777` (also at `:426` for the distributed wrapper). The module this was cloned from does the same job with the opposite posture — `megatron_lib.py:665-666`:

```python
f'umask 077; echo {shlex.quote(self.job_cmd)} > {self.scripts_dir}/single_node_wrapper_script.sh '
f'&& chmod 700 {self.scripts_dir}/single_node_wrapper_script.sh'
```

and `megatron_lib.py:579` does the same for the distributed script. The `chmod 777` form is actually still present in `megatron_lib.py:645-650` **as commented-out code** — it is the approach that module deliberately moved away from, and this reintroduces it.

To be precise about current exposure: `__init__` does `sudo chmod 700` on `scripts_dir` (`:229`), so another unprivileged local user cannot traverse in and read it today. That makes this a hardened-pattern regression rather than an open leak. But the mitigation is a single directory bit away from failing, and `scripts_dir` is a volume mount shared into the container (per the comment at `megatron_lib.py:645`), so a container process running under a different UID is in scope. The file also does not need to be writable at all, let alone world-writable, to be `nohup`-executed.

Fix is to mirror the sibling implementation on both this line and `:426`: prefix with `umask 077` and use `chmod 700`.

solaiys (commented):
## Follow-up — checkpoint lifecycle: save once, don't save on resume, clean up after

Following up on the checkpoint path with three concrete requests. All of these are about the *lifecycle* of the checkpoint rather than the correctness assertions, and none of them change what `test_checkpoint` verifies.

I traced the flag construction for both phases with the shipped defaults (`save_interval=5`, `save_iters=21`, `resume_iters=25`):

| Phase | trainer args built | checkpoints written |
| --- | --- | --- |
| 1 — save | `--train_iters 21 --save {dir} --save_interval 5` | **4** — at iterations 5, 10, 15, 20 |
| 2 — resume | `--train_iters 25 --save {dir} --save_interval 25 --load {dir}/checkpoints` | **1 configured**, at iteration 25 (the final step) |

### 1. Save phase should write a single checkpoint

Four checkpoints are written but only the last (step 20) is ever used — `last_ckpt_step = floor(21/5)*5 = 20`. For `llama3.3_70B` that is four full model+optimizer checkpoint writes to shared NFS when one would do.

`save_interval=20, save_iters=21` gives exactly one checkpoint and is a **drop-in change — every assertion stays identical**:

```
                            current (5/21)   proposed (20/21)
checkpoints written         4                1
last_ckpt_step              20               20
expected first resume step  21               21
```

The general rule is `save_interval < save_iters < 2*save_interval`, so `save_iters = save_interval + 1` is the tightest form and preserves the existing "must not be an exact multiple" requirement noted at `megatron_single.py:256-257`.

One consequence to be aware of: `log_checkpoint_io_times` averages across checkpoints (`checkpoint_io.py:115-120`), so it becomes a single sample rather than a mean of four. That seems like the right trade for the disk and wall-clock saved, but flagging it since the averaging looks deliberate.

### 2. Resume phase should not save at all

Details inline. The resume phase still passes `--save` with `--save_interval` equal to `train_iters`, so a save is configured at the final iteration and is only avoided if the Primus YAML's `disable_last_saving=true` is honoured — a flag CVS never sets itself (it appears only in comments at `megatron_single.py:257` / `megatron_distributed.py:262`). The resume phase needs `--load` only; it has no reason to write.

### 3. Remove the checkpoint directory at the end of the test

Nothing deletes it today — `ckpt_dir` is only `mkdir -p`'d (`megatron_single.py:227`, `megatron_distributed.py:230`) and `test_teardown` only tears down containers (`megatron_single.py:594-603`). So every run leaves a full checkpoint tree behind, and with the four-checkpoint save phase that is four of them per run, accumulating across runs.

Inline comments on both test files with a suggested shape. **One safety note on the distributed side:** `ckpt_dir` there is the operator-supplied `checkpoint.checkpoint_dir` (a shared NFS root), so please scope the delete to the `checkpoints/` subdirectory Primus creates rather than `rm -rf`-ing the configured root, and skip the delete when the value is empty. The single-node path is CVS-derived (`{log_dir}/ckpt_primus`) and safe to remove wholesale.

cvs/lib/training/megatron/primus_lib.py
```diff
@@ -0,0 +1,556 @@
...
+                f'export PRIMUS_EXP_NAME=""; '
+            )
+
+        batch_args = (
+            f'--micro_batch_size {self.micro_batch_size} '
+            f'--global_batch_size {self.global_batch_size} '
+            f'--train_iters {self.iterations}'
+        )
+        if self.checkpoint_dir:
+            batch_args += f' --save {self.checkpoint_dir} --save_interval {self.save_interval or self.iterations}'
```
solaiys: **Please suppress checkpoint saving during the resume phase — it is unnecessary I/O.**

`--save` is emitted whenever `checkpoint_dir` is set, with no regard for whether this run is a resume:

```python
# :399-400
if self.checkpoint_dir:
    batch_args += f' --save {self.checkpoint_dir} --save_interval {self.save_interval or self.iterations}'
if self.load_checkpoint and self.checkpoint_dir:            # :401
    batch_args += f' --load {self.checkpoint_dir}/checkpoints'
```

Combined with `_run` defaulting `save_interval` to the iteration count when the caller omits it (`megatron_single.py:245` — `job.save_interval = save_interval or iters`), and the resume call omitting it (`megatron_single.py:289-294`, `megatron_distributed.py:294-299`), phase 2 ends up building:

```
--train_iters 25 --save {ckpt_dir} --save_interval 25 --load {ckpt_dir}/checkpoints
```

`save_interval == train_iters == 25`, so a save is configured at the final iteration. Whether it actually lands depends entirely on `disable_last_saving=true` in the Primus YAML, which CVS never sets — it only appears in the comments at `megatron_single.py:257` and `megatron_distributed.py:262`. So the resume phase is currently one external YAML assumption away from writing a full extra checkpoint of a 70B model to shared NFS on every run, and leaving a stale `iter_0000025` tree behind.

A resume run needs `--load` only, so gating the save flags is enough:

```python
if self.checkpoint_dir and not self.load_checkpoint:
    batch_args += f' --save {self.checkpoint_dir} --save_interval {self.save_interval or self.iterations}'
if self.load_checkpoint and self.checkpoint_dir:
    batch_args += f' --load {self.checkpoint_dir}/checkpoints'
```

Caveat worth resolving together with my other comment on `:402`: if `PRIMUS_WORKSPACE` (`:388`) independently causes Primus to derive its own `--save`, dropping the explicit flag may not be sufficient on its own. Both this and the save/load path question need the same answer about which mechanism owns the checkpoint location — so it would be worth confirming on hardware that no `saving checkpoint at iteration` line appears in the resume-phase log.

cvs/lib/training/megatron/utils/training_config_loader.py
```diff
@@ -130,6 +130,20 @@ class LossCurveConfig(_Forbid):
...
 
+class ConvergenceConfig(_Forbid):
+    target_metric: Literal["auto", "train_loss", "eval_loss"] = "auto"
+    target_value: float = 0.0
+
+
+class CheckpointConfig(_Forbid):
+    enforce: bool = False  # if False, test_checkpoint is skipped entirely
+    save_interval: int = 5  # checkpoint written every N steps
+    save_iters: int = 21  # save phase total; last checkpoint = floor(save/interval)*interval
```
solaiys, line 141: **Please default the save phase to a single checkpoint.**

`save_interval=5` with `save_iters=21` writes **four** checkpoints (iterations 5, 10, 15, 20), but only the last is used — `last_ckpt_step = floor(21/5)*5 = 20`. For the `llama3.3_70B` distributed variant that is four full model+optimizer writes to shared NFS per run, where one is sufficient to test resume.

`save_interval=20, save_iters=21` produces exactly one checkpoint and leaves every derived value untouched:

```
                            current (5/21)   proposed (20/21)
checkpoints written         4                1
last_ckpt_step              20               20
expected first resume step  21               21
resume_iters > last step    yes (25 > 20)    yes (25 > 20)
```

So it is a drop-in change — no assertion in `test_checkpoint` changes. The three shipped configs that gained a `checkpoint` block would want the same values.

The rule to keep in mind if these are ever retuned is `save_interval < save_iters < 2*save_interval`: the lower bound guarantees a checkpoint exists, and the upper bound keeps it from being the only one while also keeping `last_ckpt_step != save_iters`, which matters because `disable_last_saving=true` suppresses the final-iteration save. `save_iters = save_interval + 1` satisfies both for any interval.

Since these invariants are load-bearing and currently live only in a comment, a `model_validator` here would turn a bad combination into a clear config-load error instead of a confusing `FAILED step counter check` at runtime.

cvs/tests/training/megatron/megatron_single.py
```diff
@@ -178,6 +193,199 @@ def test_smoke(orch, variant_config, hf_token, lifecycle, request):
...
+    expected_first = last_ckpt_step + 1
+
+    # Phase 2: load — resume from last checkpoint, train to resume_iters total.
+    # Primus reads the checkpoint and starts iterating from last_ckpt_step+1.
+    try:
+        resume_log = _run(
+            "ckpt_resume",
+            ckpt_cfg.resume_iters,
+            checkpoint_dir=ckpt_dir,
+            load_checkpoint=True,
```
solaiys, line 294: **The resume phase still configures a save here, and the checkpoint directory is never cleaned up.**

Two requests on this call.

**(a) No saving on resume.** This call omits `save_interval`, so `_run` falls back to the iteration count at `:245` (`job.save_interval = save_interval or iters` → 25). With `primus_lib.py:399` emitting `--save` whenever `checkpoint_dir` is set, phase 2 builds `--save {ckpt_dir} --save_interval 25` alongside `--load`, configuring a save at the final iteration. See my comment on `primus_lib.py:400` for the fix — gating the save flags on `not self.load_checkpoint` fixes it for both files at once.

**(b) Remove the checkpoint directory when the test finishes.** `ckpt_dir` is created at `:227` and never deleted; `test_teardown` (`:594-603`) only tears down containers. Every run therefore leaves a full checkpoint tree on disk, and with the current four-checkpoint save phase that is four of them.

Here `ckpt_dir` is CVS-derived (`{log_dir}/ckpt_primus` at `:226`), so it is safe to remove wholesale. Keeping it on failure preserves the `ls -laR` diagnostic at `:275`:

```python
finally:
    if not request.node.session.testsfailed:
        orch.exec(f"rm -rf {shlex.quote(ckpt_dir)}")
    else:
        log.info("checkpoint dir retained for debugging: %s", ckpt_dir)
```

Any equivalent shape is fine — a yield-fixture would work too. The important part is that a passing run does not leave a model checkpoint behind.

cvs/tests/training/megatron/megatron_distributed.py
```diff
@@ -178,6 +194,203 @@ def test_smoke(orch, variant_config, hf_token, lifecycle, request):
...
+    expected_first = last_ckpt_step + 1
+
+    # Phase 2: load — resume from last checkpoint, train to resume_iters total.
+    # Primus reads the checkpoint and starts iterating from last_ckpt_step+1.
+    try:
+        resume_log, resume_log_node0 = _run(
+            "ckpt_resume",
+            ckpt_cfg.resume_iters,
+            checkpoint_dir=ckpt_dir,
+            load_checkpoint=True,
```
solaiys, line 299: **Same two points as the single-node file, with one extra safety consideration.**

**(a) No saving on resume.** As in the single-node file, this call omits `save_interval`, so `_run:250` defaults it to `iters` (25) and `primus_lib.py:399` emits `--save {ckpt_dir} --save_interval 25` next to `--load`. This matters more here than single-node: it is a full `llama3.3_70B` model+optimizer write to a shared NFS path at the end of every resume phase. Fix is on `primus_lib.py:400`.

**(b) Remove the checkpoint directory when the test finishes** — but scope the delete carefully. Unlike single-node, `ckpt_dir` here is the operator-supplied `checkpoint.checkpoint_dir` (`:224`), which the config documents as a shared NFS root and which the 70B config ships as `<changeme>`. Please do **not** `rm -rf` the configured root: an operator could reasonably point it at a directory holding other artifacts, and an empty or `/`-ish value would be destructive.

Scoping the removal to the subdirectory Primus actually creates keeps it safe:

```python
finally:
    ckpt_subdir = f"{ckpt_dir.rstrip('/')}/checkpoints"
    if ckpt_dir and ckpt_dir.strip("/") and not request.node.session.testsfailed:
        orch.exec(f"rm -rf {shlex.quote(ckpt_subdir)}")
    else:
        log.info("checkpoint dir retained for debugging: %s", ckpt_dir)
```

The `ckpt_dir.strip("/")` guard is the same one `anc_lib` uses for its relocatable install prefix, and it is worth having on any config-supplied path that reaches `rm -rf`.

cvs/tests/training/megatron/megatron_single.py
solaiys: Small note on the snippets I posted here and on `megatron_distributed.py:298`: both use `shlex.quote`, but neither test file imports `shlex` today (`megatron_single.py:21-24` has `json`, `os`, `re`, `time`). So an `import shlex` goes with them — or drop the quoting on the single-node path, where `ckpt_dir` is CVS-derived from `log_dir` rather than operator-supplied.

Worth keeping it on the distributed path though, since `ckpt_dir` there comes straight from `checkpoint.checkpoint_dir` in the config.

cvs/lib/training/megatron/primus_lib.py
sukesh-amd: Fixed

cvs/lib/training/megatron/utils/training_config_loader.py
sukesh-amd: Fixed

solaiys (approved):
## Approving @ `220b8599`

The follow-up commit addresses the review. Verified against the new head; CI (`test`, `container-smoke`) is green on this SHA.

| Ask | Status |
| --- | --- |
| Save and load/verify must be the same directory | Addressed — `--save {dir}` was dropped so `PRIMUS_WORKSPACE` owns the location. Save, load (`--load {dir}/checkpoints`), and the existence check (`{dir}/checkpoints/latest_checkpointed_iteration.txt`) now agree. |
| `chmod 777` on the wrapper that contains `HF_TOKEN` | Fixed — `umask 077` + `chmod 700`, matching `megatron_lib.py`. |
| Save phase should write one checkpoint | Fixed — `save_interval` default and shipped configs are `20` with `save_iters=21`, so one save at step 20. |
| No checkpoint write during resume | Fixed — `--save_interval` is only emitted when `not self.load_checkpoint`. |
| Remove checkpoint files at the end of the test | Fixed — single-node deletes `ckpt_dir`; distributed deletes only `{checkpoint_dir}/checkpoints` and skips filesystem root. Both keep the tree if the test failed, for debugging. |

One leftover that is **not** blocking: resume still exports `PRIMUS_WORKSPACE`. If Primus ever derives a save from that env alone, a write could still happen on load. The CLI flags no longer request a save, which is what we asked for. Worth a glance in the lab log (`saving checkpoint at iteration`) the first time `enforce=true` is run; not a reason to hold the merge.

PR #362, opened 2026-08-24
Config update with RDMA lib mounts + fixes

solaiys:
config(JAX) Fix bnxt_re RDMA lib mounts + enable IB across jaxmaxtext configs

    1. RDMA lib volume mount fix in all distributed jaxmaxtext configs
       (mi300x/mi325x/mi355x x llama-3.1-8b / llama-3.3-70b / llama-3.1-405b /
       deepseek-v2-lite): mount the host libbnxt_re-rdmav34.so directly onto the
       container's real .so path read-only (":ro") instead of the ".so.host"
       indirection, and add a read-only /usr/lib/x86_64-linux-gnu/libibverbs.so.1
       mount.
    2. Flip NCCL_IB_DISABLE "1" -> "0" in those configs to enable IB/RoCE over
       bnxt_re (was pinned to the TCP fallback).

    Signed-off-by: Saravanan Solaiyappan <saravanan.solaiyappan@amd.com>

    fix(JAX) Harden jaxmaxtext poller/error-detection + add opt-out smoke test

    1. Broaden NaN/Inf detection to match loss/lm_loss/perplexity going NaN (not
       just the throughput fields, which stay numeric) and MaxText's explicit
       "Aborting training due to NaN loss" line; ignore the false-positive-prone
       TensorBoard "NaN or Inf found in input tensor" warning.
    2. Stream each poll's new node-0 log lines BEFORE scanning, and log the full
       offending chunk before raising, so failure tracebacks reach the console /
       --log-file (previously the scan raised first and only the truncated
       text[-500:] survived).
    3. Add an opt-OUT smoke test config block (training.smoke, enabled by default,
       with steps/per_device_batch_size/max_target_length); smoke() skips when
       smoke.enabled=false, for iterative experiments.
    4. Log every test's final PASS/FAIL/SKIP (with skip reason) to the console /
       --log-file, mirroring the "Starting Testcase:" line.
    5. Drop the per-row "Metric Results" HTML link for SKIPPED metric tests, so it
       no longer points at an empty/missing metric_results.html when an earlier
       stage (e.g. smoke) fails.
    6. Unit tests for the NaN-loss/abort signatures, the failure-chunk logging, and
       the smoke schema defaults.

    Signed-off-by: Saravanan Solaiyappan <saravanan.solaiyappan@amd.com>

### DIFF
--- a/cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
+++ b/cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
@@ -56,7 +56,16 @@ _TRAINING_ERR_PATTERNS = {
     'segfault': r'Segmentation fault|SIGSEGV|signal 11|core dumped',
 }
 
-_NAN_INF_RE = re.compile(r'(TFLOP/s/device|Tokens/s/device):\s*(NaN|Inf|-Inf)', re.I)
+# NaN/Inf in any reported training metric -- loss/lm_loss/perplexity go NaN while
+# the throughput fields (TFLOP/s/device, Tokens/s/device) stay numeric, so those
+# alone are NOT enough; MaxText also emits an explicit abort line. NOTE: the
+# TensorBoard "NaN or Inf found in input tensor" writer warning is intentionally
+# NOT matched -- it also fires on healthy steps (false positive).
+_NAN_INF_RE = re.compile(
+    r'(?:TFLOP/s/device|Tokens/s/device|loss|lm_loss|perplexity)\s*[:=]\s*(?:nan|[+-]?inf)\b'
+    r'|Aborting training due to NaN',
+    re.I,
+)
 
 
 def _sanitize(name):
@@ -273,7 +282,16 @@ class MaxTextTrainingJob:
         if not rdma.container_mount_file or not rdma.container_dest_file:
             log.info("rdma_lib paths not configured, skipping")
             return
-        cmd = f"sudo cp {shlex.quote(rdma.container_mount_file)} {shlex.quote(rdma.container_dest_file)}"
+        # Only copy when the mount source actually exists. Newer configs mount the
+        # host RDMA lib directly onto the container's real .so path (read-only), so
+        # the legacy ".host" copy source is absent -- skip the cp instead of
+        # emitting a spurious error (the ibv_devinfo check below is the real gate).
+        src = shlex.quote(rdma.container_mount_file)
+        dst = shlex.quote(rdma.container_dest_file)
+        cmd = (
+            f"if [ -f {src} ]; then sudo cp {src} {dst}; "
+            f"else echo 'rdma_lib: source {rdma.container_mount_file} not present, skipping cp'; fi"
+        )
         out = self.orch.exec(cmd)
         for host, output in (out or {}).items():
             log.info("[rdma_lib %s] %s", host, (output or "")[:200])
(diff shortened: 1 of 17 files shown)


cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
amd-droy: The updated jaxmaxtext configs now bind-mount the host .so directly onto …/libbnxt_re-rdmav34.so:ro (see config volumes ~L36–37), but rdma_lib.container_mount_file still points at the old .so.host path. This sudo cp container_mount_file → container_dest_file will likely fail: .so.host is no longer mounted, and the destination is read-only. Consider no-op'ing here when the library is already bind-mounted, or align the configs' rdma_lib block with the direct-mount approach.

solaiys: THis setup_rdma_lib is a workaround step test for the (Broadcom/Thor2 NIC workaround ).
originally, the the lib file was copied in 2 stage copy. ( mount + copy internally in container)

But direct mount(with readonly) also works fine. In Any container image, if its throwing error in mounting on the existing file name, then we need to fallback to the 2 stage copy. 
This is my understanding.

So I am keeping the legacy steps, but the sudo cp command will not find the file. But the test itself gated for PASS/FAIL based on the next command ibv_devinfo output.

So that sudo cp is a harmless copy command incase the the mount is directly to the dest file.


I have made the change to skip the copy if the file is not exists.
then the test output is as follows:

```
2026-08-24 18:37:34 INFO Starting Testcase: setup RDMA library
2026-08-24 18:37:34 INFO cmd = sudo -n docker exec rocm-jaxmaxtext-mi325x-deepseek-v4-284b bash -c 'if [ -f /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so.host ]; then sudo cp /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav3
4.so.host /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so; else echo '"'"'rdma_lib: source /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so.host not present, skipping cp'"'"'; fi'
2026-08-24 18:37:34 INFO #----------------------------------------------------------#
2026-08-24 18:37:34 INFO Host == 10.32.80.124 ==
2026-08-24 18:37:34 INFO #----------------------------------------------------------#
2026-08-24 18:37:34 INFO rdma_lib: source /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so.host not present, skipping cp
2026-08-24 18:37:34 INFO #----------------------------------------------------------#
2026-08-24 18:37:34 INFO Host == 10.32.80.125 ==
2026-08-24 18:37:34 INFO #----------------------------------------------------------#
2026-08-24 18:37:34 INFO rdma_lib: source /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so.host not present, skipping cp
2026-08-24 18:37:34 INFO #----------------------------------------------------------#
2026-08-24 18:37:34 INFO Host == 10.32.81.141 ==
2026-08-24 18:37:34 INFO #----------------------------------------------------------#
2026-08-24 18:37:34 INFO rdma_lib: source /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so.host not present, skipping cp
2026-08-24 18:37:34 INFO #----------------------------------------------------------#
2026-08-24 18:37:34 INFO Host == 10.32.80.126 ==
2026-08-24 18:37:34 INFO #----------------------------------------------------------#
2026-08-24 18:37:34 INFO rdma_lib: source /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so.host not present, skipping cp


```

amd-droy (approved):
lgtm. thank you.

PR #364, opened 2026-08-25
Ichristo/parallel lib transport support

cijohnson:
## Submission Checklist

### DIFF
--- a/cvs/lib/parallel/__init__.py
+++ b/cvs/lib/parallel/__init__.py
@@ -9,9 +9,12 @@ All code contained here is Property of Advanced Micro Devices, Inc.
 from cvs.lib.parallel.multiprocess_phandle import MultiProcessParallelHandle
 from cvs.lib.parallel.phandle import ParallelHandle
 from cvs.lib.parallel.scp import scp
+from cvs.lib.parallel.transport import BaseTransport, create_transport
 
 __all__ = [
+    'BaseTransport',
     'ParallelHandle',
     'MultiProcessParallelHandle',
+    'create_transport',
     'scp',
 ]
--- a/cvs/lib/parallel/__init__.py
+++ b/cvs/lib/parallel/__init__.py
@@ -6,17 +6,12 @@ All code contained here is Property of Advanced Micro Devices, Inc.
 '''
 
 # Main exports for the parallel module
-from cvs.lib.parallel.multiprocess_pssh import MultiProcessPssh
-from cvs.lib.parallel.pssh import Pssh
+from cvs.lib.parallel.multiprocess_phandle import MultiProcessParallelHandle
+from cvs.lib.parallel.phandle import ParallelHandle
 from cvs.lib.parallel.scp import scp
 
-# Backward compatibility aliases
-PsshSharded = MultiProcessPssh  # Old name for multiprocess SSH
-
 __all__ = [
-    'Pssh',
-    'MultiProcessPssh',
+    'ParallelHandle',
+    'MultiProcessParallelHandle',
     'scp',
-    # Backward compatibility
-    'PsshSharded',
 ]
--- a/cvs/lib/parallel/phandle.py
+++ b/cvs/lib/parallel/phandle.py
@@ -9,14 +9,17 @@ from __future__ import print_function
 
 import warnings
 from gevent import Timeout as GTimeout
-from gevent import killall
 from pssh.clients import ParallelSSHClient
 from pssh.exceptions import Timeout, ConnectionError, SessionError
 
 from cvs.lib.env_lib import build_env_prefix
+from cvs.lib.parallel.transport import create_transport
 from cvs.lib import globals
 
 global_log = globals.log
+
+# Re-export for unit tests that patch ParallelSSHClient on this module.
+__all__ = ['ParallelHandle', 'ParallelSSHClient', '_select_reachable_hosts']
 
 
 def _select_reachable_hosts(reachable_hosts, hosts):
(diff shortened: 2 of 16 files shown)


speriaswamy-amd (commented):
Wouldn't this fail ? if transport is HTTP ?

cvs/lib/parallel/phandle.py
```diff
@@ -27,15 +30,15 @@ def _select_reachable_hosts(reachable_hosts, hosts):
     requested = list(hosts)
     unknown_hosts = [host for host in requested if host not in reachable]
     if unknown_hosts:
-        raise ValueError(f"SFTP download requested unreachable host(s): {unknown_hosts}")
+        raise ValueError(f"Download requested unreachable host(s): {unknown_hosts}")
     return [host for host in reachable if host in requested]
 
 
-class Pssh:
+class ParallelHandle:
```
speriaswamy-amd: Handle is talking to wire client - 

```python3
 # cvs/lib/parallel/phandle.py lines 325-330

          if inactivity_timeout is not None:
              output = self.client.run_command(full_cmd, stop_on_errors=self.stop_on_errors)
          elif timeout is None:
              output = self.client.run_command(full_cmd, stop_on_errors=self.stop_on_errors)
          else:
              output = self.client.run_command(full_cmd, read_timeout=timeout, stop_on_errors=self.stop_on_errors)

   # cvs/lib/parallel/phandle.py lines 413-415

          self.log.info('SFTP upload %s -> %s on %s', local_file, remote_file, self.reachable_hosts)
          cmds = self.client.copy_file(local_file, remote_file, recurse=recurse)
          self.client.pool.join()
```

This would fail as ParallelHTTP doesn't implement upload/download API's

cvs/lib/parallel/ssh_transport.py
```diff
@@ -0,0 +1,110 @@
...
+
+from gevent import killall
+
+from cvs.lib import globals
+from cvs.lib.parallel.transport import BaseTransport
+
+log = globals.log
+
+
+def _get_parallel_ssh_client():
```
speriaswamy-amd, line 16: Wouldn't this cause circular import issue, since phandle.py imports transport ?

cvs/lib/parallel/__init__.py
```diff
@@ -6,17 +6,15 @@
 '''
 
 # Main exports for the parallel module
-from cvs.lib.parallel.multiprocess_pssh import MultiProcessPssh
-from cvs.lib.parallel.pssh import Pssh
+from cvs.lib.parallel.multiprocess_phandle import MultiProcessParallelHandle
+from cvs.lib.parallel.phandle import ParallelHandle
```
speriaswamy-amd, line 10: Couple of tests such as `test_docker.py` (unit test) and `check_cluster_health.py` in cvs-mon still uses parallel_ssh_lib.Pssh

cvs/lib/parallel/ssh_transport.py
```diff
@@ -0,0 +1,110 @@
...
+            )
+        return ParallelSSHClient(
+            hosts,
+            user=self.user,
+            password=self.password,
+            keepalive_seconds=30,
+            **self.ssh_client_kwargs,
+        )
+
+    def rebuild(self, hosts):
```
speriaswamy-amd, line 48: Suggestion from claude:

`destroy()` exists because dropping the client reference does not close timed-out SSH sessions (sshd MaxSessions; see the subset-handle leak fix). `rebuild()`, connectivity probes, and subset downloads still replace/create clients without teardown:
```python
   # cvs/lib/parallel/ssh_transport.py lines 48-50

      def rebuild(self, hosts):
          self.hosts = list(hosts)
          self.client = self._make_client(self.hosts)

   #cvs/lib/parallel/ssh_transport.py lines 52-56

      def check_connectivity(self, hosts):
          ...
          temp_client = self._make_client_with_kwargs(hosts, temp_ssh_client_kwargs)
          output = temp_client.run_command('echo 1', stop_on_errors=False, read_timeout=2)
          return [item.host for item in output if item.exception]

   #cvs/lib/parallel/phandle.py lines 462-465

          if target_hosts == self.reachable_hosts:
              client = self.client
          else:
              client = self._transport.client_for_hosts(target_hosts).client
```
  Call `destroy()` (or equivalent) on the old/temp client. Unify _make_client / _make_client_with_kwargs while you are there.

speriaswamy-amd (changes requested):

cvs/lib/parallel/__init__.py
cijohnson: yes they import "parallel_ssh_lib.Pssh"

cvs/lib/parallel_ssh_lib.py already does this

from cvs.lib.parallel.multiprocess_phandle import MultiProcessParallelHandle as Pssh  # Main interface (auto-sharding)

cvs/lib/parallel/ssh_transport.py
cijohnson: no, it is invoked here

if transport == 'ssh':
    from cvs.lib.parallel.ssh_transport import SshTransport

cijohnson:
> Wouldn't this fail ? if transport is HTTP ?

The handle API is being made transport-agnostic; HTTP is a stub. A fair reply is: HTTP is not implemented yet; 'http' fails loudly; SSH remains the only working path.

cvs/lib/parallel/ssh_transport.py
cijohnson: this is a preexisting bug, not a new bug introduced in this PR, will raise a sperate ticket to take care of it

speriaswamy-amd (approved):

cvs/lib/parallel/ssh_transport.py
speriaswamy-amd: This work item is going to be tracked in AIMVT-318

PR #366, opened 2026-08-25
Added the test count for nodesmoke tier 1,2,3 tests and updated the document as well

urtiwari:
Added the test count for nodesmoke tier 1,2,3 tests and updated the document as well

### DIFF
--- a/cvs/lib/preflight/node_smoke.py
+++ b/cvs/lib/preflight/node_smoke.py
@@ -16,6 +16,14 @@ import shlex
 from typing import Any, Dict, List, Optional
 
 from cvs.lib.preflight.base import PreflightCheck
+from cvs.lib.preflight.node_smoke_counts import aggregate_node_smoke_test_counts
+
+NODE_SMOKE_TIER1_SECTION = "node_smoke_tier1"
+LEGACY_NODE_SMOKE_SECTION = "node_smoke"
+
+NODE_SMOKE_TIER1_LABEL = "Node Smoke Tier 1"
+NODE_SMOKE_TIER2_LABEL = "Node Smoke Tier 2"
+NODE_SMOKE_TIER3_LABEL = "Node Smoke Tier 3"
 
 # Writable fallback when reporting.artifacts_root_dir is omitted from config.
 # Do not use ``{user-id}`` here — in-code defaults bypass resolve_test_config_placeholders.
@@ -42,6 +50,14 @@ def get_nested_config(config_dict, section, key, default):
     if isinstance(current, dict) and key in current:
         return current[key]
     return default
+
+
+def get_preflight_nested(cfg: dict, section: str, legacy_section: str, key: str, default):
+    """Read ``section.key``, falling back to ``legacy_section.key`` when unset."""
+    value = get_nested_config(cfg, section, key, None)
+    if value is None and legacy_section:
+        value = get_nested_config(cfg, legacy_section, key, None)
+    return default if value is None else value
 
 
 def resolve_rdma_interfaces(cfg: dict) -> List[str]:
@@ -94,7 +110,9 @@ def _resolve_dump_path(cfg: dict) -> str:
     """Return node_smoke dump directory; empty config uses reporting artifacts root."""
     artifacts_root = get_nested_config(cfg, "reporting", "artifacts_root_dir", DEFAULT_ARTIFACTS_ROOT_DIR)
     default_dump = f"{str(artifacts_root).rstrip('/')}/node_smoke"
-    configured = get_nested_config(cfg, "node_smoke", "dump_path", default_dump)
+    configured = get_preflight_nested(
+        cfg, NODE_SMOKE_TIER1_SECTION, LEGACY_NODE_SMOKE_SECTION, "dump_path", default_dump
+    )
     configured_s = str(configured or "").strip()
     return configured_s if configured_s else default_dump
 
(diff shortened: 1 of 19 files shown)


cvs/lib/preflight/node_smoke.py
```diff
@@ -286,72 +304,77 @@ def __init__(self, phdl, node_list: List[str], config_dict=None):
...
-        self.mode = _normalize_mode(get_nested_config(cfg, "node_smoke", "connectivity_mode", "skip"))
-        self.primus_dir = get_nested_config(cfg, "node_smoke", "primus_dir", "")
-        self.venv_activate = get_nested_config(cfg, "node_smoke", "venv_activate", "")
-        self.gpus_per_node = int(get_nested_config(cfg, "node_smoke", "gpus_per_node", 8))
-        self.master_port = int(get_nested_config(cfg, "node_smoke", "master_port", 1234))
-        self.ssh_timeout = int(get_nested_config(cfg, "node_smoke", "ssh_timeout", 300))
+        tier1 = NODE_SMOKE_TIER1_SECTION
+        legacy = LEGACY_NODE_SMOKE_SECTION
+        self.mode = _normalize_mode(get_preflight_nested(cfg, tier1, legacy, "connectivity_mode", "skip"))
+        self.primus_dir = get_preflight_nested(cfg, tier1, legacy, "primus_dir", "")
```
solaiys, line 310: **Blocking: `auto_setup` still only reads `node_smoke`, so the documented `node_smoke_tier1` config never clones Primus.**

This line now pulls `primus_dir` from `node_smoke_tier1` (then `node_smoke`). A few lines later, `run()` still does:

```python
setup = PrimusSetup(self.phdl, hosts, self.config_dict)
```

`PrimusSetup` defaults to `config_section="node_smoke"` with no fallback. It does not use `get_preflight_nested`.

What that means:

- Shipped `preflight_config.json` still has `node_smoke`, so today's file happens to work for Tier 1.
- The sample in `docs/reference/configuration-files/preflight.rst` only has `node_smoke_tier1`. The check loads `primus_dir` fine, then `auto_setup` fails with `node_smoke.primus_dir is required for auto_setup`.
- The docs say legacy keys are renamed to `node_smoke_tier1` at load time. `config_dict` only runs `normalize_legacy_preflight_rdma_config` — there is no rename. If you add that rename later without fixing this call, the shipped config will break the same way.

**Fix:** construct setup with the same sections the check already uses:

```python
setup = PrimusSetup(
    self.phdl,
    hosts,
    self.config_dict,
    config_section=NODE_SMOKE_TIER1_SECTION,
    fallback_section=LEGACY_NODE_SMOKE_SECTION,
)
```

Please add a unit test that `auto_setup=True` with only `node_smoke_tier1` set still sees `primus_dir` (mock SSH so you don't actually clone).

cvs/lib/preflight/tier3_info.py
```diff
@@ -320,15 +338,15 @@ def run(self) -> Dict[str, Any]:
                 hosts,
                 self.config_dict,
                 config_section=self.CONFIG_SECTION,
-                fallback_section="node_smoke",
+                fallback_section=NODE_SMOKE_TIER1_SECTION,
```
solaiys: **Blocking: this fallback change breaks Primus auto_setup for the shipped config.**

Previously this was `fallback_section="node_smoke"`. Now it is `node_smoke_tier1`. `PrimusSetup` only looks at `config_section` then that one fallback — it does not understand `tier3_info` or `node_smoke`.

Shipped `cvs/input/config_file/preflight/preflight_config.json` still has `tier3_info` + `node_smoke` (no `node_smoke_tier3` / `node_smoke_tier1`). With `connectivity_mode: run` and `auto_setup: true`:

1. `resolve_tier3_setting` still finds `primus_dir` from `node_smoke` (legacy helper). The check thinks it is ready to run.
2. `PrimusSetup` looks at `node_smoke_tier3`, then `node_smoke_tier1`. Both missing.
3. Setup returns FAIL: `node_smoke_tier3.primus_dir is required for auto_setup`.
4. Tier 3 is skipped even though Primus paths are in the config.

**Fix:** use the same chain as `resolve_tier3_setting` for Primus paths. Either:

- teach `PrimusSetup` multiple fallbacks: `node_smoke_tier3` → `tier3_info` → `node_smoke_tier1` → `node_smoke`, or
- actually rename the legacy keys at load time (as the docs claim), then point setup at the new names.

Please add a test with the shipped shape (`tier3_info` + `node_smoke`, empty `primus_dir` under Tier 3) and assert `PrimusSetup` still resolves `primus_dir`.

cvs/tests/preflight/preflight_checks.py
```diff
@@ -727,48 +734,55 @@ def test_node_smoke(phdl, config_dict):
...
         )
     else:
-        log.info("Primus node_smoke PASS on %d/%d nodes", passing, total)
+        log.info("%s PASS on %d/%d nodes", NODE_SMOKE_TIER1_LABEL, passing, total)
 
     preflight_update_test_result()
 
 
-def test_tier3_info(phdl, config_dict):
+test_node_smoke = test_node_smoke_tier1
```
solaiys: **Blocking: this alias is a second pytest test. `cvs run preflight_checks` will run Node Smoke twice.**

Pytest collects every module-level `test_*` name. `test_node_smoke = test_node_smoke_tier1` (and `test_tier3_info = test_node_smoke_tier3` below) therefore shows up as extra items.

`cvs run preflight_checks` with no function name passes the whole file to pytest (`cli_plugins/run_plugin.py`). That is also the command in the preflight docs. Collection order is:

1. `test_node_smoke_tier1`
2. `test_node_smoke` — same function, runs again
3. `test_node_smoke_tier3`
4. `test_tier3_info` — same function, runs again

The second pass hits the same `dump_path` / Primus install while the first may still be writing. Setting `__test__ = False` on the alias also disables the original, because they are the same function object.

**Fix:** keep `cvs run preflight_checks test_node_smoke` working without collecting the alias when the whole module runs. For example in a preflight `conftest.py`:

```python
_ALIASES = {
    "test_node_smoke": "test_node_smoke_tier1",
    "test_tier3_info": "test_node_smoke_tier3",
}

def pytest_collection_modifyitems(items):
    names = {item.name for item in items}
    drop = {alias for alias, canon in _ALIASES.items() if alias in names and canon in names}
    items[:] = [item for item in items if item.name not in drop]
```

Targeted `cvs run preflight_checks test_node_smoke` still works (only the alias is requested). Please `--collect-only` the module and confirm you get one Node Smoke Tier 1 item and one Tier 3 item.

solaiys (commented):
## Review: blocking issues

The test-count reporting itself looks fine. Three things will break real runs before that suffix matters.

### Primus auto_setup still only reads `node_smoke`

`NodeSmokeCheck` now loads `node_smoke_tier1` (with a `node_smoke` fallback), but `PrimusSetup(...)` is still constructed with the default section `node_smoke` and no fallback. The sample in `docs/reference/configuration-files/preflight.rst` only has `node_smoke_tier1`, so `auto_setup` fails even though the check already has `primus_dir`. The docs also say legacy keys are renamed at load time — `config_dict` does not do that.

### Tier 3 auto_setup dropped the working `node_smoke` fallback

`fallback_section` changed from `node_smoke` to `node_smoke_tier1`. Shipped `preflight_config.json` still uses `tier3_info` + `node_smoke`. The check itself still finds those paths via `get_preflight_nested`, then `PrimusSetup` looks at `node_smoke_tier3` then `node_smoke_tier1` (both missing) and fails with `node_smoke_tier3.primus_dir is required for auto_setup`.

### `test_node_smoke = test_node_smoke_tier1` is collected as a second pytest test

Same for `test_tier3_info = test_node_smoke_tier3`. `cvs run preflight_checks` (no function name) collects the whole module, so Node Smoke Tier 1 and Tier 3 each run twice — the second pass races the same `dump_path`.

solaiys (approved):
Re-reviewed `f4476cb0` (Addressed the review comments). The three blockers from the previous pass are fixed:

- **Tier 1 auto_setup** now constructs `PrimusSetup` with `node_smoke_tier1` and a `node_smoke` fallback, and `PrimusSetup` walks a section chain (with tests for canonical-only and legacy-only configs).
- **Tier 3 auto_setup** uses `setting_resolver=resolve_tier3_setting`, so empty `tier3_info.primus_dir` still inherits from `node_smoke`. Load-time `normalize_legacy_preflight_node_smoke_sections` copies the shipped keys as the docs claimed.
- **Double collection** — the `test_node_smoke` / `test_tier3_info` module aliases are gone. `cvs run preflight_checks test_node_smoke` is mapped in `run_plugin.py` so a full-module run no longer executes Node Smoke twice.

No remaining blocking issues. Approved.

PR #365, opened 2026-08-25
AIMVT-302: Implement HTTP Agent to serve exec, exec_cmd_list etc

speriaswamy-amd:
## Summary
- Implements the per-node HTTP agent (`cvs/core/agent/http_agent.py`) using FastAPI/asyncio, with rank registration, an `/v1/exec` endpoint supporting inline and file-backed output modes, and a shutdown endpoint.
- Adds `ProcessRegistry`/`AgentRegistry` to track spawned processes and registered agents per host, with process-group based termination.
- Adds `timeout` and `inactivity_timeout` handling for exec requests, killing the process group when either fires.
- Updates `cvs/core/agent/messages.py` request/response schemas to support the above.

Jira: [AIMVT-302](https://amd-hub.atlassian.net/browse/AIMVT-302)

## Test plan
- [x] `cvs/core/agent/unittests/test_http_agent.py` covers registration, exec (inline/file modes), timeout/inactivity-timeout, and shutdown paths
- [x] `cvs/core/agent/unittests/test_messages.py` updated for schema changes

PR #368, opened 2026-08-26
AIMVT-303: Implement HTTP client in ParallelHTTPClient class

speriaswamy-amd:
## Summary
- Adds `ParallelHTTPClient` (`cvs/core/agent/http_client.py`), an async httpx-based client providing API parity with `ParallelSSHClient`: `run_command`, `health`, `shutdown`, lazy shared connection reuse, and `rebuild`/`destroy` for host-set changes.
- Adds `HostOutput` dataclass and exception mapping (`HTTPConnectionError`/`HTTPProtocolError`) so unreachable vs. failed-request hosts can be distinguished the same way `prune_unreachable_hosts` distinguishes pssh exceptions.
- Supports `stop_on_errors`, `host_args` (per-host command substitution), and inline/file output modes, collecting file-mode output back into `list[str]` for a uniform response shape.

Stacked on #365 (AIMVT-302) — this PR's diff is limited to `http_client.py` and its tests; base branch will be retargeted to `main` once #365 merges.

Jira: [AIMVT-303](https://amd-hub.atlassian.net/browse/AIMVT-303)

## Test plan
- [x] `cvs/core/agent/unittests/test_http_client.py` covers `run_command` (inline/file/exit-code-only modes, `host_args`, `stop_on_errors`), `health`, `shutdown`, and exception classification, all against mocked HTTP transport

### DIFF
--- a/cvs/core/agent/http_client.py
+++ b/cvs/core/agent/http_client.py
@@ -8,7 +8,6 @@ All code contained here is Property of Advanced Micro Devices, Inc.
 import asyncio
 import uuid
 from dataclasses import dataclass
-from pathlib import Path
 
 import httpx
 
@@ -104,8 +103,19 @@ class ParallelHTTPClient:
             self._client = None
 
     def _build_exec_requests(
-        self, cmd: str, host_args: list[str] | None, read_timeout: float | None
+        self,
+        cmd: str,
+        host_args: list[str] | None,
+        read_timeout: float | None,
+        env: dict[str, str] | None,
+        inactivity_timeout: float | None,
+        output_mode: messages.ExecOutputMode,
     ) -> dict[str, messages.ExecRequest]:
+        # Imported here, not at module level: cvs.core.run_layout pulls in cvs/core/__init__.py's
+        # orchestrator factory, which reaches back into cvs/core/agent/ in ways that risk a cycle
+        # (same reasoning as cvs/lib/utils_lib.py's lazy import of RunLayout).
+        from cvs.core.run_layout import RunLayout
+
         hosts = list(self._agent_urls)
         if host_args is not None:
             if len(host_args) != len(hosts):
@@ -113,19 +123,37 @@ class ParallelHTTPClient:
             commands = [cmd % args for args in host_args]
         else:
             commands = [cmd] * len(hosts)
+        run_dir = RunLayout.get().run_dir
+        out_dir = None
+        if output_mode == messages.ExecOutputMode.FILE:
+            out_dir = run_dir / "exec_output"
+            out_dir.mkdir(parents=True, exist_ok=True)
         return {
             host: messages.ExecRequest(
                 cmd=command,
-                env={},
-                cwd=Path.cwd(),
+                env=env or {},
+                cwd=run_dir,
                 timeout=round(read_timeout) if read_timeout is not None else None,
-                inactivity_timeout=None,
+                inactivity_timeout=round(inactivity_timeout) if inactivity_timeout is not None else None,
                 cmd_id=uuid.uuid4().hex,
-                out_path=None,
-                output_mode=messages.ExecOutputMode.INLINE,
+                out_path=out_dir,
+                output_mode=output_mode,
             )
             for host, command in zip(hosts, commands)
         }
+
+    async def _collect_output(self, exec_response: messages.ExecResponse) -> tuple[list[str], list[str]]:
+        '''FILE mode ships only a tail preview inline; the full output lives on the shared FS at
+        stdout_path/stderr_path, so read it back here to give callers the same list[str] shape
+        regardless of which output_mode produced the response (INLINE/EXIT_CODE_ONLY never set
+        stdout_path, so this falls through to the inline fields for those unchanged).'''
+        if exec_response.stdout_path is not None and exec_response.stderr_path is not None:
+            stdout_text, stderr_text = await asyncio.gather(
+                asyncio.to_thread(exec_response.stdout_path.read_text),
+                asyncio.to_thread(exec_response.stderr_path.read_text),
+            )
+            return stdout_text.splitlines(), stderr_text.splitlines()
+        return exec_response.stdout or [], exec_response.stderr or []
 
     async def _run_one(
         self,
@@ -143,12 +171,13 @@ class ParallelHTTPClient:
             )
             response.raise_for_status()
             exec_response = messages.parse_message(messages.ExecResponse, response.text)
+            stdout, stderr = await self._collect_output(exec_response)
         except Exception as exc:  # noqa: BLE001 - captured per-host so one bad host doesn't sink the others
             return HostOutput(host=host, stdout=[], stderr=[], exit_code=None, exception=_classify_exception(exc))
         return HostOutput(
             host=host,
-            stdout=exec_response.stdout or [],
-            stderr=exec_response.stderr or [],
+            stdout=stdout,
+            stderr=stderr,
             exit_code=exec_response.exit_code,
             exception=None,
         )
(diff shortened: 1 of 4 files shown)


cvs/core/agent/http_client.py
```diff
@@ -0,0 +1,237 @@
...
+        client: httpx.AsyncClient,
+        host: str,
+        url: str,
+        request: messages.ExecRequest,
+        read_timeout: float | None,
+    ) -> HostOutput:
+        try:
+            response = await client.post(
+                f"{url}{messages.EXEC_PATH}",
+                content=request.model_dump_json(),
```
atnair-amd: **Blocking:** `content=request.model_dump_json()` does not set `Content-Type: application/json`. The real FastAPI `/v1/exec` endpoint rejects this request with HTTP 422, so every normal command becomes `HTTPProtocolError`. Please use `json=request.model_dump(mode="json")` or set the JSON content type explicitly, and add an integration test against `create_app()`.

cvs/core/agent/http_client.py
```diff
@@ -0,0 +1,237 @@
...
+                timeout=httpx.Timeout(read_timeout, connect=self._connect_timeout),
```
atnair-amd: **Blocking:** This uses `read_timeout` as the HTTP response deadline, while line 136 sends `round(read_timeout)` as the agent's process deadline. `/v1/exec` returns only after execution finishes; a timed-out process can also need the agent's 10-second termination grace. Repro: a 1s command timeout returns `HTTPConnectionError` to this client after 1s, although the healthy agent returns `timed_out=True` after cleanup. Please give the HTTP read deadline headroom beyond the execution deadline.

cvs/core/agent/http_client.py
```diff
@@ -0,0 +1,237 @@
...
+                details = ", ".join(f"{output.host}: {output.exception}" for output in failed)
+                raise ParallelHTTPClientError(f"{len(failed)} host(s) failed: {details}")
+        return outputs
+
+    async def _fan_out(self, method: str, path: str, stop_on_errors: bool) -> dict[str, bool]:
+        client = self._get_client()
+
+        async def call(host: str, url: str) -> tuple[str, bool | Exception]:
+            try:
+                response = await client.request(method, f"{url}{path}")
```
atnair-amd: **Warning:** `_fan_out()` supplies no request timeout, so `health()` and `shutdown()` ignore the configured `connect_timeout` and use httpx's implicit 5-second deadline. Shutdown can legitimately exceed that while terminating processes. Please define explicit per-operation timeout behavior.

cvs/core/agent/http_client.py
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+    async def _collect_output(self, exec_response: messages.ExecResponse) -> tuple[list[str], list[str]]:
+        '''FILE mode ships only a tail preview inline; the full output lives on the shared FS at
+        stdout_path/stderr_path, so read it back here to give callers the same list[str] shape
+        regardless of which output_mode produced the response (INLINE/EXIT_CODE_ONLY never set
+        stdout_path, so this falls through to the inline fields for those unchanged).'''
+        if exec_response.stdout_path is not None and exec_response.stderr_path is not None:
+            stdout_text, stderr_text = await asyncio.gather(
+                asyncio.to_thread(exec_response.stdout_path.read_text),
```
atnair-amd: **Warning:** FILE-mode dereferences paths supplied by the response without verifying that they are the expected `<run_dir>/exec_output/<cmd_id>.stdout|stderr` files. A malformed or untrusted response can make the controller read another controller-readable file. Please validate resolved paths, parent directory, and expected filename before reading.

cvs/core/agent/http_client.py
```diff
@@ -0,0 +1,237 @@
...
+
+    async def __aexit__(self, *exc_info) -> None:
+        await self.destroy()
+
+    def _auth_header(self) -> dict[str, str]:
+        return {messages.AUTH_HEADER: f"{messages.AUTH_SCHEME} {self._token}"}
+
+    def _get_client(self) -> httpx.AsyncClient:
+        if self._client is None:
+            self._client = httpx.AsyncClient(headers=self._auth_header(), transport=self._transport)
```
atnair-amd: **Warning:** httpx defaults to 100 total connections. With 101 concurrently blocked hosts, the 101st request queues until a connection frees; this defeats simultaneous fan-out and can become `PoolTimeout` when a request timeout is set. If this client supports clusters above 100 hosts, configure explicit limits and cover that behavior.

cvs/core/agent/http_client.py
```diff
@@ -0,0 +1,237 @@
...
+            response.raise_for_status()
+            exec_response = messages.parse_message(messages.ExecResponse, response.text)
+            stdout, stderr = await self._collect_output(exec_response)
+        except Exception as exc:  # noqa: BLE001 - captured per-host so one bad host doesn't sink the others
+            return HostOutput(host=host, stdout=[], stderr=[], exit_code=None, exception=_classify_exception(exc))
+        return HostOutput(
+            host=host,
+            stdout=stdout,
+            stderr=stderr,
+            exit_code=exec_response.exit_code,
```
atnair-amd, line 258: **Warning:** `ExecResponse` carries `timed_out` and `truncated`, but `HostOutput` drops both. Callers cannot distinguish an agent-killed command from another signal termination or detect truncated inline output. Please preserve these fields in the public result.

atnair-amd (changes requested):
## Request changes

The command-execution path is currently incompatible with the FastAPI agent: the client sends a raw JSON body without an `application/json` content type, and the server rejects it with HTTP 422.

Once that is corrected, the command timeout needs to be decoupled from the HTTP response deadline. An agent-side timeout can legitimately require its termination grace period before it can return an `ExecResponse`; the client currently reports that healthy agent as an unreachable host.

The remaining inline comments cover timeout policy for health/shutdown, FILE-mode response-path validation, connection-pool fan-out at scale, and propagation of agent execution state.

I reproduced the reported behaviors with local loopback servers and the actual FastAPI agent implementation. No existing review comments were present when this review was prepared.

cvs/core/agent/http_client.py
```diff
@@ -0,0 +1,302 @@
...
+    ) -> dict[str, messages.ExecRequest]:
+        # Imported here, not at module level: cvs.core.run_layout pulls in cvs/core/__init__.py's
+        # orchestrator factory, which reaches back into cvs/core/agent/ in ways that risk a cycle
+        # (same reasoning as cvs/lib/utils_lib.py's lazy import of RunLayout).
+        from cvs.core.run_layout import RunLayout
+
+        hosts = list(self._agent_urls)
+        if host_args is not None:
+            if len(host_args) != len(hosts):
+                raise ValueError(f"host_args has {len(host_args)} entries but there are {len(hosts)} hosts")
```
atnair-amd, line 191: **Non-blocking:** When `connect_timeout` is omitted, this passes `connect=None` to `httpx.Timeout`, which disables the TCP connection deadline. That means health, shutdown, and command calls can wait for the OS-level SYN timeout rather than the selected read deadline when a host black-holes connection attempts. Please let `httpx.Timeout(read_timeout)` apply its finite connection phase by default, or define an explicit finite default.

cvs/core/agent/http_client.py
```diff
@@ -0,0 +1,302 @@
...
+        try:
+            response = await client.post(
+                f"{url}{messages.EXEC_PATH}",
+                json=request.model_dump(mode="json"),
+                timeout=self._http_timeout(_exec_http_read_timeout(read_timeout)),
+            )
+            response.raise_for_status()
+            exec_response = messages.parse_message(messages.ExecResponse, response.text)
+            stdout, stderr = await self._collect_output(request, exec_response)
+        except Exception as exc:  # noqa: BLE001 - captured per-host so one bad host doesn't sink the others
```
atnair-amd, line 252: **Low priority:** `round()` turns positive fractional timeouts below 0.5 into zero. For example, `read_timeout=0.4` serializes an agent process timeout of `0`, immediately terminating a command that could otherwise complete. Consider rejecting non-positive rounded values or using `ceil()` / a minimum of one second.

atnair-amd (commented):
The six prior findings are addressed in `8f6c8be1` and their threads can be resolved.

These two follow-ups are non-blocking:
- default `connect_timeout=None` disables the TCP connection phase timeout;
- positive sub-second timeouts can round to zero.

Neither affects the corrected normal command path, but both are worth addressing when tightening timeout semantics.

atnair-amd (approved):

cvs/core/agent/http_client.py
speriaswamy-amd: Addressed

cvs/core/agent/http_client.py
speriaswamy-amd: Addressed

PR #372, opened 2026-08-27
feat(vllm): split single and distributed suites

atnair-amd:
## Summary
- replace the unified `vllm` command with explicit `vllm_single` and `vllm_distributed` suites
- scope `vllm_single` to the first cluster host while preserving multi-host MP/Ray execution and one-host fallback for `vllm_distributed`
- derive node count from orchestrator hosts, update shipped configs/threshold keys, and reuse the existing vLLM Run Deck profile
- update vLLM documentation for the new commands; `cvs run vllm` is intentionally removed

## Tracking
- [AIMVT-316](https://amd-hub.atlassian.net/browse/AIMVT-316)

## Test plan
- [x] `make fmt-check`
- [x] `make lint`
- [x] `make ut` (1,536 tests, 1 skipped)
- [x] collect `vllm_single` sample suite (55 tests)
- [x] collect `vllm_distributed` sample suite (55 tests)
- [x] verify installed CLI lists both split suites


cvs/lib/inference/utils/vllm_config_loader.py
```diff
@@ -197,66 +194,39 @@ class VariantConfig(_Forbid):
...
 
         Single-node: ISL=<isl>,OSL=<osl>,TP=<tp>,CONC=<concurrency>
-        Distributed:  ISL=<isl>,OSL=<osl>,TP=<tp>,PP=<pp>,CONC=<concurrency>
+        Distributed:  ISL=<isl>,OSL=<osl>,TP=<tp>,PP=<pp>,NNODES=<n>,CONC=<concurrency>
         """
         base = f"ISL={isl},OSL={osl},TP={self.params.tensor_parallelism},"
-        if int(self.params.pipeline_parallel_size) > 1:
-            base += f"PP={self.params.pipeline_parallel_size},"
+        pp = int(pipeline_parallel_size if pipeline_parallel_size is not None else self.params.pipeline_parallel_size)
+        if int(nnodes) > 1:
```
solaiys: **Run Deck still looks up the old 3-arg cell key, so distributed thresholds will not attach.** `PP=` / `NNODES=` are added only when the caller passes `nnodes>1`. The suite fixture and `_test_metric` do that, but `cvs/lib/report/cell_build.py` still calls `variant_config.cell_key(isl, osl, conc)` (defaults `nnodes=1`). For a distributed cell that becomes `ISL=…,TP=8,CONC=16` while the threshold file is `ISL=…,PP=2,NNODES=2,CONC=16`, so the deck treats every gated metric as missing/record-only.

This is a regression versus main, where `cell_key(isl, osl, conc)` used `params.pipeline_parallel_size` and still emitted `PP=` for distributed configs.

Fix: record the effective topology on the variant in the `vllm_targets` fixture (e.g. `_effective_nnodes` / `_effective_pp`) and have `cell_key` use those when the kwargs are omitted, **or** teach `CellRecordBuilder.build_one` to pass `nnodes` / `pipeline_parallel_size` without breaking SGLang’s 3-arg `cell_key`. Add a unit test that `cell_key(isl, osl, conc)` matches `expected_cells(nnodes=N, …)` after the fixture has run.

cvs/lib/inference/vllm_topology.py
```diff
@@ -0,0 +1,54 @@
...
+    if len(hosts) == 1:
+        return (hosts,), 1
+
+    is_ray = variant.roles.server.serve_args.get("distributed-executor-backend") == "ray"
+    effective_pp = int(variant.params.pipeline_parallel_size)
+    if effective_pp == 1 and not is_ray:
+        raise ValueError("vllm_distributed requires pipeline_parallel_size>1 unless distributed-executor-backend=ray")
+    if not variant.roles.server.ib_netdev:
+        raise ValueError("vllm_distributed requires roles.server.ib_netdev on multi-host clusters")
+    return (hosts,), effective_pp
```
solaiys: **`vllm_distributed` launches on every cluster host, but every shipped recipe is a 2-node job.** This returns the full `orch.hosts` tuple. `VllmJob` then sets `--nnodes` to that length while configs keep `pipeline_parallel_size=2` and every `*_distributed_threshold.json` is keyed `…,PP=2,NNODES=2,…`.

A normal CVS `cluster.json` has N>2 nodes (the templates say to add entries to scale). On an 8-node cluster this will (1) start an 8-node vLLM world with TP=8, PP=2 (GPU count 64 vs TP×PP=16 — serve fails or mis-shards), and (2) look up `NNODES=8` keys that are not in the threshold file. `validate_thresholds_cover_sweep` only errors when `enforce_thresholds` is true; shipped configs have it false, so this is a warning plus a bad launch rather than a hard stop.

SGLang already scopes the orchestrator to the hosts listed in the inference config rather than the whole cluster. Do the same here: restore a declared host count (or host list) and take the first N cluster nodes, **or** fail before launch when `len(hosts)` does not match the recipe (do not rely on `enforce_thresholds`). The one-host fallback can stay.

solaiys (changes requested):
Two merge-blocking correctness issues: the new `cell_key(..., nnodes=...)` contract is not applied on the Run Deck path (distributed threshold specs will not attach), and `vllm_distributed` uses every cluster host while every shipped recipe/threshold still assumes a 2-node job (`PP=2`, `NNODES=2`).

Inline comments have the concrete fixes.

cvs/lib/inference/vllm_job.py
```diff
@@ -144,13 +144,16 @@ def __init__(
         self.ib_hcas = ib_hcas or []
         self.goodput_slo = goodput_slo
         self.log_subdir = log_subdir
+        self.hosts = tuple(orch.hosts)
```
hnimra-amd, line 147: `VllmJob` does not consume `vllm_targets`. It snapshots `orch.hosts` and derives `--nnodes` / `--pipeline-parallel-size` from that length.

`vllm_targets` is what the suite fixture validates against the threshold matrix, and what `_test_metric` uses for `cell_key(..., nnodes=len(vllm_targets[0]))`. If we later slice to the first N cluster hosts only in `build_vllm_targets` (the SGLang-style fix) without also scoping the orchestrator, the job will still bring up a world on every host.

Please drive `self.hosts` from the selected target group (pass it into `VllmJob`, or have the job take `hosts=` explicitly) so launch topology and threshold keys cannot diverge.

cvs/lib/inference/vllm_topology.py
```diff
@@ -0,0 +1,54 @@
...
+    the complete cluster and use their normal single-host fallback when needed.
+    """
+    if mode != "single":
+        return cluster
+
+    scoped = deepcopy(cluster)
+    nodes = scoped.get("node_dict")
+    if isinstance(nodes, dict) and nodes:
+        host, node = next(iter(nodes.items()))
+        scoped["node_dict"] = {host: node}
```
hnimra-amd, line 37: "First cluster host" here is `next(iter(node_dict))`, not `head_node_dict.mgmt_ip`. The unit test even uses a cluster whose head is `node1` and asserts the scoped run uses `node0`.

That will surprise operators who set a specific head / jump host. Either document that vllm_single always uses JSON insertion order of `node_dict` (and that `head_node_dict` is overwritten), or prefer the configured head when it is present in `node_dict`.

hnimra-amd (changes requested):
The suite split itself is the right customer-facing change: `cvs run vllm_single` / `vllm_distributed`, first-host scoping for single, and sharing the existing vLLM Run Deck profile are all good.

I agree with @solaiys on the two merge blockers (not re-filing them):
1. Run Deck still calls `cell_key(isl, osl, conc)` so distributed `PP=`/`NNODES=` keys never attach.
2. `vllm_distributed` takes every cluster host while every shipped recipe/threshold is a 2-node job.

Please also address the remaining inline comments. I would not merge until (1), (2), and the `VllmJob` vs `vllm_targets` split of truth are fixed — otherwise a later "first N hosts" patch can look correct in fixtures and still launch on the full cluster.

**Docs drift:** `cvs/lib/inference/utils/docs/cell-key-format.md` and the utils AGENTS.md still describe the old 3-arg `cell_key` and load-time `_check_thresholds_cover_sweep`. The workloads README still hard-codes `NNODES=2` while the suite now claims "all cluster hosts".

cvs/input/config_file/inference/vllm_mi300x_workloads/mi300x_vllm_deepseek-r1-0528_fp8_distributed_threshold.json
```diff
@@ -1,7 +1,7 @@
 {
   "_comment": "Uncalibrated placeholders for deepseek-r1-0528_fp8 (distributed, TP=8, PP=2, CONC=16). Every value is 0 and means 'not yet measured' -- covers client.*, gpu.* and prom.*. enforce_thresholds is false in the config, so these record without gating. Replace with measured values from a calibration run before enabling enforcement.",
   "_comment_accuracy": "The accuracy block is keyed by accuracy task id, then by lm-eval metric key, so it cannot be pre-enumerated the way client/gpu/prom can -- its contents follow whichever tasks config.json selects in accuracy.tasks, which is empty. Shape to fill in alongside a task: \"accuracy\": {\"<task-id>\": {\"<lm_eval_task>.<metric>\": {\"kind\": \"min\", \"value\": 0}}}, where <metric> is the results.json key with commas replaced by __ (e.g. gsm8k.exact_match__strict-match). Exempt from the sweep-cell coverage check via NON_SWEEP_THRESHOLD_KEYS.",
-  "ISL=1024,OSL=1024,TP=8,PP=2,CONC=16": {
+  "ISL=1024,OSL=1024,TP=8,PP=2,NNODES=2,CONC=16": {
```
amd-droy: why is NNODES needed here? TP and PP will give you the topology.

cvs/lib/report/pytest_extras.py
```diff
@@ -13,13 +13,10 @@
...
     """Build an ``inf_res_dict`` lookup key for a sweep cell."""
-    cell_key_fn = getattr(variant_config, "cell_key", None)
-    if callable(cell_key_fn):
-        return cell_key_fn(isl, osl, concurrency)
     model = getattr(getattr(variant_config, "model", None), "id", "")
     gpu = getattr(variant_config, "gpu_arch", "")
-    policy = seq_combo.get("policy", "default") if isinstance(seq_combo, dict) else "default"
-    return (model, gpu, str(isl), str(osl), str(policy), int(concurrency))
+    name = seq_combo.get("name", "default") if isinstance(seq_combo, dict) else "default"
+    return (model, gpu, isl, osl, name, concurrency)
```
amd-droy: The result-tuple path dropped normalization: this used to be (model, gpu, str(isl), str(osl), str(policy), int(concurrency)). Raw isl/osl/concurrency will miss inf_res_dict when one side is "16" and the other is 16 (parametrize vs HTML extras), so the card shows empty actuals.

cvs/input/config_file/inference/vllm_mi300x_workloads/mi300x_vllm_deepseek-r1-0528_fp8_distributed_threshold.json
hnimra-amd: @amd-droy from what I understand, TP and PP say how the model is split across GPUs (tensor vs pipeline). They do not always tell you how many machines were in the run. Node count comes from the cluster file and which command you run (vllm_single vs vllm_distributed). The same config can end up on 1 host or 2, with different performance. So the threshold key adds NNODES=2 to mean: “these pass/fail limits are for a run that actually used 2 hosts”; not “only use 2 nodes” (the cluster and suite already control that).
For the usual case today such as for distributed, PP=2, two hosts, PP=2 and NNODES=2 often go together. NNODES is still useful so a 1-host fallback or a future multinode layout doesn’t accidentally reuse the wrong threshold row when TP/PP/CONC look the same.

It’s not redundant noise. Once nnodes is removed from the vLLM config, NNODES= in the key is how you tie a threshold row to the actual host count of that run (2-node distributed vs 1-node fallback, same config).

amd-droy: already discussed. NNODES can be removed, TP*PP will give no of gpus in use.

cvs/lib/inference/utils/vllm_config_loader.py
atnair-amd: Addressed in 5011e3d5 and b165dbe6. The suite now binds one EffectiveVllmTopology before coverage/report construction; cell_key() and expected_cells() consume that bound topology, and the distributed Run Deck regression test verifies the PP-qualified threshold attaches and produces the expected fail status.

cvs/lib/inference/vllm_topology.py
atnair-amd: Addressed in 5011e3d5 and 1947e4df. vllm_distributed now fails before launch unless the effective cluster has one host (fallback) or exactly two hosts; larger clusters are rejected independently of threshold enforcement. The supported placement contract is documented.

cvs/lib/inference/vllm_job.py
atnair-amd: Addressed in 5011e3d5. vllm_targets binds the immutable EffectiveVllmTopology to the variant. VllmJob consumes that topology for nnodes/PP and verifies topology.hosts exactly matches tuple(orch.hosts), so a future target slice without matching orchestrator scope fails instead of launching a different world.

cvs/lib/inference/vllm_topology.py
atnair-amd: Addressed in ce2bf1f4. First-host insertion order is intentional for this PR. The workload README and both vLLM docs now state that mapping-style node_dict uses the first JSON key and that vllm_single rewrites head_node_dict.mgmt_ip. The regression test name now explicitly pins insertion order and configured-head overwrite.

cvs/input/config_file/inference/vllm_mi300x_workloads/mi300x_vllm_deepseek-r1-0528_fp8_distributed_threshold.json
atnair-amd: Addressed in b165dbe6 and the consolidated config update from #379. NNODES was removed from vLLM threshold identity and shipped threshold keys; distributed keys retain PP as the topology discriminator.

cvs/lib/report/pytest_extras.py
atnair-amd: Addressed in b165dbe6. Both the vLLM result producer and the HTML row-extra consumer normalize ISL/OSL/name to strings and concurrency to int, with regression coverage including the empty GPU slot used after #379.

solaiys (approved):
LGTM

The two earlier blockers are addressed: Run Deck uses the bound effective topology so distributed `PP=` cell keys attach, and `vllm_distributed` rejects clusters larger than two hosts.

hnimra-amd (approved):
Previous blockers look addressed: Run Deck uses the bound topology so distributed PP= keys attach, vllm_distributed rejects N>2 before launch, and VllmJob checks topology hosts against the orchestrator. LGTM.

amd-droy (approved):

PR #373, opened 2026-08-28
fix(report): import generate_rundeck from submodule directly

sukesh-amd:
Python resolves submodule names before invoking __getattr__, so `from cvs.lib.report.rundeck import generate_rundeck` returned the module object instead of the function, causing TypeError at session finish.

### DIFF
--- a/cvs/lib/report_plugins.py
+++ b/cvs/lib/report_plugins.py
@@ -617,7 +617,7 @@ class HtmlReportManager:
         if not self.is_enabled:
             return
 
-        from cvs.lib.report.rundeck import generate_rundeck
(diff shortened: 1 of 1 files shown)


solaiys (approved):
The diagnosis is correct: `cvs.lib.report.rundeck` has both a submodule `generate_rundeck.py` and a lazy `__getattr__` export of the same name. `from cvs.lib.report.rundeck import generate_rundeck` binds the module, so `generate_rundeck(session, self)` TypeErrors at session finish.

Importing the function from `cvs.lib.report.rundeck.generate_rundeck` matches what `__init__.py` already does internally. This is the only call site. No concurrency risk.

CI `test` and `container-smoke` are green.

LGTM.

PR #374, opened 2026-08-28
jaxmaxtext : old jax suite removal + doc update + some fixes

solaiys:
## Summary

JAX MaxText suite hardening + docs/reference cleanup. Retires the legacy `jax`
training path in favor of `jaxmaxtext`, wires `nccl.ib_gid_index` through as a
real (user-set) NCCL export, fixes two report/config-loading bugs, and adds
full README/Sphinx documentation. 5 commits.

### `f33f93a` fix(JAX) Source scratch dir from paths.temp_dir and skip tokenizer for synthetic data
- Resolve the in-container scratch base from a new optional `paths.temp_dir`
  (fallback `/tmp/cvs/jaxmaxtext`) instead of the in-container `id -un` — docker
  jobs run as root, so the old logic collided on `/tmp/root` across users. All
  jaxmaxtext configs set `temp_dir=/tmp/{user-id}/jaxmaxtext`.
- Skip the HuggingFace tokenizer download (both `setup_tokenizer()` and the
  `test_setup_tokenizer` stage) when every enabled run uses
  `dataset_type=synthetic`; replaced the deprecated `huggingface-cli download`
  with `hf download`. Added `Paths.temp_dir` (optional).

### `0e454ae` fix(jax) make ib_gid_index=<changeme> as user input in config files
- Wire `nccl.ib_gid_index` to `export NCCL_IB_GID_INDEX` (like `ib_hca`), so the
  `nccl` block — not `env_vars` — is authoritative; removed the redundant
  `NCCL_IB_GID_INDEX` from `env_vars`.
- Removed the unused `nccl.ib_tc` / `ib_sl` (kept in `env_vars`), and added
  `ib_gid_index` to the `<changeme>` hard-exit validator so distributed runs
  fail clearly until it's set. Updated all 12 distributed configs.

### `d687121` chore(JAX) Retire legacy jax suite/lib/configs; add jaxmaxtext README + docs
- Removed the legacy `jax` suite (`cvs/tests/training/jax/`), its driver
  `cvs/lib/jax_training_lib.py`, the `training/jax/` configs, and `jax.rst`.
- Repointed all remaining references to `jaxmaxtext`: `copy_config_plugin`
  help/test, and docs (`run-cvs-tests.rst`, `cvs-install.rst`,
  `configure-config.rst`, `what-is-cvs.rst`, `_toc.yml.in`).
- Added `docs/reference/configuration-files/jaxmaxtext.rst` (per-block field
  tables, config snippet, threshold-file format), `cvs/lib/training/jaxmaxtext/README.md`,
  and expanded the config/tests READMEs (inventory, `temp_dir`, `ib_gid_index`,
  `rdma_lib` backup semantics, `smoke` + `checkpoint_resume` blocks).

### `9604f65` fix(report) Link Config File in HTML env table regardless of filename
- The HTML env table linkified the config/cluster row by matching a filename
  substring ("config"/"cluster"), so configs named `..._distributed.json` never
  got a link. Now uses the role recorded at bundle-copy time instead of guessing.

### `e74e1bb` fix(config) Fall back to packaged threshold when config runs from a copied path
- `substitute_config()` resolved `threshold_json` only next to the config, so
  running a config copied elsewhere (without its sibling threshold) failed.
  Added a fallback that looks the threshold up by filename in the packaged
  `input/config_file/**` tree (framework-neutral; sibling still takes precedence).

## Test plan
- `make ut` — 1531 passed, 1 skipped.
- `make fmt-check` / `make lint` — clean (ruff + pylint 10/10).
- Sphinx docs build — no jax/jaxmaxtext reference warnings; new JAX MaxText page renders.
- New unit tests: env-table link role-tracking, threshold packaged-fallback,
  `ib_gid_index` export + `<changeme>` validator, synthetic tokenizer-skip, `temp_dir` scratch.


amd-droy (approved):
looks good. thank you.

PR #375, opened 2026-08-28
fix(preflight): handle amd-smi builds that emit concatenated JSON documents

paklui:
## Summary
- On amd-smi 27.0.0 (the build with AI-NIC/Pensando integration), `amd-smi list --json` prints **two concatenated top-level JSON documents** from one call — a GPU array, then an AI-NIC array. `_parse_json()` called plain `json.loads()`, which rejects the trailing document, so `node_health` fails on every node with `did not return valid JSON: Extra data` even though the GPUs are healthy and node-health is a mandatory gate, so this cascades into `ifoe_l2_connectivity`, `transferbench_smoke`, and `rdma_connectivity` all being marked `BLOCKED`.
- Fix has two parts, both in `cvs/lib/preflight/scaleup_fabric.py`:
  1. `_parse_json()` decodes a stream of concatenated documents via `json.JSONDecoder().raw_decode()` in a loop, returning `List[Any]` instead of a single document.
  2. `parse_amd_smi_gpu_json()` skips records with no GPU identifier (`gpu_id is None`) **before** the BDF check. This matters because the AI-NIC record's BDF ends in `.0` — identical in form to a GPU BDF — so it would otherwise pass the existing `.0` filter and be miscounted as an extra GPU (e.g. 5 "GPUs" reported on a 4-GPU node instead of the correct 4).

## Verification
Reproduced on a real 4-GPU Helios-R node (amd-smi `27.0.0+920e5a8a`, ROCm 10.0.0, gfx1250):

```
BEFORE: ['amd-smi list --json did not return valid JSON: Extra data']
AFTER : 4 GPUs, errors=[]
```

End-to-end `preflight_checks`: `1 failed, 10 passed` → `11 passed`.

Single-document output (older amd-smi builds) is unaffected — the loop simply yields one document, and the pre-existing test `test_amd_smi_gpu_inventory_requires_four_descriptors` (single doc, 4 GPUs) still passes unchanged.

## Test plan
- [x] Added regression test `test_amd_smi_gpu_inventory_ignores_concatenated_ai_nic_document` — feeds a GPU array plus a concatenated AI-NIC array, asserts exactly 4 GPUs with the NIC excluded
- [x] Full `cvs/lib/preflight/unittests/test_scaleup_fabric.py` module: 24/24 tests pass
- [x] `make fmt-check` clean

### DIFF
--- a/cvs/lib/preflight/scaleup_fabric.py
+++ b/cvs/lib/preflight/scaleup_fabric.py
@@ -90,14 +90,33 @@ def _walk_dicts(value: Any, inherited_bdf: Optional[str] = None) -> Iterable[Tup
             yield from _walk_dicts(child, inherited_bdf)
 
 
-def _parse_json(output: str, command_name: str) -> Tuple[Optional[Any], List[str]]:
+def _parse_json(output: str, command_name: str) -> Tuple[List[Any], List[str]]:
+    """Decode one or more whitespace-separated top-level JSON documents.
+
+    Some amd-smi builds print multiple JSON documents back-to-back for a
+    single ``list --json`` invocation (e.g. a GPU array followed by an
+    AI-NIC array), which a plain ``json.loads`` rejects as trailing data.
+    """
     raw = (output or "").strip()
     if not raw:
-        return None, [f"{command_name} returned empty output"]
-    try:
-        return json.loads(raw), []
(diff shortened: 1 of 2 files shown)


speriaswamy-amd (approved):
Looks good to me

PR #377, opened 2026-08-31
Add cvs config command for browsing and copying

cijohnson:
config templates

Currently with many config templates avaialble in the cvs repo, users can only list all the config files, which is cluttered and not user friendly. if there is a way to list the config files by directory/category, user will get sub set of relavent config file templates.

Introduce list, list-dirs, and copy subcommands with grouped directory output, shared config_files helpers, and generate-style help for bare invocations. Deprecate copy-config --list while keeping copy-config copy behavior for backward compatibility.

## Submission Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -186,7 +186,7 @@ def test_example(orch, config_dict):
 
 ### Configuration References
 - [cvs/input/cluster_file/README.md](cvs/input/cluster_file/README.md) - Cluster configuration format
-- Sample configs: `cvs copy-config --list` for available templates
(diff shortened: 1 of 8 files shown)


solaiys (approved):
Adds `cvs config list` / `list-dirs` / `copy` with shared helpers, keeps `copy-config` copy behavior, and deprecates `copy-config --list`. Plugin discovery order, path scoping, and copy overwrite/`--force` look correct. Relative paths under `config_file` / `cluster_file` / `env_file` do not collide, so first-match copy is safe. No concurrency issues (single-threaded CLI). Tests cover list/copy and the deprecation path; CI is green.

LGTM.

PR #379, opened 2026-08-31
chore(vllm): consolidate inference workload configs

atnair-amd:
## Summary
- Consolidate the MI325X vLLM workload configs into one directory.
- Keep reusable workload config names under mi3xx; name platform-specific thresholds mi325x.
- Select 1k1k at concurrencies 16 and 32 by default, with threshold entries for those selected cells.
- Keep accuracy evaluation opt-in for offline-compatible default runs.
- Remove obsolete and unshipped legacy vLLM configuration files, and update docs.

## Validation
- `make fmt-check lint ut`
- All 28 configs validate against the vLLM loader and resolve to their explicit MI325X thresholds.
- `cvs config list inference/vllm`


cvs/input/config_file/inference/vllm/mi3xx_vllm_llama33-70b_fp8_single_config.json
```diff
@@ -104,10 +104,20 @@
...
+        "combo": "w_llama33-70b_fp8_1k1k",
+        "concurrency": 32
       }
     ]
   },
   "accuracy": {
-    "tasks": []
+    "tasks": [
+      {
+        "id": "gsm8k",
```
solaiys: **Enabling GSM8K on every config will fail `test_accuracy_eval` on the default offline path.** This used to be `"tasks": []` (pytest auto-skips). Now every one of the 32 configs selects `gsm8k`. `run_accuracy_tasks` builds `source /tmp/server_env_script.sh && … lm_eval --tasks gsm8k`, and that env script exports `HF_HUB_OFFLINE=1` and `TRANSFORMERS_OFFLINE=1` from `roles.server.env` (line 46). lm-eval then cannot download the GSM8K split (or a tokenizer that is not already in `HF_HUB_CACHE`). A nonzero exit becomes `RuntimeError` and a failed HTML row. Accuracy does not set `lifecycle.failed`, so perf still records — but every default `cvs run vllm` now has a red accuracy stage unless the dataset is already cached.

Keep `tasks: []` (opt-in) as before, **or** run lm-eval without the offline exports and document that GSM8K must be pre-cached, **or** skip the accuracy node when the dataset is missing. Do not ship enabled-by-default GSM8K on an offline server env.

solaiys (dismissed):
One merge-blocking regression: every shipped config now enables GSM8K by default, but `test_accuracy_eval` sources the server env script (`HF_HUB_OFFLINE=1` / `TRANSFORMERS_OFFLINE=1`), so lm-eval cannot fetch the GSM8K dataset (or a missing tokenizer) and the accuracy row fails on a stock run. Previously `accuracy.tasks` was empty and that node was skipped.

Inline comment has the concrete fix. Config/threshold pairing, cell keys, nnodes/PP, and docs path updates otherwise look consistent with the main-branch vLLM loader.

cvs/input/config_file/inference/vllm/mi3xx_vllm_llama33-70b_fp8_single_config.json
atnair-amd: Fixed in b830e7aa. Restored empty `accuracy.tasks` and `accuracy` threshold maps across all shipped vLLM configs, so GSM8K remains opt-in and default offline runs continue to skip the accuracy stage. Verified all 32 configs against the loader schema and ran `test_vllm_config_loader`.

cvs/input/config_file/inference/vllm/mi3xx_vllm_glm-51_fp8_single.json
```diff
@@ -1,10 +1,10 @@
 {
```
cijohnson, line 1: need to remove _config.json to be consitent with sglang and atom

cvs/input/config_file/inference/vllm/mi3xx_vllm_glm-51_fp8_single_config.json
```diff
@@ -1,10 +1,10 @@
 {
-  "_comment": "vllm single workload: glm-51_fp8 on MI300X, TP=8 single node. Sweep carries conc16 1k1k/1k8k/8k1k; only 1k1k is selected in sweep.runs. num_prompts is 320, not the 3200 schema default, to keep this a characterization pass rather than a full-length benchmark -- raise it to 3200 for comparable published numbers. Set container.image, model.id and the models mount before running.",
+  "_comment": "vllm single workload: glm-51_fp8 on MI325X, TP=8 single node. Sweep carries 1k1k/1k8k/8k1k at concurrencies 16 and 32; only 1k1k is selected in sweep.runs. num_prompts is 320, not the 3200 schema default, to keep this a characterization pass rather than a full-length benchmark -- raise it to 3200 for comparable published numbers. Set container.image, model.id and the models mount before running.",
   "schema_version": 1,
   "framework": "vllm",
-  "gpu_arch": "mi300x",
+  "gpu_arch": "mi325x",
```
cijohnson: do we need this?

cvs/input/config_file/inference/vllm/mi325x_vllm_llama33-70b_fp8_single_threshold.json
```diff
@@ -1,5 +1,5 @@
 {
```
cijohnson, line 1: threshold should be mi325 and not mi3xx as the threshold will change for each platform

cijohnson (commented):
add config and threshold only for the platform tested, for the old vll, we can remove as 0.1.0 release did not shp it

cijohnson (changes requested):

cvs/input/config_file/inference/vllm/mi3xx_vllm_glm-51_fp8_single_config.json
atnair-amd: Removed in 89aab364. `gpu_arch` did not affect vLLM launch or threshold selection; it was only redundant report/cache metadata once the selected config and threshold identify the platform. The vLLM schema and result keys no longer require it.

cvs/input/config_file/inference/vllm/mi3xx_vllm_glm-51_fp8_single.json
atnair-amd: Addressed in eba798a7. The 28 workload configs now use the mi3xx_vllm_…json convention, without the _config suffix.

cvs/input/config_file/inference/vllm/mi325x_vllm_llama33-70b_fp8_single_threshold.json
atnair-amd: Addressed in eba798a7. The corresponding threshold files now use the platform-specific mi325x_vllm_…_threshold.json prefix, and every config threshold_json reference was updated.

atnair-amd:
Addressed the remaining review feedback in eba798a7: removed the unshipped MI35X legacy workload pairs, removed _config from the 28 generic config filenames, and renamed the corresponding platform-specific thresholds to mi325x_.

cvs/input/config_file/inference/vllm/mi3xx_vllm_qwen35-397b-a17b_bf16_distributed.json
```diff
@@ -108,6 +107,10 @@
       {
         "combo": "w_qwen35-397b-a17b_bf16_1k1k",
         "concurrency": 16
+      },
+      {
+        "combo": "w_qwen35-397b-a17b_bf16_1k1k",
+        "concurrency": 32
```
amd-droy, line 113: why is the sweep block needed? everything is there in thresholds file, isn't this duplicated?

atnair-amd: no, in vllm/atom design, there is a sweep definition block and a runs block. in the sweep definition block, the user can specify and predefine any sweep experiments of their choosing. The run block is the opt in block - if the user wants to run only 1k1k conc 16 they can do so. This is so that the user can have several predefined configs and run the experiments that they want

atnair-amd:
The full 15-cell threshold catalog and custom record-only sweep behavior have been deferred to a dedicated follow-up PR. This PR retains its focused config-consolidation scope.

atnair-amd:
0ca9f59c fixes post-run vLLM report generation: result keys retain the shared six-field shape with an empty GPU column after gpu_arch was removed from the config schema. Added a regression test for the console results table.

amd-droy (approved):

cvs/input/config_file/inference/vllm/mi325x_vllm_gpt-oss-120b_mxfp4_distributed_threshold.json
```diff
@@ -0,0 +1,265 @@
+{
```
cijohnson, line 1: is the name correct? is it 120b or 20b ?

atnair-amd: Corrected in fcffa431: the workload is GPT-OSS 120B MXFP4. Both single-node and distributed config/threshold pairs were renamed from gpt-oss-20b_fp8 to gpt-oss-120b_mxfp4.

cijohnson (approved):

PR #382, opened 2026-09-01
fix(health): scope TransferBench CPU endpoints on sparse NUMA [AIMVT-314]

speriaswamy-amd:
## Summary
- Fixes [AIMVT-314](https://amd-hub.atlassian.net/browse/AIMVT-314): Helios-R/Venice reports ~38 NUMA nodes but only 2 have CPUs. TransferBench `p2p`/`scaling` enumerated all of them, aborted at CPU 2 (`cannot allocate memory due to process memory policy/cpuset`), and `parse_tb_p2p_bw()` crashed with `AttributeError` on missing UniDir averages.
- Cluster `env_vars` never reached the `sudo bash -c` wrapper, and health JSON had no `NUM_CPU_DEVICES` knob. This does **not** switch to `podp2p` (that would drop CPU↔GPU coverage).
- Inject env **inside** `sudo bash -c`. Auto-detect populated CPU NUMA count (`auto_num_cpu_devices`, default true) or set `transferbench.num_cpu_devices` / `transferbench.env`. Harden the p2p parser to `fail_test()` with a NUMA hint instead of crashing; match the cpuset abort in `scan_test_results()`.

## Test plan
- [x] New unit tests in `test_transferbench_health.py` (env prefix, NUMA count parsing, abort fixture vs scoped success)
- [x] Hardware k19-1: auto-detect `NUM_CPU_DEVICES=2`; `test_transfer_bench_p2p` **PASSED** in 65s (4 GPUs, UniDir GPU→GPU avg ~190 GB/s). No NUMA abort, no parser crash.


cvs/tests/health/transferbench_cvs.py
```diff
@@ -7,16 +7,156 @@
...
 log = globals.log
 
+_ENV_VAR_NAME_RE = re.compile(r'^[A-Za-z_][A-Za-z0-9_]*$')
+_LINUX_ID_LIST_PART_RE = re.compile(r'^(\d+)(?:-(\d+))?$')
+_DETECT_NUM_CPU_DEVICES_CMD = (
+    "n=$(grep -l '[0-9]' /sys/devices/system/node/node*/cpulist 2>/dev/null | wc -l | tr -d ' '); "
+    "if [ -n \"$n\" ] && [ \"$n\" -gt 0 ]; then printf '%s\\n' \"$n\"; "
+    "else awk '/^Mems_allowed_list:/ {print $2}' /proc/self/status; fi"
+)
+_TB_NUMA_ABORT_HINT = (
```
cijohnson, line 32: should we use

bash -c '
c=0
for f in /sys/devices/system/node/node*/cpulist; do
  [ -f "$f" ] || continue
  grep -q "[0-9]" "$f" || continue
  c=$((c+1))
done
[ "$c" -gt 0 ] && printf "%s\n" "$c"
'

NOTE: wrapped in bash -c and removed else part, Do you think we need else part? i think it will be a unreachable code

speriaswamy-amd: Switched the probe to the cpulist for-loop. Kept the `Mems_allowed_list` fallback when `c=0`: with an unmatched glob (no `node*/cpulist`) the loop body never runs, so that path is reachable in containers/cgroup-only views. `orch.exec` already runs the snippet under a remote shell, so I left off an extra `bash -c` wrapper.

Pushed in 001072d3.

cijohnson (approved):

PR #383, opened 2026-09-01
fix(preflight): configurable l2ping ssh_timeout (Helios-R 180s abort)

speriaswamy-amd:
## Summary
- IFoE L2 ping hardcoded `ssh_timeout=180` and `loss_threshold_pct=0.0`. Schema only allowed `enabled` + `pings_per_port`, so operators could not override.
- On Helios-R (k19-1), `afmctl test ping -c 3` often runs **~2–4 min per invocation** (half-cabled / failing dst accelerators). That exceeds 180s, so PSSH killed the sweep and pruned the node even when the fabric was still producing tables. `afmctl -t` itself defaults to 5 minutes.
- This PR makes `ssh_timeout` (default **600s**) and `loss_threshold_pct` first-class `connectivity_check.ifoe.l2ping` keys, wires them through `IfoeL2ConnectivityCheck`, and shows the timeout on the preflight HTML report.

## Test plan
- [x] Unit tests in `test_ifoe_l2_connectivity.py` (schema accept/reject, config wiring, default 600s)
- [x] Hardware: k19-1 with `ssh_timeout=600` — slow invocations completed (~4 min); pytest-html recorded 600s. Remaining FAIL is fabric (port 33 + ~50% full-mesh loss), not the timeout.


cvs/input/config_file/preflight/README_preflight_config.md
```diff
@@ -308,6 +308,13 @@ port and validates per-port and aggregate summary accounting.
   - Enables the mandatory L2 connectivity gate before TransferBench and RDMA
 - **`pings_per_port`** (default: `3`)
   - Number of ping samples sent per selected IFoE port pair
+- **`ssh_timeout`** (default: `600`)
```
cijohnson: is ssh_timeout right name? shouldn;t it be ping_timeout or read_timeout? becasue this timeout is passed to run_command read_timeout ( https://parallel-ssh.readthedocs.io/en/latest/native_parallel.html#pssh.clients.native.parallel.ParallelSSHClient.run_command)

I think the existing code used the name DEFAULT_SSH_TIMEOUT_SEC , that is why you are using the same in the user input aswell.

Lets make user input json as read_timeout or ping_timeout, later during the optimization PR, lib can fix DEFAULT_SSH_TIMEOUT_SEC

cijohnson (changes requested):

cvs/input/config_file/preflight/README_preflight_config.md
speriaswamy-amd: Renamed the customer JSON key to `ping_timeout` (PSSH `read_timeout` for each `afmctl` ping). `IfoeL2ConnectivityCheck` / `DEFAULT_SSH_TIMEOUT_SEC` are unchanged so the lib rename can land in the later optimization PR.

Pushed in 18f1bf47.

cijohnson (approved):

PR #385, opened 2026-09-03
feat(inference): expand lm-eval accuracy configuration

atnair-amd:
## Summary
- Pin the runtime lm-eval dependency to 0.4.12 and expose its evaluation controls with defaults.
- Preserve existing Atom accuracy configuration through the legacy task alias and shared regression coverage.

## Test plan
- [x] make fmt-check
- [x] make lint
- [x] make ut

### DIFF
--- a/cvs/lib/inference/utils/accuracy_config.py
+++ b/cvs/lib/inference/utils/accuracy_config.py
@@ -4,31 +4,104 @@ All rights reserved.
 
 Accuracy-evaluation config schema, shared across inference suites.
 
-`AccuracyTask`/`AccuracyConfig` define the `config.json`-side selection schema
-for lm-eval-harness based accuracy tasks (see cvs/lib/inference/utils/AGENTS.md
-for the broader accuracy-evaluation design). This module holds selection only
--- no threshold/gating values, which live in the sibling threshold.json file
-and are joined against `AccuracyConfig.tasks` at runtime by a later unit.
+`AccuracyTask` describes an lm-evaluation-harness 0.4.12 invocation against a
+live OpenAI-compatible server. CVS owns the live-server endpoint, model path,
+output path, and sample logging. Every other supported ``lm_eval run`` setting
+has an explicit field and its 0.4.12 default.
 '''
 
 from __future__ import annotations
 
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Literal, Optional, Union
 
-from pydantic import model_validator
+from pydantic import Field, field_validator, model_validator
 
 from cvs.lib.utils.config_loader import _Forbid
 
 
 class AccuracyTask(_Forbid):
     id: str
-    task: str
-    num_fewshot: int = 0
-    metadata: Dict[str, Any] = {}
-    include_path: str = ""
+    # ``task`` remains accepted so existing Atom and vLLM configs need no
+    # migration. ``tasks`` is the lm-eval 0.4.12 spelling and supports a
+    # comma-separated string or a list of task names.
+    tasks: Optional[Union[str, List[str]]] = None
+    task: Optional[str] = None
+
+    # CVS/OpenAI adapter selection. These are deliberately narrower than
+    # lm-eval's generic --model field because CVS always targets a live local
+    # OpenAI-compatible server.
+    backend: str = ""
+    lm_eval_model: Optional[Literal["local-completions", "local-chat-completions"]] = None
+    extra_model_args: str = ""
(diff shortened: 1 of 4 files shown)


cvs/lib/inference/utils/lm_eval_job.py
```diff
@@ -38,40 +40,109 @@ def normalize_client_base_url(base_url: str) -> str:
...
         "--model",
         model_flag,
         "--model_args",
-        model_args,
+        rendered_model_args,
         "--tasks",
-        task.task,
-        "--num_fewshot",
-        str(task.num_fewshot),
+        *task.task_names(),
```
solaiys: **Blocking: `--tasks` must be one comma-separated argument.**

lm-eval 0.4.12 declares `--tasks` as `type=str` (`metavar=task1,task2`). `*task.task_names()` emits extra argv tokens, so a list or comma-separated `tasks` becomes `--tasks hellaswag gsm8k` and the harness errors before evaluation.

Pass a single value, same as the old `task.task` path:

```python
"--tasks",
",".join(task.task_names()),
```

Update `test_complete_lm_eval_surface_is_forwarded` to expect `--tasks hellaswag,gsm8k`, not a space-separated pair.

solaiys (dismissed):
## Review: blocking issue

lm-eval 0.4.12's `--tasks` is a single comma-separated string (`type=str`, metavar `task1,task2`). Splatting `task_names()` turns `tasks: ["hellaswag", "gsm8k"]` or `task: "hellaswag,gsm8k"` into `--tasks hellaswag gsm8k`, which argparse treats as an unknown positional. Pass `",".join(task.task_names())` as one argument (the historical `task.task` form). The new unit test currently asserts the broken CLI shape.

cvs/lib/inference/utils/lm_eval_job.py
atnair-amd: Dismissing this as a 0.4.12 CLI mismatch, not a harness-contract bug.

The `type=str` / `metavar=task1,task2` parser is the older `lm_eval/__main__.py` interface. The published `lm-eval==0.4.12` wheel (what `LM_EVAL_INSTALL_CHECK_CMD` pins) uses the new `lm_eval/_cli/run.py` parser:

```python
"--tasks", nargs="+", metavar="<task>", action=SplitArgs
```

Usage is `--tasks <task> <task>`. `HarnessCLI.parse_args()` inserts `run` when the first token is not a subcommand, so `lm_eval --model … --tasks hellaswag gsm8k --output_path …` is valid. `nargs="+"` stops at the next `--` flag, so `gsm8k` is not an unknown positional.

`SplitArgs` also splits commas, so a single `hellaswag,gsm8k` token would parse too. The splat form matches 0.4.12's own usage string; switching to a comma-joined single argument is not required for this pin, and comma-only lists have been brittle upstream (EleutherAI/lm-evaluation-harness#3448).

The unit test asserting `--tasks hellaswag gsm8k` is checking the 0.4.12-correct argv, so I am leaving `*task.task_names()` as-is.

cvs/lib/inference/utils/accuracy_config.py
```diff
@@ -4,31 +4,128 @@
...
+    extra_model_args: str = ""
     num_concurrent: int = 8
-    apply_chat_template: bool = False
-    gen_kwargs: Dict[str, Any] = {}
+
+    # lm-eval 0.4.12 evaluation settings.
+    num_fewshot: Optional[int] = None
+    batch_size: str = "1"
+    max_batch_size: Optional[int] = None
+    device: Optional[str] = "cuda:0"
```
hnimra-amd: This default always forwards `--device cuda:0` (via `_optional_arg` when the value is not `None`). In 0.4.12 `--device` defaults to unset.

CVS still runs `lm_eval` as an API client (`local-completions` / `local-chat-completions`) on the same node as the server. Pinning that client to GPU 0 is not required for those backends and can contend with the serving job.

Default `device` to `None` and omit the flag unless the config sets it.

hnimra-amd (commented):
Not re-filing the `--tasks` splat. For lm-eval 0.4.12 that matches `nargs="+"` / `SplitArgs`, and `HarnessCLI` inserts `run` when the first token is `--model`.

Two schema-default notes below. The `device` one is the one I would treat as should-fix; `num_fewshot` is a scoring-contract check.

hnimra-amd (approved):
LGTM!

amd-droy (approved):
looks good to me. thanks Atul.

PR #386, opened 2026-09-03
feat(vllm): adopt canonical workload configuration

atnair-amd:
## Summary
- Replace the vLLM workload schema with direct server and benchmark parameters plus canonical explicit run cells.
- Move static environment to the container, preserve canonical cell identity in reports, and migrate the packaged workload catalog.

## Test plan
- [x] make fmt-check
- [x] make lint
- [x] make ut
- [x] pytest --collect-only for single and distributed vLLM configs


cvs/lib/inference/vllm_job.py
```diff
@@ -246,11 +223,6 @@ def _server_argv(self, rank: int) -> list:
             "--port",
             str(self.port_no),
         ]
```
solaiys, line 231: This used to emit a derived `--max-model-len` (ISL+OSL, range ratio, prefix pad) whenever `serve_args` did not pin one. That path is gone, and none of the 28 packaged catalog configs set `server_params.max_model_len`, so `vllm serve` now uses each model's full default context (often 128k).

Smoke still injects `max_model_len=512`, which shows the serve path needs a cap; the benchmark cells do not get one. That can OOM or change KV-cache behavior versus the previous suite.

Fix: restore the derivation when the option is omitted, or put an explicit `max_model_len` on every packaged `server_params` (and document it as required). Empty `{}` sweeps with ISL=1024/OSL=1024 are not a substitute.

cvs/lib/inference/vllm_job.py
```diff
@@ -306,9 +280,6 @@ def build_server_cmd(self):
         env_lines = [
```
solaiys, line 288: The previous job always exported `VLLM_USE_AITER_UNIFIED_ATTENTION=1`, `VLLM_ROCM_USE_AITER_MHA=0`, and `VLLM_ROCM_USE_AITER_FUSED_MOE_A16W4=1` into the serve env script. This PR removes them and sets `server_env = {}`, but packaged `container.env` only has `HF_HUB_OFFLINE` / `TRANSFORMERS_OFFLINE` (those already lived on `roles.server.env`).

The summary says static environment moves to the container; these AITER flags were static and did not move. Every ROCm vLLM cell will now run without them unless the image happens to set them.

Add the three exports to packaged `container.env` (or keep writing them here). `container.env` is already wired through `ContainerOrchestrator.get_environment()`.

solaiys (dismissed):
Two regressions from the schema move block merge: derived `--max-model-len` is no longer applied and none of the packaged configs set it, and the ROCm AITER exports were dropped from the serve env script without landing in `container.env`.

cvs/input/config_file/inference/vllm/mi3xx_vllm_deepseek-v4-flash_fp8_distributed.json
```diff
@@ -1,30 +1,34 @@
...
-    "remote": 0
-  },
   "container": {
+    "_comment": "Container definition and static environment inherited by server, benchmark, and accuracy commands.",
     "lifetime": "per_run",
     "name": "vllm_deepseek-v4-flash_fp8_distributed_mi325x",
     "image": "<changeme>",
+    "env": {
+      "HF_HUB_OFFLINE": "1",
+      "TRANSFORMERS_OFFLINE": "1",
```
amd-droy, line 27: Are these 2 used anywhere? if not, can we drop them - "HF_HUB_OFFLINE": "1","TRANSFORMERS_OFFLINE": "1"?

amd-droy:
3+ hosts — fails before launch with: “vllm_distributed currently supports exactly two cluster hosts…” - Not for this PR/Release, but later we need to update it for n nodes.

hnimra-amd (commented):
A few contract/docs notes on the new option serializer and cell overrides. Packaged catalog runs still look fine.

cvs/lib/inference/utils/vllm_config_loader.py
```diff
@@ -1,253 +1,322 @@
...
+    """Serialize generic snake-case option maps for vLLM command lines."""
+    _validate_cli_option_map(options, section="option")
+    argv = []
+    for name, value in options.items():
+        if value is None:
+            continue
+        flag = f"--{name.replace('_', '-')}"
+        if value is True:
+            argv.append(flag)
+        elif isinstance(value, list):
```
hnimra-amd, line 184: List-valued options used to repeat the flag per element (`--middleware a --middleware b`). `serialize_cli_options` now emits `--middleware a b`.

That matches `nargs="+"` (e.g. `--served-model-name`), but vLLM `--middleware` is `action="append"` and expects multiple `--middleware` arguments. The ray-backend unit test was updated to assert the new encoding using `middleware` as the example, so it no longer guards the append case.

Please restore per-element flags, or special-case known append options. Otherwise a list `middleware` (and similar append flags) will fail at serve launch.

cvs/lib/inference/vllm_job.py
```diff
@@ -559,28 +539,20 @@ def run_client(self):
...
             self.out_dir,
             "--result-filename",
             "results",
         ]
-        # The bench client loads the tokenizer from --model to count tokens. Some
-        # models (e.g. Kimi-K2.6) ship a custom tokenizer via tokenizer_config
-        # auto_map, which transformers refuses to load without trust-remote-code.
-        # Mirror the server's setting so the client can load the same tokenizer.
-        if self.serve_args.get("trust-remote-code") is True:
+        if self.variant.benchmark_params.ignore_eos:
```
hnimra-amd: Contract hole, not a packaged-catalog break: `--ignore-eos` and `--trust-remote-code` are read from `self.variant.benchmark_params` (the config defaults), while `num_prompts` / `random_range_ratio` / prefix come from the cell `benchmark_params` passed into `VllmJob`.

Sweep overrides of those two typed fields are merged in `resolved_runs()`, then dropped from `benchmark_options` because they are model fields, so they never reach the bench command. Current catalog JSON sets them on top-level `benchmark_params`, so those files still work.

If sweeps are meant to override all `benchmark_params`, please drive both flags from the same resolved map as `num_prompts`.

cvs/input/config_file/inference/vllm/README.md
```diff
@@ -117,15 +145,15 @@ Every environment-specific value is redacted. Per config:
...
-| `model.id` | Local model path (e.g. `/models/GLM-5.1-FP8`) or an HF repo id |
+| `server_params.model` | Local model path (e.g. `/models/GLM-5.1-FP8`) |
 | `container.image` | The vLLM/ROCm image tag under test |
 | `container.runtime.args.volumes[1]` | Replace `<changeme-models-mount>` with the host models directory |
-| `roles.server.ib_netdev` | *(distributed only)* socket interface name for `NCCL_SOCKET_IFNAME` / `GLOO_SOCKET_IFNAME` / `TP_SOCKET_IFNAME`. Must be **UP and hold a routable IPv4 reaching the other node** — check `ip -o -4 addr show`, not just `ip -o link show`. An interface that exists but is DOWN/addressless fails engine init with gloo `Unable to find address for: <name>` |
-| `params.master_addr` | *(distributed only)* head node IP |
+| `ib_netdev` | *(distributed only)* socket interface name for `NCCL_SOCKET_IFNAME` / `GLOO_SOCKET_IFNAME` / `TP_SOCKET_IFNAME`. Must be **UP and hold a routable IPv4 reaching the other node** — check `ip -o -4 addr show`, not just `ip -o link show`. |
 
 `paths.models_dir` is `/models`, the in-container mount point — it is exported
 as `HF_HUB_CACHE`. When `model.id` is an absolute path under `/models`, vLLM
```
hnimra-amd: Leftover `model.id` here — this should be `server_params.model` to match the table above and the new schema.

cvs/lib/inference/utils/vllm_config_loader.py
```diff
@@ -1,253 +1,322 @@
...
+        _validate_cli_option_map(self.extra_options(), section="benchmark_params")
+        return self
 
-class VariantConfig(_Forbid):
-    """Unified typed config for both single-node and distributed vllm runs.
 
-    Standalone (does not extend BaseVariantConfig) so it can be constructed
-    without the threshold_json field the base requires — absent from unit-test
-    fixtures. Production configs always supply it via substitute_config.
+_BENCHMARK_RESERVED = {
```
hnimra-amd, line 144: `_BENCHMARK_RESERVED` (including `percentile_metrics` / `metric_percentiles`) is enforced on `sweeps` only. The same names as extras on `benchmark_params` still serialize after the harness-owned flags in `run_client`, so you can get two `--percentile-metrics` on the bench command.

Low severity — packaged configs do not do this — but please reject those names on `benchmark_params` extras the same way as on sweeps, since CVS owns percentile reporting.

cvs/input/config_file/inference/vllm/mi3xx_vllm_gpt-oss-120b_mxfp4_distributed.json
```diff
@@ -1,30 +1,32 @@
 {
-  "_comment": "vllm distributed workload: gpt-oss-120b_mxfp4 on MI325X, TP=4, PP=2 across 2 nodes. Sweep carries 1k1k/1k8k/8k1k at concurrencies 16 and 32; only 1k1k is selected in sweep.runs. num_prompts is 320, not the 3200 schema default, to keep this a characterization pass rather than a full-length benchmark -- raise it to 3200 for comparable published numbers. Set container.image, model.id and the models mount before running.",
-  "schema_version": 1,
-  "framework": "vllm",
+  "_comment": "vLLM distributed workload. Set container.image, server_params.model, ib_netdev, and the models mount before running.",
   "enforce_thresholds": false,
   "threshold_json": "mi325x_vllm_gpt-oss-120b_mxfp4_distributed_threshold.json",
+  "_comment_network": "HCA devices may be discovered automatically; ib_netdev selects the distributed socket interface.",
+  "ib_hca_devices": "auto",
+  "ib_netdev": "<changeme>",
```
amd-droy: this will be going inside container --> env, like this 
"container": {
    "lifetime": "per_run",
    "name": "sglang_container",
    "_image_example": "rocm/sgl-dev:v0.5.14-rocm720-mi30x-20260629",
    "image": "<changeme>",
    "runtime": {
    ...
        "devices": ["/dev/dri", "/dev/kfd", "/dev/infiniband/rdma_cm"],
        "env": {
          "NCCL_IB_HCA": "<changeme> rdma0,rdma1,rdma2,rdma3,rdma4,rdma5,rdma6,rdma7",
          "HCA_ID_PREFIX": "<changeme>",
          "NCCL_SOCKET_IFNAME": "<changeme> eno0",
          "GLOO_SOCKET_IFNAME": "<changeme> eno0",
          "NCCL_IB_GID_INDEX": "3",
          "NCCL_DEBUG": "ERROR",
       ...
        }
      }
    }

cvs/input/config_file/inference/vllm/mi3xx_vllm_gpt-oss-120b_mxfp4_distributed.json
```diff
@@ -1,30 +1,32 @@
...
-  },
   "container": {
+    "_comment": "Container definition and static environment inherited by server, benchmark, and accuracy commands.",
     "lifetime": "per_run",
     "name": "vllm_gpt-oss-120b_mxfp4_distributed_mi325x",
     "image": "<changeme>",
+    "env": {
+      "HF_HUB_OFFLINE": "1",
+      "TRANSFORMERS_OFFLINE": "1"
+    },
```
amd-droy, line 28: you can remove this if not used anywhere - "HF_HUB_OFFLINE": "1",
      "TRANSFORMERS_OFFLINE": "1"

cvs/input/config_file/inference/vllm/mi3xx_vllm_mistral-large-3_bf16_single.json
```diff
@@ -33,84 +35,53 @@
...
-      },
-      {
-        "combo": "w_mistral-large-3_bf16_1k1k",
-        "concurrency": 32
-      }
-    ]
+  "_comment_sweeps": "Canonical run cells; entries override benchmark_params for that cell.",
+  "sweeps": {
+    "ISL=1024,OSL=1024,TP=8,PP=1,CONC=16": {},
+    "ISL=1024,OSL=1024,TP=8,PP=1,CONC=32": {}
```
amd-droy, line 75: sweeps will have all the combinations, conc=4,8,16,32,64,128,256 and runs will have only those which we want to run

cvs/input/config_file/inference/vllm/mi3xx_vllm_mistral-large-3_bf16_single.json
```diff
@@ -33,84 +35,53 @@
...
   },
+  "_comment_runs": "Explicit ordered run cells to execute.",
+  "runs": [
+    "ISL=1024,OSL=1024,TP=8,PP=1,CONC=16",
+    "ISL=1024,OSL=1024,TP=8,PP=1,CONC=32"
+  ],
   "accuracy": {
     "tasks": []
   }
```
amd-droy: are the accuracy arguments hardcoded inside the libraries? Not needed for this release, but you can bring them here in a later PR.

cvs/input/config_file/inference/vllm/mi3xx_vllm_mistral-large-3_bf16_single.json
atnair-amd: not keeping those in since i havent validated them.

cvs/lib/inference/utils/vllm_config_loader.py
atnair-amd: how often is --middleware used? i dont think we have a single config where we are explicitly using middleware.

cvs/input/config_file/inference/vllm/mi3xx_vllm_deepseek-v4-flash_fp8_distributed.json
atnair-amd: keeping these in to force local usage since pulling is not tested

cvs/input/config_file/inference/vllm/mi3xx_vllm_gpt-oss-120b_mxfp4_distributed.json
atnair-amd: keeping as it forces offline path

hnimra-amd (approved):
lgtm!

PR #387, opened 2026-09-06
Sglang/xdit documentation and sglang config final updates

amd-droy:
**Summary:**
Migrate SGLang single, distributed, and disaggregated suites from the legacy config / named benchmark_params layout to the unified workload schema (paths, container, server_params, benchmark_params, accuracy, sweeps / sweep). Update the loader, libraries, tests, and docs so those templates actually run, including sweep cell selection and container launch env.

**Config templates:**
Replace mi30x_*   workloads with mi3xx_sglang_{llama_70b,deepseek_r1_0528}_{single,distributed,disaggregated}.json.

Each template now has:
enforce_thresholds (top-level; false records perf metrics without failing SLOs)
threshold_json → mi325_sglang_*_threshold.json beside the config
paths, container.runtime.args (volumes as host:container[:opts], env with NCCL/Gloo/ADD_EXPORT_ENV)
server_params (model, TP/PP, role hosts, ports)
flat benchmark_params and accuracy.tasks
sweeps (per-combo overrides) and sweep.runs (which cells to parametrize)
Removed unused/legacy fields from these templates: named variants, schema_version / roles / params.inference_tests, nic_type, NCCL_IB_HCA_LIST / nccl_ib_hca_list, duplicate enforce_thresholds under bench params.

Threshold files kept for Llama 70B and DeepSeek-R1-0528 (mi325_*). Dropped unused GPT-OSS / GLM / Kimi threshold JSONs from the suite set.

**Loader and runtime:**
sglang_config_loader.py builds orchestrator/controller views from the unified schema; still loads legacy configs when present.
sweep.runs empty → every performance cell in the threshold JSON; non-empty → listed combos only.
Combo matching: exact ISL,OSL,TP,PP,CONC first, then unique ISL,OSL,CONC if TP/PP differs (threshold TP/PP is descriptive).
Server TP/PP stay on server_params; sweep cells only change ISL/OSL/CONC and sweeps overrides (num_prompts).
Strip runtime.args.env from the orchestrator container block and skip aggregate values (ADD_EXPORT_ENV) so Docker gets scalar -e KEY=VALUE only. List entries are expanded into real env keys and into launch-script export lines.

**Libraries / tests:**
Disagg/distributed IBV setup always runs the HCA check; GID index comes from config, not a hardcoded 3.
Perf tests apply perf_cell params per selected sweep cell.
Drop the ISL/OSL/TP/PP console summary table; keep the detailed performance results table.
Unit tests for loader, sweep selection, packaged configs, and env flattening.

**Docs:**
docs/reference/configuration-files/inference/sglang.rst
docs/how-to/test-suites/inference/sglang.rst
xDiT how-to/reference updates and removal of standalone flux1_t2i.rst / wan22_i2v.rst (folded into xDiT docs). 
[cvs_results-diagg-final.zip](https://github.com/user-attachments/files/31970668/cvs_results-diagg-final.zip)


docs/reference/configuration-files/inference/index.rst
```diff
@@ -10,8 +10,7 @@ JSON configuration schemas for inference and serving benchmarks under ``cvs/inpu
 
 - :doc:`ATOM (vLLM Benchmarking) </reference/configuration-files/inference/atom>` — ATOM inference benchmarks
 - :doc:`vLLM Inference </reference/configuration-files/inference/vllm>` — vLLM serving throughput and latency
-- :doc:`SGLang Disaggregated Prefill-Decode </reference/configuration-files/inference/sglang>` — disaggregated LLM serving
-- :doc:`Flux.1 Text-to-Image </reference/configuration-files/inference/flux1_t2i>` — xDiT text-to-image generation
-- :doc:`WAN 2.2 Image-to-Video </reference/configuration-files/inference/wan22_i2v>` — xDiT image-to-video generation
+- :doc:`SGLang Inference </reference/configuration-files/inference/sglang>` — single-node, distributed, and disaggregated LLM serving
+- :doc:`xDiT Inference </reference/configuration-files/inference/xdit>` — FLUX.1/FLUX.2 text-to-image and WAN 2.2 image-to-video
```
solaiys, line 14: **Blocking: Sphinx TOC still points at the deleted Flux/WAN pages.**

This index (and the how-to inference table) correctly switch to `:doc:` … `/xdit`, but `docs/sphinx/_toc.yml.in` was not updated. It still has:

```yaml
- file: reference/configuration-files/inference/sglang
  title: SGLang Disaggregated
- file: reference/configuration-files/inference/flux1_t2i
  title: Flux.1 Text-to-Image
- file: reference/configuration-files/inference/wan22_i2v
  title: WAN 2.2 Image-to-Video
```

This PR deletes `flux1_t2i.rst` / `wan22_i2v.rst` and adds `xdit.rst`, so the docs build will fail on missing TOC entries.

Replace those two files with `reference/configuration-files/inference/xdit` and retitle the SGLang entry to “SGLang Inference” to match this page.

cvs/lib/inference/sglang/sglang_distributed_lib.py
```diff
@@ -332,14 +326,11 @@ def install_container_packages(self) -> None:
...
-        if re.search('broadcom|thor', self.nic_type, re.I):
-            self.inf_dict['nccl_ib_gid_index'] = 3
-            cmd = "bash -c " + shlex.quote(f"cp {self.mount_vol}.host {self.mount_vol}; sleep 2; ibv_devinfo; sleep 2;")
-            out_dict = self._container_exec(cmd)
-            hca_id_regex = rf'hca_id:\s+{re.escape(self.hca_id_prefix)}'
-            for node, out in out_dict.items():
-                if not re.search(hca_id_regex, out or '', re.I):
-                    fail_test(f'Broadcom libbnxt rdma driver is not properly copied on node {node}')
+        out_dict = self._container_exec("ibv_devinfo")
+        hca_id_regex = rf'hca_id:\s+{re.escape(self.hca_id_prefix)}'
```
solaiys, line 330: **Blocking: this check now always runs, but the shipped `HCA_ID_PREFIX` is not an `ibv_devinfo` `hca_id`.**

Previously this path only ran for Broadcom/Thor (`nic_type`), copied `libbnxt`, and forced GID 3. Now every distributed/disagg job greps `hca_id:\s+{hca_id_prefix}` and `test_setup_ibv_devices` is uncommented, so a mismatch fails the suite.

`ibv_devinfo` reports `hca_id: bnxt_re0` / `mlx5_0`. The new configs put NCCL netdev names in `HCA_ID_PREFIX` (`"<changeme> rdma"`, and DeepSeek distributed `"rdma <changeme>"`). `rdma0` is `NCCL_IB_HCA`, not `hca_id`.

Set `HCA_ID_PREFIX` to the real prefix (`bnxt_` / `mlx5_`, matching the `bnxt_` default in `_apply_inf_defaults`) and keep `rdma0,rdma1,...` on `NCCL_IB_HCA` only. Same pattern on the disagg copy of this function.

cvs/input/config_file/inference/sglang/mi3xx_sglang_deepseek_r1_0528_distributed.json
```diff
@@ -0,0 +1,137 @@
...
+          "/mnt/dtni/models:/root/models",
+          "/dev/infiniband:/dev/infiniband",
+          "/usr/local/lib/libbnxt_re-rdmav34.so:/usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so:ro",
+          "/usr/lib/x86_64-linux-gnu/libibverbs.so.1:/usr/lib/x86_64-linux-gnu/libibverbs.so.1:ro",
+          "/lib/libibverbs.d:/lib/libibverbs.d"
+        ],
+        "devices": ["/dev/dri", "/dev/kfd", "/dev/infiniband/rdma_cm"],
+        "env": {
+          "NCCL_IB_HCA": "<changeme> rdma0,rdma1,rdma2,rdma3,rdma4,rdma5,rdma6,rdma7",
+          "HCA_ID_PREFIX": "rdma <changeme>",
```
solaiys: **Blocking: `HCA_ID_PREFIX` placeholder is reversed vs the other workloads, and the hint is the wrong identifier.**

Llama dist/disagg and DeepSeek disagg use `"<changeme> rdma"`. This file uses `"rdma <changeme>"`. After a naive `<changeme>` replace that becomes `rdma <host>`, which will not match `ibv_devinfo` `hca_id:` (and `test_setup_ibv_devices` now fails the run).

Use a single convention on all dist/disagg configs: `HCA_ID_PREFIX` is an `ibv_devinfo` prefix (`"<changeme>"` or `"<changeme> bnxt_"`), not an NCCL netdev list.

cvs/tests/inference/sglang/sglang_distributed.py
```diff
@@ -51,12 +51,12 @@ def test_launch_container(orch, variant_config, lifecycle, request):
     lifecycle.complete_stage(request, "container_launch", t0)
 
 
-# def test_setup_ibv_devices(im_obj, lifecycle, request):
-#     globals.error_list = []
-#     t0 = time.monotonic()
-#     im_obj.exec_nic_setup_scripts()
-#     im_obj.check_ibv_devices()
-#     lifecycle.complete_stage(request, "ibv_setup", t0)
+def test_setup_ibv_devices(im_obj, lifecycle, request):
```
solaiys, line 54: **Blocking (same IB prefix issue): this test is now live.**

It calls `exec_nic_setup_scripts()` then `check_ibv_devices()`. The first now hard-fails unless `HCA_ID_PREFIX` appears in `ibv_devinfo` `hca_id:` on every server node. With the shipped `rdma` example (and DeepSeek distributed’s reversed placeholder) this stage fails before launch. Same for `sglang_disagg_distributed.py`.

Fix the config prefixes (or the matcher) before leaving this uncommented.

solaiys (changes requested):
## Review: blocking issues

Loader/sweep matching, env flattening (`ADD_EXPORT_ENV` → scalar `-e` + launch-script exports), and the packaged `mi3xx_*` workload tests look consistent. Two things will break merge/docs or a real distributed/disagg run.

### 1. Sphinx TOC still lists the deleted Flux/WAN pages

`docs/sphinx/_toc.yml.in` still has `reference/configuration-files/inference/flux1_t2i` and `wan22_i2v`, and still titles SGLang as “Disaggregated”. This PR deletes those RST files and adds `xdit.rst`, and the inference index/how-to tables already point at xDiT. The sidebar TOC was not updated, so the docs build will fail on missing files.

Fix: in `_toc.yml.in`, replace the flux/wan entries with `reference/configuration-files/inference/xdit` and retitle the SGLang entry to match the new index.

### 2. Newly enabled `test_setup_ibv_devices` will fail with the shipped HCA prefix examples

`exec_nic_setup_scripts` now always greps `ibv_devinfo` for `hca_id: {HCA_ID_PREFIX}` (no longer Broadcom-only, and it no longer copies `libbnxt`). `test_setup_ibv_devices` is uncommented for distributed and disagg.

Shipped `HCA_ID_PREFIX` values are NCCL netdev hints (`rdma`), not `ibv_devinfo` `hca_id` prefixes (`bnxt_re0` / `mlx5_0`). DeepSeek distributed also reverses the placeholder (`"rdma <changeme>"` vs `"<changeme> rdma"` on the other files). Following the in-string example fails the new IB setup test on every node.

Fix: make `HCA_ID_PREFIX` a real `ibv_devinfo` prefix (`<changeme>` or `bnxt_` / `mlx5_`), keep `rdma0,...` only on `NCCL_IB_HCA`, and use the same placeholder convention on all six dist/disagg configs.

docs/reference/configuration-files/inference/index.rst
amd-droy: Addressed. Updated docs/sphinx/_toc.yml.in.

cvs/input/config_file/inference/sglang/mi3xx_sglang_deepseek_r1_0528_distributed.json
amd-droy: addressed.

cvs/lib/inference/sglang/sglang_distributed_lib.py
amd-droy: Addressed -  "HCA_ID_PREFIX": "<changeme> rdma, bnxt_, or mlx5_",

cvs/tests/inference/sglang/sglang_distributed.py
amd-droy: fixed, it can now choose from  " rdma, bnxt_, or mlx5_". Most of the times, _bnxt does not show in ibv_devinfo.

docs/reference/configuration-files/inference/index.rst
solaiys: Confirmed on `e3b4d1c30e`: `_toc.yml.in` now points at `xdit` and titles SGLang “SGLang Inference”. This one is fixed.

cvs/input/config_file/inference/sglang/mi3xx_sglang_deepseek_r1_0528_distributed.json
solaiys: The reversed `"rdma <changeme>"` is gone and all four dist/disagg files now use the same string. That part is fixed.

The remaining blocker is that the new value is still not a single `ibv_devinfo` `hca_id` prefix — see the follow-up on `exec_nic_setup_scripts`.

cvs/lib/inference/sglang/sglang_distributed_lib.py
solaiys: This is not addressed. Putting `"<changeme> rdma, bnxt_, or mlx5_"` in JSON does not make the code choose among those strings.

`exec_nic_setup_scripts` is unchanged:

```python
hca_id_regex = rf'hca_id:\s+{re.escape(self.hca_id_prefix)}'
```

`self.hca_id_prefix` is the entire env value. After deleting `<changeme>`, that is the literal `rdma, bnxt_, or mlx5_`, so the grep looks for `hca_id: rdma, bnxt_, or mlx5_`, which `ibv_devinfo` never prints.

The reference doc already says this must be a **single** prefix. Make the JSON match that:

```json
"HCA_ID_PREFIX": "<changeme>"
```

with a `_comment` like `ibv_devinfo hca_id prefix (bnxt_ or mlx5_), not the NCCL_IB_HCA list`. Do not list `rdma` here — that belongs on `NCCL_IB_HCA`.

If you really want auto-detect among several prefixes, the matcher has to split and succeed when **any** of them appears in `ibv_devinfo`. The config string alone does not do that.

cvs/tests/inference/sglang/sglang_distributed.py
solaiys: Same remaining issue: the test still calls `exec_nic_setup_scripts()`, and that function still treats `HCA_ID_PREFIX` as one regex. It cannot “choose from rdma, bnxt_, or mlx5_” until the matcher (or the JSON value) is a single real `hca_id` prefix.

cvs/input/config_file/inference/sglang/mi3xx_sglang_llama_70b_distributed.json
```diff
@@ -0,0 +1,137 @@
...
+          "/mnt/dtni/models:/root/models",
+          "/dev/infiniband:/dev/infiniband",
+          "/usr/local/lib/libbnxt_re-rdmav34.so:/usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so:ro",
+          "/usr/lib/x86_64-linux-gnu/libibverbs.so.1:/usr/lib/x86_64-linux-gnu/libibverbs.so.1:ro",
+          "/lib/libibverbs.d:/lib/libibverbs.d"
+        ],
+        "devices": ["/dev/dri", "/dev/kfd", "/dev/infiniband/rdma_cm"],
+        "env": {
+          "NCCL_IB_HCA": "<changeme> rdma0,rdma1,rdma2,rdma3,rdma4,rdma5,rdma6,rdma7",
+          "HCA_ID_PREFIX": "<changeme> rdma, bnxt_, or mlx5_",
```
solaiys: **Still blocking:** this is a prose list, not a prefix the matcher can use.

`<changeme> <default>` means the remainder is the runtime value. After deleting the token this becomes `rdma, bnxt_, or mlx5_`, and `hca_id_regex = rf'hca_id:\s+{re.escape(self.hca_id_prefix)}'` looks for that entire literal.

Use a single `ibv_devinfo` `hca_id` prefix, same on all four dist/disagg configs:

```json
"HCA_ID_PREFIX": "<changeme>"
```

Leave `rdma0,rdma1,...` on `NCCL_IB_HCA`. `rdma` is not a typical `hca_id:` value (`bnxt_re0` / `mlx5_0` are).

solaiys (changes requested):
## Re-review

TOC and the reversed DeepSeek placeholder are fixed.

The HCA change is not. `"HCA_ID_PREFIX": "<changeme> rdma, bnxt_, or mlx5_"` is still one string, and `exec_nic_setup_scripts` still does `re.escape(self.hca_id_prefix)` against `hca_id:`. After stripping `<changeme>` the grep looks for `hca_id: rdma, bnxt_, or mlx5_`, which `ibv_devinfo` never prints. The matcher was not updated to choose among prefixes.

Please set `HCA_ID_PREFIX` to a single real prefix (`"<changeme>"` or `"<changeme> mlx5_"` / `bnxt_`) and keep `rdma0,...` on `NCCL_IB_HCA` only — matching the reference doc on this PR.

cvs/input/config_file/inference/sglang/mi3xx_sglang_llama_70b_distributed.json
solaiys: @amd-droy ,  Do you expect the user to fill this HCA_ID_PREFIX ?
if yes, you can ignore this comment i think.

But if its a PREFIX string of the ib devices name ( rdma0 or bnxt_re0 or mlx5_0 ) from the NCCL_IB_HCA list, we can do it programmatically right?

cvs/lib/inference/sglang/sglang_distributed_lib.py
amd-droy: updated - "HCA_ID_PREFIX": "<changeme>"

cvs/input/config_file/inference/sglang/mi3xx_sglang_deepseek_r1_0528_distributed.json
amd-droy: updated - "HCA_ID_PREFIX": "<changeme>"

cvs/tests/inference/sglang/sglang_distributed.py
amd-droy: done

solaiys:
LGTM.
Please Check with Ignatious before merging it.

solaiys (approved):
Previous blockers are resolved on `0d671c1c54`:

- Sphinx TOC lists xDiT and titles SGLang correctly.
- Dist/disagg `HCA_ID_PREFIX` is a single `"<changeme>"` for the user to replace with their `ibv_devinfo` `hca_id` prefix (`bnxt_` / `mlx5_`). `NCCL_IB_HCA` still holds the `rdma0,...` list.

No remaining blocking issues from my side.

cvs/lib/inference/sglang/sglang_config_loader.py
```diff
@@ -343,19 +462,151 @@ def _sync_legacy_inference_container_name(self):
...
+        inference_tests[task_id] = task
+    return inference_tests
+
+
+_RUNTIME_ENV_TO_INFERENCE = {
+    "NCCL_IB_HCA": "nccl_ib_hca",
+    "HCA_ID_PREFIX": "hca_id_prefix",
+    "NCCL_SOCKET_IFNAME": "nccl_socket_ifname",
+    "GLOO_SOCKET_IFNAME": "gloo_socket_ifname",
+    "GLOO_TCP_IFNAME": "gloo_tcp_ifname",
```
atnair-amd: suggestion (non-blocking): `GLOO_TCP_IFNAME` is loaded but the launch scripts never read it

This mapping stores `runtime.args.env.GLOO_TCP_IFNAME` as `inference['gloo_tcp_ifname']`, and Docker `-e` will set the variable. `SglangDistributed._server_env_body` / `SglangDisaggPD` setup still do `export GLOO_TCP_IFNAME={self.inf_dict['gloo_socket_ifname']}` (untouched in this PR), so sourcing `/tmp/server_env_script.sh` overwrites it.

Attack: `GLOO_SOCKET_IFNAME=eno0` and `GLOO_TCP_IFNAME=ib0` in a dist/disagg template.
Expected: Gloo TCP uses `ib0`.
Actual: after `setup_server_container_env`, `GLOO_TCP_IFNAME=eno0`.

Today’s templates set both to the same `eno0`, so runs are fine. If the field is meant to be independent, export `gloo_tcp_ifname` (fallback to socket) in those four env bodies; otherwise drop `GLOO_TCP_IFNAME` from the templates/docs so it is not a no-op knob.

atnair-amd (approved):
Is there a validation run linked (in a ticket or elswhere), othw LGTM

amd-droy:
> Is there a validation run linked (in a ticket or elswhere), othw LGTM

Good idea. Just attached it. @atnair-amd, can you please have a look.

[cvs_results-diagg-final.zip](https://github.com/user-attachments/files/31971454/cvs_results-diagg-final.zip)

cvs/input/config_file/inference/sglang/mi3xx_sglang_llama_70b_single.json
```diff
@@ -0,0 +1,124 @@
...
+        "privileged": true,
+        "shm_size": "128G",
+        "volumes": [
+          "/home/{user-id}:/home/{user-id}",
+          "/mnt/dtni/models:/root/models"
+        ],
+        "devices": ["/dev/dri", "/dev/kfd"],
+        "env": {
+          "NCCL_DEBUG": "ERROR",
+          "ADD_EXPORT_ENV": ["SGLANG_USE_AITER=1", "AMDGCN_USE_BUFFER_OPS=1", "ROCM_QUICK_REDUCE_QUANTIZATION=INT8"]
```
cijohnson: this should be

"env": {
          "NCCL_DEBUG": "ERROR",
         "SGLANG_USE_AITER": "1",
         "AMDGCN_USE_BUFFER_OPS": "1",
        "ROCM_QUICK_REDUCE_QUANTIZATION": "INT8"
        }

cvs/input/config_file/inference/sglang/mi3xx_sglang_llama_70b_single.json
```diff
@@ -0,0 +1,124 @@
...
+        "limit": "100",
+        "num_concurrent": "4",
+        "exec_timeout_sec": "7200",
+        "extra_model_args": "tokenizer_backend=huggingface"
+      }
+    ]
+  },
+  "sweeps": {
+    "ISL=1024,OSL=1024,TP=8,PP=1,CONC=4": {
+      "num_prompts": "50"
```
cijohnson: why num_prompts i set to 50? i see bench mark params section is defaulting it to 50

only if user needs to experiment by changing defaults then they can set num_prompts or random_range_ratio etc in sweep

so we can add it as a comment, no need to set 50

cijohnson (changes requested):
@amd-droy, I have added two comments, which is applicable to all the configs in this PR

cvs/input/config_file/inference/sglang/mi3xx_sglang_llama_70b_single.json
amd-droy: @cijohnson , I had to keep this part flexible, later if I need to add/remove any of them, removing it from this list will suffice. I do not need to touch the code. But, if I add then as separate env variable, I need to update the code. 

The way the list is working now, is like it will get each argument from the list and export them in the container env. Let me know your thoughts.

cvs/input/config_file/inference/sglang/mi3xx_sglang_llama_70b_single.json
amd-droy: removed the overriding of num_prompts.

cvs/input/config_file/inference/sglang/mi3xx_sglang_llama_70b_single.json
amd-droy: all configs updated.

cijohnson (approved):

PR #388, opened 2026-09-08
(jaxmaxtext): new config layout, DeepSeek-V4-284B, docs update

solaiys:
## Overview
Migrates the JAX MaxText training suites to a new, externalized config layout, adds the DeepSeek-V4-284B config, refreshes the docs, and finalizes the shipped config set. 5 commits; ~43 files.
### 1. New config layout + code support
`feat: migrate training configs to the new layout + code support`
- On-disk configs restructured: CVS params at the root, a `container` block whose `env{}` is forwarded verbatim via `docker run -e` (no more env script), a `train_params` block (`hf_model_id`, `train_script_paths`, `maxtext_config` passthrough, structured `xla_flags`), root-level test blocks, and a `sweeps{name: overrides}` map + `runs` selector.
- Loader normalization (`normalize_training_config`): infers `distributed` from the NCCL IB device vars in `container.env`, folds `xla_flags` into a single `XLA_FLAGS`, reads `steps` / `enable_checkpointing` / `tokenizer_path` from `maxtext_config`, and hard-exits on any unresolved `<changeme>` in `container.env`.
- Training lib: per-node logs + per-node dmesg capture, per-node output-dir handling (fixes NFS cross-client cache races), and always-on fatal-error signatures (JAX distributed fatal, glog FATAL).
### 2. DeepSeek-V4-284B
`feat: add DeepSeek-V4-284B config with MaxText branch checkout + install`
- New `mi325x_jaxmaxtext_deepseek-v4-284b` config with post-launch `maxtext_branch` checkout (`release/v26.7`) + `maxtext_install_cmd`.
### 3. Legacy JAX config files
`fix: adding the legacy jax test config files`
### 4. Docs
`docs: update rst files and re-direct README to the rst files`
- Full rewrite of the config reference and how-to to the new layout; available-configs table; README files reduced to redirects (single source of truth in `docs/`).
### 5. Config polish
`fix: configs update - image <changeme> tag, smoke blocks, etc.`
- `enforce_thresholds: false`, `NNODES` removed from env, `env{}` / `maxtext_config{}` grouped in a uniform order, single-line threshold entries.
- `container.image` carries a `<changeme>` tag; `smoke{}` present in all configs (added where missing); `checkpoint_resume{}` everywhere.
- Corrected `hf_model_id` / `tokenizer_path`: llama-3.1-405b → `NousResearch/Meta-Llama-3.1-405B`, llama-3.1-70b → `NousResearch/Meta-Llama-3.1-70B`.
## Config set
Ships 12 configs (+ thresholds) for MI300X, MI325X, and MI35X. The `mi355x_*` configs are removed pending on-hardware validation and will be re-added later.
## Setup notes for users
Replace `<changeme>` in `container.image` (every config) and in the `container.env` NCCL/RDMA fields (distributed configs) before running.
## Validation
- All configs load with full sweep/threshold coverage
- make ut passed. ✅

Ran 1736 tests in 13.234s
OK (skipped=1)
Exit code 0.
- Docs build clean (no warnings in the changed files)


cvs/tests/training/jaxmaxtext/_common.py
```diff
@@ -197,6 +196,15 @@ def launch_container(orch, variant_config, lifecycle, request):
         lifecycle.failed = True
         pytest.fail(f"container {name} not running after setup_containers()")
 
+    # Optional: check out a specific MaxText branch inside the freshly launched
+    # container (no-op unless training.maxtext_branch is set). Do it here, once,
+    # so every downstream stage runs against the requested code.
+    try:
+        MaxTextTrainingJob(orch, variant_config, hf_token="").checkout_maxtext_branch()
```
amd-droy, line 203: if orch.exec returns no hosts, both loops are no-ops and checkout/install are treated as success. Assert every orch.hosts entry is present, and fail on non-zero exit, not only on stdout.

cvs/input/config_file/training/jaxmaxtext/mi300x_jaxmaxtext_deepseek-v2-lite_distributed.json
```diff
@@ -40,89 +34,29 @@
...
+    "_env_comment": "Static ENV exported into the container on every node. Per-node/dynamic vars (JAX_COORDINATOR_IP, NNODES, NODE_RANK, JAX_PROCESS_INDEX) and credentials (HF_TOKEN, HF_HOME, LD_LIBRARY_PATH, PYTHONPATH) are injected by the launcher and must NOT be set here. NCCL_IB_*/*_SOCKET_IFNAME/NCCL_IB_GID_INDEX are the cluster-specific RDMA/NIC device selection (shipped as <changeme>; distributed runs hard-exit at config load until set). JAX_COORDINATOR_PORT / JAX_DISTRIBUTED_* are the JAX coordinator knobs.",
+    "env": {
       "GPU_MAX_HW_QUEUES": "2",
       "HSA_FORCE_FINE_GRAIN_PCIE": "1",
       "HIP_FORCE_DEV_KERNARG": "1",
       "HSA_NO_SCRATCH_RECLAIM": "1",
       "XLA_PYTHON_CLIENT_MEM_FRACTION": "0.97",
+
+      "NCCL_IB_HCA_LIST": "rdma0,rdma1,rdma2,rdma3,rdma4,rdma5,rdma6,rdma7 <changeme>",
+      "NCCL_IB_HCA": "rdma0,rdma1,rdma2,rdma3,rdma4,rdma5,rdma6,rdma7 <changeme>",
```
amd-droy, line 46: Are NCCL_IB_HCA_LIST and NCCL_IB_HCA both required? In sglang, I found only NCCL_IB_HCA is used and the _LIST is unused.

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
```diff
@@ -360,45 +431,26 @@ def build_training_cmd(self):
...
-                        "export NODE_RANK=0",
-                        "export JAX_PROCESS_INDEX=0",
-                    ]
-                )
-
-            launcher_lines.append("export PYTHONPATH=$PYTHONPATH:/workspace/maxtext/")
-            log_file = f"{self.out_dir}/out-node{i}/training.log"
+            log_file = self._node_log(i)
             launcher_lines.append(
                 f"cd /workspace/maxtext && python {shlex.quote(train_script)} "
```
amd-droy: checkout_maxtext_branch() uses training.maxtext_root, but the launcher always cds / sets PYTHONPATH to /workspace/maxtext. A custom maxtext_root checks out tree A and trains the image tree. Wire these paths (and train_script_paths) to maxtext_root, or drop the field.

cvs/lib/training/jaxmaxtext/utils/training_config_loader.py
```diff
@@ -300,16 +312,152 @@ def enabled_sweeps(self):
...
+    quotes so it survives as ONE ``docker run -e XLA_FLAGS="..."`` token: the
+    orchestrator emits ``-e KEY=VALUE`` unquoted and the command is evaluated by a
+    remote shell, so an unquoted spaced value would split into separate tokens.
+    Returns ``""`` for an empty map -- the caller then omits ``XLA_FLAGS`` entirely
+    rather than exporting an empty one that would clobber XLA's own defaults.
+    """
+    if not xla_flags:
+        return ""
+    joined = " ".join(f"--{k}={v}" for k, v in xla_flags.items())
+    return f'"{joined}"'
```
amd-droy, line 336: Don’t bake " into XLA_FLAGS. Quote at the docker -e layer so the value inside the container is --flag=... with no extra quotes.

solaiys: XLA_FLAGS itself a env variable with space separated multiple option&value pairs. this XLA_FLAGS is then set to the docker's -e layer only.

**example :**
cmd = sudo -n **docker run** -d --name rocm-jaxmaxtext-mi325x-llama3.1-8b -v /home/ssolaiya@amd.com/.ssh:/host_ssh -v /home/ssolaiya@amd.com:/home/ssolaiya@amd.com -v /dev/infiniband:/dev/infiniband -v /usr/local/lib/libbnxt_re-rdmav34.so:/usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so:ro -v /usr/lib/x86_64-linux-gnu/libibverbs.so.1:/usr/lib/x86_64-linux-gnu/libibverbs.so.1:ro -v /lib/libibverbs.d:/lib/libibverbs.d -v /tmp/ssolaiya@amd.com/jax/TRAINING_LOGS:/workspace/maxtext/output -e GPUS=8 -e MULTINODE=true -e GPU_MAX_HW_QUEUES=2 -e HSA_FORCE_FINE_GRAIN_PCIE=1 -e HIP_FORCE_DEV_KERNARG=1 -e HSA_NO_SCRATCH_RECLAIM=1 -e XLA_PYTHON_CLIENT_MEM_FRACTION=0.97 -e NCCL_IB_HCA_LIST=rdma0,rdma1,rdma2,rdma3,rdma4,rdma5,rdma6,rdma7 -e NCCL_IB_HCA=rdma0,rdma1,rdma2,rdma3,rdma4,rdma5,rdma6,rdma7 -e NCCL_SOCKET_IFNAME=eno0 -e GLOO_SOCKET_IFNAME=eno0 -e NCCL_IB_GID_INDEX=3 -e NCCL_DEBUG=VERSION -e NCCL_IB_DISABLE=0 -e NCCL_PROTO=Simple -e NCCL_IB_TC=41 -e NCCL_IB_SL=0 -e NCCL_CHECKS_DISABLE=1 -e NCCL_CROSS_NIC=0 -e NVTE_ALLOW_NONDETERMINISTIC_ALGO=1 -e NVTE_USE_HIPBLASLT=1 -e NVTE_FUSED_ATTN=1 -e NVTE_CK_USES_BWD_V3=1 -e NVTE_CK_USES_FWD_V3=1 -e NVTE_CK_IS_V3_ATOMIC_FP32=0 -e NVTE_CK_HOW_V3_BF16_CVT=2 -e NVTE_FUSED_ATTN_CK=1 -e NVTE_FUSED_ATTN_AOTRITON=0 -e JAX_COORDINATOR_PORT=12346 -e JAX_DISTRIBUTED_INITIALIZATION_TIMEOUT_SECONDS=1800 -e JAX_DISTRIBUTED_HEARTBEAT_TIMEOUT_SECONDS=900 **_-e XLA_FLAGS="--xla_gpu_memory_limit_slop_factor=95 --xla_gpu_reduce_scatter_combine_threshold_bytes=8589934592 --xla_gpu_enable_command_buffer= --xla_gpu_enable_latency_hiding_scheduler=True --xla_gpu_all_gather_combine_threshold_bytes=8589934592 --xla_gpu_enable_triton_gemm=False --xla_gpu_enable_cublaslt=True --xla_gpu_autotune_level=0 --xla_gpu_enable_all_gather_combine_by_dim=FALSE"_** --device /dev/kfd --device /dev/dri --device /dev/infiniband --cap-add SYS_PTRACE --cap-add IPC_LOCK --cap-add SYS_ADMIN --security-opt seccomp=unconfined --security-opt apparmor=unconfined --group-add video --network host --ipc host --ulimit memlock=-1 --ulimit nofile=65535:65535 --privileged $(for dev in /dev/infiniband/*; do echo -n "--device $dev:$dev "; done) rocm/jax-training:maxtext-v26.6 sleep infinity

cvs/tests/training/jaxmaxtext/_common.py
solaiys: Fixed. Added _exec_all_nodes_or_fail() which runs the command with detailed=True, asserts every host in orch.hosts reported back (an empty/short result — e.g. no reachable nodes — now raises instead of silently passing), and fails on any non-zero per-host exit code — on top of the existing HEAD-match verify and __MAXTEXT_INSTALL_OK__ sentinel. The checkout, verify, and install steps all go through it now. Added test_raises_when_a_node_is_unreachable and test_raises_on_nonzero_exit.

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
solaiys: Fixed. build_training_cmd() now cds to training.maxtext_root and sets PYTHONPATH to it (default unchanged: /workspace/maxtext), so a branch checkout in that tree is the code that actually trains. train_script_paths stay explicit absolute paths (first existing wins) and are expected to live under maxtext_root. Added test_launcher_uses_custom_maxtext_root.

cvs/input/config_file/training/jaxmaxtext/mi300x_jaxmaxtext_deepseek-v2-lite_distributed.json
solaiys: Good catch — NCCL only honors NCCL_IB_HCA; NCCL_IB_HCA_LIST isn't a standard NCCL/RCCL variable and was redundant here. Removed it from all 9 distributed jaxmaxtext configs and from the docs; distributed detection still keys off NCCL_IB_HCA. (The megatron suite has its own separate NCCL_IB_HCA_LIST usage, left untouched.)

cvs/lib/training/jaxmaxtext/utils/training_config_loader.py
```diff
@@ -300,16 +312,152 @@ def enabled_sweeps(self):
         return selected
 
 
+# ---------- config layout normalization ----------
+
+# NCCL IB device keys used to infer whether a config is a multi-node/RDMA run
+# (single-node configs carry no IB device selection).
+_NCCL_IB_DEVICE_KEYS = ("NCCL_IB_HCA", "NCCL_IB_HCA_LIST")
```
atnair-amd: suggestion (non-blocking): drop `NCCL_IB_HCA_LIST` from distributed detection

Attack: `container.env` has only `NCCL_IB_HCA_LIST` (no `NCCL_IB_HCA`).
Expected: after removing `_LIST` from configs/docs, `distributed` is inferred only from `NCCL_IB_HCA` (as the review thread said).
Actual: `any(env.get(k) for k in _NCCL_IB_DEVICE_KEYS)` is still true, so this config is treated as distributed.

Fix: `_NCCL_IB_DEVICE_KEYS = ("NCCL_IB_HCA",)`.

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
```diff
@@ -318,6 +323,95 @@ def setup_rdma_lib(self):
...
+                log.info("[maxtext checkout %s]\n%s", host, text.strip()[-500:])
+
+            # Confirm every node is on the requested branch (detached HEAD or a
+            # failed checkout would report something else).
+            verify = self._exec_all_nodes_or_fail(
+                f"cd {shlex.quote(root)} && git rev-parse --abbrev-ref HEAD", "MaxText branch verify"
+            )
+            for host, text in verify.items():
+                current = text.strip().splitlines()[-1].strip() if text.strip() else ""
+                if current != branch:
```
atnair-amd, line 367: question: does `release/v26.7` stay on a named branch after `git checkout`?

Attack: `training.maxtext_branch` is a tag (or otherwise leaves a detached HEAD).
Expected: verify accepts the checkout that `git checkout` just succeeded.
Actual: `git rev-parse --abbrev-ref HEAD` is `HEAD`, which is `!= branch`, so the DSv4 launch fails after a successful checkout.

If the image's MaxText repo has a real branch named `release/v26.7`, this is fine. If it is a tag, compare `git rev-parse HEAD` to `git rev-parse <ref>^{commit}` instead of `--abbrev-ref`.

atnair-amd (approved):
The new on-disk layout (root CVS params, `container.env` → `docker run -e`, `train_params` + `sweeps`/`runs`) is a coherent mechanism, and the follow-up commit addresses the earlier threads: `_exec_all_nodes_or_fail`, launcher `cd`/`PYTHONPATH` on `maxtext_root`, and dropping `NCCL_IB_HCA_LIST` from the shipped configs.

I am **not** re-raising the baked `XLA_FLAGS` quotes. `cvs/core/runtimes/docker.py` still emits `-e {k}={v}` unquoted into a remote shell, so wrapping the joined flags in quotes is what keeps `--flag=... --flag=...` as one docker token. Quoting at the docker layer would be the right long-term fix, but it is in the runtime, not this PR.

**Nits (non-blocking)**
- Three single-node configs set `container.env.NODE_RANK=0` even though the docs say per-node vars must not live there. The launcher overwrites it, so this is redundant rather than wrong; drop it so copy-paste into a distributed config does not freeze rank 0 on every node if someone later removes the launcher export.
- `jaxmaxtext_distributed.py` still says the coordinator is first node *when `jax_distributed.coordinator_ip` is `auto`*. That override path is gone; the launcher always uses `orch.hosts[0]`.
- DeepSeek-V4-284B has no `train_params.xla_flags` (every other shipped config folds a quoted `XLA_FLAGS`). Confirm that is intentional vs an omission.

**Tests / CI:** `test` and `container-smoke` are green. Unit tests cover the review-comment fixes; they still do not load the packaged JSON files (only a synthetic fixture). Not blocking.

cvs/lib/training/jaxmaxtext/jaxmaxtext_training_lib.py
solaiys: MaxText repo has a real branch named release/v26.7.

atnair-amd:
is there a validation run linked?

amd-droy (approved):
lgtm. thanks Saravanan.

cvs/lib/training/jaxmaxtext/utils/training_config_loader.py
solaiys: Good call — done. Dropped NCCL_IB_HCA_LIST from distributed detection and simplified the line to "distributed": bool(env.get("NCCL_IB_HCA")) (constant is now _NCCL_IB_DEVICE_KEY = "NCCL_IB_HCA"). No behavior change for the shipped configs (all distributed ones set NCCL_IB_HCA; single-node set neither), and it keeps detection consistent with NCCL_IB_HCA being the only HCA-selection var NCCL/RCCL honors.

solaiys:
> is there a validation run linked?

I have ran the training test. one report attached here.
[mi325x_llama-3.1-8b-dist.html](https://github.com/user-attachments/files/31971379/mi325x_llama-3.1-8b-dist.html)

PR #389, opened 2026-09-08
feat(vllm): report threshold gates as metric subtests

atnair-amd:
## Summary
- Collapse one-row-per-metric pytest tests into `test_verify_cell_metrics` so each cell reports active threshold gates as pass/fail/skip subtests.
- Keep HTML reports aligned with those verdicts, including skipped GPU/Prometheus metrics, without treating record-only values as passing tests.

## Test plan
- [x] make fmt-check
- [x] make lint
- [x] make ut


solaiys (approved):
LGTM

hnimra-amd:
LGTM

Instead of one pytest row per metric, there’s now one verification test per cell. the actual gates show up as subtests (pass/fail/skip). if thresholds aren’t being enforced, it just records the numbers in the html and skips instead of marking them as passed.

hnimra-amd (approved):

PR #391, opened 2026-09-08
Flatten Megatron configs, smoke, and sweep cell keys and update docs

sukesh-amd:
## Document unified suites and Primus checkpoint save

- Move suite and config detail from the long test/config READMEs into published RST (`docs/how-to/test-suites/training/megatron.rst` and `docs/reference/configuration-files/training/megatron.rst`).
- Pass Primus `--save` on checkpoint writes so resume can load a real checkpoint.

## Drive test_smoke from a top-level smoke block

- `test_smoke` reads `smoke.enabled`, `iters`, `micro_batch_size`, `global_batch_size`, and `precision` from the variant JSON instead of hardcoded knobs.
- `enabled=false` skips smoke. All packaged Megatron variant configs include this block.

## Flatten Megatron configs onto paths, train_params, and container.env

- Drop nested `config` / `model_params` / `schema_version` / `framework`.
- Use `gpu_name`, `paths`, `train_params`, and `container.env` (`NNODES`, `MASTER_ADDR`, `NCCL_*`, sockets as `docker run -e`).
- Flatten those sections into the job dict at load time. Wrapper scripts no longer re-export NCCL/socket/`MASTER_ADDR`/`NNODES`.
- Threshold specs with `"optional": true` SKIPPED in `test_metric` when the metric is missing (`training.mem_usage`; distributed also `training.scaling_efficiency_pct`).
- Checkpoint I/O parser matches Primus `loading distributed checkpoint from` as well as Megatron-LM `loading checkpoint from`.
- Keep how-to and schema RST in sync with the flattened layout.

## Require sweep keys to match MBS/GBS/PRECISION

- Rename every packaged `sweep.combinations` key and matching `sweep.runs` entry to `MBS=<micro_batch_size>,GBS=<global_batch_size>,PRECISION=<precision>` (same string as the threshold cell and the pytest parametrize ID).
- Config load fails if a combination key does not match those three fields, so pytest IDs and threshold lookup cannot drift.
- Log directories still sanitize `=` and `,` to `_` (for example `MBS_4_GBS_128_PRECISION_FP8`).
- Update how-to and schema RST examples to the new keys.

## Test plan

- [ ] `make fmt-check lint ut`
- [ ] Load a filled `*_single.json` and `*_distributed.json` via `load_training_variant` (all `<changeme>` replaced)
- [ ] Single-node: smoke from `smoke.*`; one sweep cell; optional mem_usage SKIPPED if the parser does not emit it
- [ ] Distributed Primus: checkpoint save/resume with `--save`; load log contains `loading distributed checkpoint from`
- [ ] Confirm combo keys are `MBS=…,GBS=…,PRECISION=…`; a leftover slug like `llama3_1_8b-mi300x-…` fails load


cvs/lib/training/megatron/primus_lib.py
```diff
@@ -397,16 +397,15 @@ def _build_primus_cmd(self):
             f'--train_iters {self.iterations}'
         )
         if self.checkpoint_dir and not self.load_checkpoint:
-            batch_args += f' --save_interval {self.save_interval or self.iterations}'
+            batch_args += f' --save --save_interval {self.save_interval or self.iterations}'
```
solaiys, line 400: Primus/Megatron `save` is a directory (default `null`), not a boolean. `--save --save_interval N` makes argparse/Hydra take `--save_interval` as the save path and drops the interval value. Resume already uses `--load {checkpoint_dir}/checkpoints`; the save phase must pass the same directory.

Use `--save {self.checkpoint_dir}/checkpoints --save_interval {self.save_interval or self.iterations}` (matching `test_checkpoint`'s `latest_checkpointed_iteration.txt` path). Add a unit test that the save-phase argv contains that path and does not pass a bare `--save`.

solaiys (changes requested):
Blocking: Primus `save` is a checkpoint directory, not a boolean flag. Bare `--save` will be parsed as the path `--save_interval` and the actual interval token is left dangling, so checkpoint write/resume still cannot work.

cvs/lib/training/megatron/primus_lib.py
sukesh-amd: during training --save path will be patched with primus_workspace path

cvs/input/config_file/training/megatron/mi300x_megatron_deepseek-v2-lite_distributed.json
```diff
@@ -59,8 +44,25 @@
...
         ]
       }
+    },
+    "env": {
+      "NNODES": "<changeme>",
+      "MASTER_ADDR": "<changeme>",
+      "NCCL_IB_HCA": "<changeme>",
+      "NCCL_SOCKET_IFNAME": "<changeme>",
+      "GLOO_SOCKET_IFNAME": "<changeme>",
+      "NCCL_IB_GID_INDEX": "3",
```
solaiys, line 54: this also should be \<changeme\> with example.
Either give the example in a comment line key:value.
OR
add the \<changeme\> tag at the end of the actual value what we are using

ex: 
      "NCCL_IB_HCA": "rdma0,rdma1,rdma2,rdma3,rdma4,rdma5,rdma6,rdma7 **\<changeme\>**",
      "NCCL_SOCKET_IFNAME": "eno0 **\<changeme\>**",
      "GLOO_SOCKET_IFNAME": "eno0 **\<changeme\>**",
      "NCCL_IB_GID_INDEX": "3 **\<changeme\>**",

cvs/input/config_file/training/megatron/mi300x_megatron_deepseek-v2-lite_distributed.json
```diff
@@ -59,8 +44,25 @@
           "/dev/infiniband/rdma_cm"
         ]
       }
+    },
+    "env": {
+      "NNODES": "<changeme>",
```
solaiys, line 49: We should remove NNODES from the config.
We decided to populate this value inside the code right?
based on the cluster host len, you can assign the value to this env internally.

cvs/input/config_file/training/megatron/mi300x_megatron_deepseek-v2-lite_distributed.json
sukesh-amd: ok

cvs/input/config_file/training/megatron/mi300x_megatron_deepseek-v2-lite_distributed.json
sukesh-amd: will update this in the next PR

solaiys (approved):
changeme tag should accompany with the example values.
submit it in the next PR.

PR #393, opened 2026-09-09
AIMVT-239: Add automated SSH key distribution test suite

ahskabir:
## Motivation

Multi-node RCCL and training test suites require password-less SSH between all cluster nodes as a prerequisite, but CVS has no automated way to set this up — users currently do it by hand per cluster. This adds a `cluster_key_distribution` test suite that distributes a shared cluster keypair, authorizes it (plus optionally the controlling station's own key) in `authorized_keys`, writes a managed `~/.ssh/config` Host block, and verifies password-less connectivity — all driven by a single JSON config.

## Technical Details

  - `cvs/lib/ssh_keys_lib.py`: config validation/normalization, SFTP key upload, authorized_keys injection (append-only, idempotent — never truncates the file), ~/.ssh/config managed-block rendering, and ring/full-mesh connectivity verification.
  - `cvs/tests/cluster_key_distribution/ssh_keys_distribution.py`: pytest suite, 6 ordered tests (dir setup → key distribution → cluster pubkey authorization → controlling-station pubkey authorization → ssh config install → passwordless SSH verification).
  - `cvs/input/config_file/cluster_key_distribution/ssh_key_distribute_config.json`: sample <changeme> config.
  - `ssh_dir_nfs_shared` config flag: when the remote ~/.ssh dir is the same NFS-mounted directory on every node, all mutating steps (key upload, authorized_keys writes, ~/.ssh/config write) route through the head node only via orch.exec_on_head/orch.head, instead of running redundantly — and racily — on every node against the same underlying file. Without this, concurrent sed -i (temp-file + rename) from multiple nodes against one NFS-shared file reproduced as base64: write error: Stale file handle on a live cluster.
  - Incidental fix: pytest.ini had ignore:.*Broken pipe.*:BrokenPipeError in filterwarnings, which is invalid under pytest 8+ (BrokenPipeError isn't a Warning subclass) and aborted every test run with UsageError — removed, since filterwarnings never matched raised
  exceptions anyway, only warnings.warn() calls.

## Test Plan

  - 58 mocked unit tests in cvs/lib/unittests/test_ssh_keys_lib.py (config validation, key upload incl. NFS-shared path, authorized_keys injection, ssh config rendering, connectivity verification) — no live cluster dependency.
  - make test (full suite: 1794 unit tests + CLI smoke tests).
  - Live run against a real 4-node cluster via the `cluster_key_distribution` suite, both with `ssh_dir_nfs_shared` true and false.

## Test Result

 - make test: all 1794 unit tests pass, all CLI smoke tests pass.
  - Live 4-node cluster run: 6/6 suite tests pass, including `install_ssh_config` on the node that previously hit the stale-NFS-handle race — confirmed fixed with no errors in the run log.

## Submission Checklist

### DIFF
--- a/cvs/lib/ssh_keys_lib.py
+++ b/cvs/lib/ssh_keys_lib.py
@@ -1,0 +1,438 @@
+'''
+Copyright 2025 Advanced Micro Devices, Inc.
+All rights reserved. This notice is intended as a precaution against inadvertent
+publication and does not imply publication or any waiver of confidentiality.
+The year included in the foregoing notice is the year of creation of the work.
+All code contained here is Property of Advanced Micro Devices, Inc.
+'''
+
+import base64
+import ipaddress
+import os
+import shlex
+
+from cvs.lib import globals
+
+log = globals.log
+
+DEFAULT_KEY_NAME = "cluster_id"
+DEFAULT_REMOTE_SSH_DIR = "~/.ssh"
+SSH_CONFIG_BEGIN = "# BEGIN CVS cluster_key_distribution (managed)"
+SSH_CONFIG_END = "# END CVS cluster_key_distribution (managed)"
+
+_KNOWN_DEFAULT_KEY_NAMES = {"id_rsa", "id_ed25519", "id_ecdsa", "id_dsa"}
+
+
+# ---------------------------------------------------------------------------
+# Pure / logic functions
+# ---------------------------------------------------------------------------
+
+
+def validate_key_distribution_config(config_dict):
+    """Validate config subsection and return a normalized dict with defaults applied."""
+    norm = dict(config_dict)
+
+    for field in ("cluster_key_private_path", "cluster_key_public_path"):
+        val = norm.get(field, "")
+        if not val:
+            raise ValueError(f"ssh_key_distribution.{field} is required and must be non-empty")
+        if not os.path.isfile(val):
+            raise ValueError(f"ssh_key_distribution.{field}: file not found: {val!r}")
+
+    controlling = norm.get("controlling_station_pubkey_path", "")
+    if controlling and not os.path.isfile(controlling):
+        raise ValueError(f"ssh_key_distribution.controlling_station_pubkey_path: file not found: {controlling!r}")
+
+    norm.setdefault("key_name", DEFAULT_KEY_NAME)
+    norm.setdefault("remote_ssh_dir", DEFAULT_REMOTE_SSH_DIR)
+    norm.setdefault("ssh_dir_nfs_shared", False)
+    norm.setdefault("ssh_config_host_pattern", "")
+    norm.setdefault("verify_connectivity", True)
+    norm.setdefault("verify_mode", "ring")
+    norm.setdefault("verify_timeout", 20)
+    norm.setdefault("ssh_config_write_mode", "managed_block")
+    norm.setdefault("controlling_station_pubkey_path", "")
+
+    key_name = norm["key_name"]
+    if key_name in _KNOWN_DEFAULT_KEY_NAMES:
+        log.warning(
+            "ssh_key_distribution.key_name=%r matches a well-known SSH default identity; "
+            "distribution may overwrite an existing key on remote nodes",
+            key_name,
+        )
+
+    return norm
+
+
+def collect_cluster_hostnames(cluster_dict):
+    """Return ordered, de-duplicated SSH-reachable identifiers: node keys + distinct vpc_ips."""
+    seen = set()
+    result = []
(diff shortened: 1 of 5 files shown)


cvs/lib/ssh_keys_lib.py
```diff
@@ -0,0 +1,461 @@
...
+            dst = nodes[(i + 1) % len(nodes)]
+            if src == dst:
+                continue
+            pairs.append((src, dst))
+            cmd_list.append(f"ssh -F {ssh_config_path} -o BatchMode=yes -o ConnectTimeout={timeout} {dst} true")
+        out = orch.all.exec_cmd_list(cmd_list, timeout=timeout + 10)
+        for (src, dst), output in zip(
+            pairs, [out.get(node, "") for node in nodes if node != nodes[0] or len(nodes) < 2]
+        ):
+            results[(src, dst)] = "error" not in str(output).lower() and output is not None
```
speriaswamy-amd, line 384: Blocking: we use "error" not in str(raw).lower() for judging failures, we should be using exit nodes. probe with detailed=True (or ssh ... true && echo CVS_SSH_OK) and require exit code 0

cvs/lib/ssh_keys_lib.py
```diff
@@ -0,0 +1,461 @@
...
+
+        # exec_cmd_list returns {node: output}; map back by position
+        node_outputs = [out.get(node, "") for node in nodes]
+        results = {}
+        for i, (src, dst) in enumerate(pairs):
+            raw = node_outputs[i] if i < len(node_outputs) else ""
+            results[(src, dst)] = raw is not None and "error" not in str(raw).lower()
+
+    else:
```
speriaswamy-amd, line 393: Blocking: This block generated n - 1 commands and executes on 1 node. But exec_cmd_list requires len(cmd_list) to match the host count, this would raise errors multinode job. Better option would be to use orch.exec(cmd, hosts=[src], detailed=True) as Pssh handle is deprecated

cvs/lib/ssh_keys_lib.py
```diff
@@ -0,0 +1,461 @@
...
+        # full_mesh: O(n*(n-1)) probes, one exec per source node
+        for src in nodes:
+            peers = [n for n in nodes if n != src]
+            if not peers:
+                continue
+            cmd_list = [
+                f"ssh -F {ssh_config_path} -o BatchMode=yes -o ConnectTimeout={timeout} {dst} true" for dst in peers
+            ]
+            # Run all probes from src via a temporary single-host handle
+            from cvs.lib.parallel_ssh_lib import Pssh
```
speriaswamy-amd, line 403: This is deprecated, should be using the new  `from cvs.lib.parallel.phandle import ParallelHandle`

cvs/lib/ssh_keys_lib.py
```diff
@@ -0,0 +1,461 @@
...
+    result = {}
+    for node, detail in out.items():
+        if isinstance(detail, dict):
+            result[node] = detail.get("exit_code", 0) == 0
+        else:
+            result[node] = True
+    return result
+
+
+def _upload_local_file(orch, local_path, remote_path):
```
speriaswamy-amd, line 454: This method doesn't seem to be used anywhere?

cvs/lib/ssh_keys_lib.py
```diff
@@ -0,0 +1,461 @@
...
+    key_name = norm_config["key_name"]
+    remote_ssh_dir = norm_config["remote_ssh_dir"]
+    nfs_shared = norm_config.get("ssh_dir_nfs_shared", False)
+    pubkey_remote = f"{remote_ssh_dir}/{key_name}.pub"
+
+    cmd = build_authorize_pubkey_cmd(remote_ssh_dir, pubkey_remote)
+    exec_target = orch.exec_on_head if nfs_shared else orch.exec
+    out = exec_target(cmd, timeout=30, detailed=True)
+
+    results = {node: True for node in orch.all.reachable_hosts}
```
speriaswamy-amd, line 289: We turn one result into per node status dict, i.e even if First node fails other nodes stay True, here's a walk through from claude on 3+ nodes: ```
alk through a 3-node cluster, head = n1, NFS write fails on head (exit_code=1):

Start: {n1: True, n2: True, n3: True} — every node assumed success.
out is only {n1: {exit_code: 1}}.
The loop updates only n1 to False.
Return: {n1: False, n2: True, n3: True}.
```
We should follow `upload_cluster_keys` pattern and mark every node as failed

speriaswamy-amd (changes requested):
Looks good overall, requires couple of changes though:

PR #396, opened 2026-09-09
jaxmaxtext: parseable sweep keys + GPU-generic mi3xx configs

solaiys:
# jaxmaxtext: parseable sweep keys + GPU-generic mi3xx configs

Two changes to the JAX MaxText suite: make sweep keys self-describing/parseable, and collapse the near-identical mi300x/mi325x configs into GPU-generic `mi3xx` configs.

## 1. Parseable sweep keys + per-sweep steps

- **Sweep keys are now a parseable spec**: `BS=<batch>,PRECISION=<BF16|FP8>,SL=<seqlen>` (dropped `NNODES`/`STEPS` from the key). The loader parses the key into `maxtext_config` overrides (`per_device_batch_size` / `quantization` / `max_target_length`). **FP8 is GPU-aware** — `nanoo_fp8` on MI300X/MI325X (CDNA3), `fp8` on MI350-class (CDNA4).
- Each sweep `{}` keeps a short `_comment` and may carry **extra overrides** (e.g. a per-sweep `steps`), which take precedence over the parsed key.
- Removed `dtype`/`weight_dtype` from sweeps (always the bfloat16 default) and the now-derived params.
- A per-sweep **`steps` override** now drives the run YAML, the poll/timeout budget, and completion detection (previously only the base `maxtext_config.steps` was honored).
- Threshold cell keys renamed to match; added a `_node_count_comment` (values are from 2N runs; 1N single-node; 4N for 405b/dsv4-284b).
- Updated the config reference + how-to docs; added loader/lib unit tests.

## 2. Merge mi300x + mi325x → mi3xx

- Combined the near-identical mi300x/mi325x configs into **8 GPU-generic `mi3xx_*` configs** (`gpu_name: mi3xx`, CDNA3). They ship the MI325X batch sizes; on MI300X a large batch may OOM → lower `per_device_batch_size` (the `BS` in the key).
- **Threshold files stay per-platform** (`mi300x_*` and `mi325x_*` kept as-is); each `mi3xx_*` config's `threshold_json` ships with a `<changeme>` tag so the user selects the GPU-specific threshold. The loader now **guards** this with a clear error (since `threshold_json` is read at config load).
- Repointed the config-copy CLI plugin tests to the `mi3xx` config.
- Renamed the how-to doc `training/jax.rst` → `training/jaxmaxtext.rst` and updated all references (toctree, index, reference `:doc:`).
- Docs: `mi3xx` runs on both MI300X/MI325X (with the MI300X OOM caveat), threshold selection, and refreshed available-configs table + examples.

## Validation

- `make fmt-check` clean, `make lint` 10.00/10, `make ut` green (1782 tests), docs build clean (no new warnings).
- Results updated in the below ticket.
   https://amd-hub.atlassian.net/browse/AIMVT-328


cvs/lib/training/jaxmaxtext/utils/training_config_loader.py
```diff
@@ -458,6 +503,20 @@ def load_training_variant(config_path, cluster_dict):
     internal shape (see `normalize_training_config`), attaches the thresholds, and
     builds the typed `TrainingVariantConfig`.
     """
+    # The GPU-generic mi3xx configs ship threshold_json with a '<changeme>' tag so
+    # the user selects the platform-specific threshold file (mi300x vs mi325x)
+    # before running. Guard here for a clear message -- otherwise substitute_config
+    # would fail with a generic file-not-found on the tagged name.
+    try:
+        _threshold_json = str(json.loads(Path(config_path).read_text()).get("threshold_json", ""))
```
amd-droy: This re-parses the JSON only to catch <changeme> in threshold_json; substitute_config already reads the file, so check the loaded value (or the resolved path) instead of a second json.loads.

cvs/lib/training/jaxmaxtext/utils/training_config_loader.py
```diff
@@ -336,6 +338,42 @@ def _build_xla_flags_env(xla_flags):
...
+    """MaxText FP8 quantization value for a GPU family.
+
+    CDNA4 (MI350/MI355 class) supports OCP FP8 -> ``fp8``; CDNA3 (MI300X/MI325X)
+    uses the NANOO/fnuz FP8 format -> ``nanoo_fp8``.
+    """
+    return "fp8" if str(gpu_name or "").lower().startswith("mi35") else "nanoo_fp8"
+
+
+def _parse_sweep_key(name, gpu_name=""):
+    """Parse a sweep key like ``BS=3,PRECISION=FP8,SL=8192`` into maxtext overrides.
```
amd-droy, line 349: _parse_sweep_key ignores unknown tokens and maps any non-FP8 precision to BF16, so a typo like BS→BATCH or PRECISION=FP16 still loads and runs the base config.

cvs/input/config_file/training/jaxmaxtext/mi35x_jaxmaxtext_llama-3.1-70b_single_threshold.json
```diff
@@ -1,6 +1,7 @@
 {
+  "_node_count_comment": "Threshold values here were captured on a 2-node (2N) run. Single-node configs use 1N; large models (llama-3.1-405b, deepseek-v4-284b) use 4N. Throughput/step-time scale with the GPU count, so if you run a different number of nodes, update these values to the appropriate targets for that node count.",
```
amd-droy: The same “captured on a 2-node (2N) run” _node_count_comment is pasted onto single-node and 4N threshold files; specialize it per file or keep it only in the RST.

atnair-amd (approved):
lgtm

cvs/lib/training/jaxmaxtext/utils/training_config_loader.py
solaiys: threshold_json guard: Moved the check into substitute_config (it already reads the config), so it validates the loaded threshold_json value — dropped the extra json.loads and the pre-read in load_training_variant. Now generic for all frameworks.

cvs/input/config_file/training/jaxmaxtext/mi35x_jaxmaxtext_llama-3.1-70b_single_threshold.json
solaiys: _node_count_comment: Specialized per file — 1N for single-node, 4N for 405b/deepseek-v4-284b, 2N for the rest.

cvs/lib/training/jaxmaxtext/utils/training_config_loader.py
solaiys: _parse_sweep_key: Now raises on malformed KEY=VALUE tokens — unknown keys (BATCH=…), bad PRECISION (FP16), and non-integer BS/SL — so typos fail at load instead of silently using the base config. Tokens without = stay opaque (arbitrary cell names still work). Added tests.

With this validation, the future data type ( precision) need to be updated as and when a new format need to be supported.
However, user can overwrite the actual params KEY:VALUE inside the sweep{} block as 
"quantization": **new value**

solaiys:
@amd-droy , Fixed all the 3 comments and tested.

training test results updated in https://amd-hub.atlassian.net/browse/AIMVT-328

> make ut confirmed green:
> 
> Ran 1798 tests in 12.757s
> OK (skipped=1)

PR #399, opened 2026-09-09
Flatten Torchtitan configs, smoke, and sweep cell keys and update docs

amd-rthummal:
## Document unified suites and Primus checkpoint save

- Move suite and config detail from the long test/config READMEs into published RST (`docs/how-to/test-suites/training/torchtitan.rst` and `docs/reference/configuration-files/training/torchtitan.rst`).
- Pass Primus `--save` on checkpoint writes so resume can load a real checkpoint.

## Drive test_smoke from a top-level smoke block

- `test_smoke` reads `smoke.enabled`, `iters`, `micro_batch_size`, `global_batch_size`, and `precision` from the variant JSON instead of hardcoded knobs.
- `enabled=false` skips smoke. All packaged torchtitan variant configs include this block.

## Flatten torchtitan configs onto paths, train_params, and container.env

- Drop nested `config` / `model_params` / `schema_version` / `framework`.
- Use `gpu_name`, `paths`, `train_params`, and `container.env` (`NNODES`, `MASTER_ADDR`, `NCCL_*`, sockets as `docker run -e`).
- Flatten those sections into the job dict at load time. Wrapper scripts no longer re-export NCCL/socket/`MASTER_ADDR`/`NNODES`.
- Threshold specs with `"optional": true` SKIPPED in `test_metric` when the metric is missing (`training.mem_usage`; distributed also `training.scaling_efficiency_pct`).
- Checkpoint I/O parser matches Primus `loading distributed checkpoint from` as well as Torchtitan `loading checkpoint from`.
- Keep how-to and schema RST in sync with the flattened layout.

## Require sweep keys to match MBS/GBS/PRECISION

- Rename every packaged `sweep.combinations` key and matching `sweep.runs` entry to `MBS=<micro_batch_size>,GBS=<global_batch_size>,PRECISION=<precision>` (same string as the threshold cell and the pytest parametrize ID).
- Config load fails if a combination key does not match those three fields, so pytest IDs and threshold lookup cannot drift.
- Log directories still sanitize `=` and `,` to `_` (for example `MBS_4_GBS_128_PRECISION_FP8`).
- Update how-to and schema RST examples to the new keys.

## Test plan

- [ ] `make fmt-check lint ut`
- [ ] Load a filled `*_single.json` and `*_distributed.json` via `load_training_variant` (all `<changeme>` replaced)
- [ ] Single-node: smoke from `smoke.*`; one sweep cell; optional mem_usage SKIPPED if the parser does not emit it
- [ ] Distributed Primus: checkpoint save/resume with `--save`; load log contains `loading distributed checkpoint from`
- [ ] Confirm combo keys are `MBS=…,GBS=…,PRECISION=…`; a leftover slug like `llama3_1_8b-mi300x-…` fails load


cvs/input/config_file/training/torchtitan/mi355x_torchtitan_llama-3.1-8b_distributed.json
```diff
@@ -0,0 +1,144 @@
...
+  "threshold_json": "mi355x_torchtitan_llama-3.1-8b_distributed_threshold.json",
+  "paths": {
+    "hf_token_file": "/home/{user-id}/.hf_token",
+    "log_dir": "/home/{user-id}/LOG_DIR",
+    "scripts_dir": "/home/{user-id}/SCRIPTS",
+    "data_cache_dir": "/shared/data/amd_aac/rajesh/models",
+    "rocm_dir": ""
+  },
+  "train_params": {
+    "multi_node": {
```
solaiys: **Blocking:** this is not flattened. Megatron (and the single-node TorchTitan files in this PR) use one `train_params` object with `model_name`, `tokenizer_model`, and parallelism at the top level. This still nests `multi_node.llama3_1_8b.mi300x` — and the inner GPU key is `mi300x` inside an `mi355x_*` file.

`TorchTitanTrainingJob` does `self.model_params = variant_config.train_params` then `train_params.get("model_name", "llama3_3_70b")`. There is no top-level `model_name` here, so every distributed stem silently trains Llama 3.3 70B. Even if this blob were copied up, inner `model_name` is `"llama3"`, which is not a `TORCHTITAN_MODELS` key (`llama3_1_8b` is).

Same shape on the other `mi355x_torchtitan_*_distributed.json` files. Flatten them to match the single-node configs. `test_download_tokenizer` also indexes `variant_config.model_params["tokenizer_model"]` and will KeyError on these files. Please add a unit test that loads a shipped distributed JSON and asserts `train_params["model_name"]`.

cvs/lib/training/torchtitan/utils/training_config_loader.py
```diff
@@ -1,210 +1,349 @@
...
+    data_cache_dir: str
+    rocm_dir: str = ""
+
+
+class TorchTitanContainerSpec(ContainerSpec):
+    env: Dict[str, str] = Field(default_factory=dict)
+
+
+# container.env uses real process env names; jobs still read the lowercase aliases.
+_CONTAINER_ENV_TO_JOB = {
```
solaiys: **Blocking:** this map copies `NNODES` / `MASTER_ADDR` / `NCCL_SOCKET_IFNAME` / `GLOO_SOCKET_IFNAME` / `NCCL_DEBUG` / `NCCL_IB_GID_INDEX` into the job dict, but not `NCCL_IB_HCA`. The job reads `self.config.get("nccl_ib_hca_list", "")` (always empty) and `build_training_job_cmd` still does `export NCCL_IB_HCA={self.nccl_ib_hca_list};`, which emits `export NCCL_IB_HCA=;` and **clears** the value docker already passed as `-e NCCL_IB_HCA`.

Add `"NCCL_IB_HCA": "nccl_ib_hca"` (Megatron) and export that value — or drop the wrapper `export NCCL_*` lines so `docker run -e` is the single source of truth.

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_llama-3.3-70b_single.json
```diff
@@ -0,0 +1,132 @@
...
+  "schema_version": 1,
+  "framework": "torchtitan_single",
+  "gpu_name": "MI355X",
+  "enforce_thresholds": true,
+  "threshold_json": "mi355x_torchtitan_llama-3.3-70b_single_threshold.json",
+  "paths": {
+    "hf_token_file": "/home/{user-id}/.hf_token",
+    "log_dir": "/home/{user-id}/LOGS/torchtitan",
+    "scripts_dir": "/home/{user-id}/SCRIPTS/torchtitan",
+    "data_cache_dir": "/shared/data/amd_aac/rajesh/models",
```
solaiys: **Blocking:** packaged templates must not hard-code another user's NFS path. Megatron uses `/home/{user-id}/cache`; JaxMaxText uses `{shared_fs}/cache/…`. Same `/shared/data/amd_aac/rajesh/models` path is in llama-3.3-70b / mixtral / qwen3 singles and most distributed stems.

Also vs the flattened Megatron layout: do not put `NNODES` in `container.env`. Inject `str(len(node_dict))` at docker launch (Megatron `conftest.py`). Distributed files still have `"NNODES": "<changeme>"`, which will `int()`-crash in `TorchTitanTrainingJob` if the cluster merge does not replace it.

docs/reference/configuration-files/training/torchtitan.rst
```diff
@@ -0,0 +1,733 @@
...
+.. list-table::
+   :widths: 4 2 2 2 2
+   :header-rows: 1
+
+   * - Model
+     - MI300X
+     - MI325X
+     - MI355X
+     - Mode
+   * - Llama 3.1 8B
```
solaiys: **Blocking:** this table marks Llama 3.1 8B (and 3.3 70B / 405B / Qwen3 / Mixtral below) as available on MI300X and MI325X, but this PR deletes every `mi3xx_torchtitan_*` file and ships only `mi355x_*`.

Megatron keeps shared `mi3xx_*` JSON + SKU `threshold_json` (`mi300x_*` / `mi325x_*`). JaxMaxText ships per-SKU files. Restore flattened `mi3xx_torchtitan_*` (or per-SKU `mi300x_` / `mi325x_`) configs, or change this table to MI355X-only so it matches what is actually packaged.

solaiys (changes requested):
## Review: blocking issues

Compared the shipped TorchTitan JSON to flattened Megatron (`paths` + flat `train_params` + `container.env`, SKU `gpu_name` / `threshold_json`, no `NNODES` in JSON) and to JaxMaxText (per-SKU files, `{user-id}` / `{shared_fs}` paths). Sweep keys `MBS=…,GBS=…,PRECISION=…` and the **single-node** `train_params` shape match that layout. Distributed configs and env wiring do not, and will mis-run.

### 🔴 Distributed `train_params` is still nested `multi_node.<model>.mi300x`
The job reads flat keys. Nested JSON has no top-level `model_name`, so every distributed stem defaults to Llama 3.3 70B. Flatten like Megatron / the single-node files.

### 🔴 Wrapper exports empty `NCCL_IB_HCA` and wipes `docker -e`
`_CONTAINER_ENV_TO_JOB` omits `NCCL_IB_HCA`. The job still `export`s `nccl_ib_hca_list` (default `""`), clearing the docker env.

### 🔴 Config layout does not match Megatron / JaxMaxText
Hard-coded `/shared/data/amd_aac/rajesh/models`; `NNODES` still in JSON (Megatron injects from `len(node_dict)`); `mi3xx_torchtitan_*` deleted while the docs table still marks MI300X/MI325X as available.

Inline comments below. Please also get @cijohnson's approval before merge.

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_llama-3.1-8b_distributed.json
amd-rthummal: resolved

cvs/lib/training/torchtitan/utils/training_config_loader.py
amd-rthummal: resolved

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_llama-3.3-70b_single.json
amd-rthummal: resolved

docs/reference/configuration-files/training/torchtitan.rst
amd-rthummal: resolved

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_deepseek-v2-lite_distributed.json
```diff
@@ -0,0 +1,128 @@
+{
+  "schema_version": 1,
```
solaiys: We decided to drop this as we are not going to map and track this version with the releases.

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_qwen3-32b_distributed.json
```diff
@@ -0,0 +1,128 @@
...
+  "enforce_thresholds": true,
+  "threshold_json": "mi355x_torchtitan_qwen3-32b_distributed_threshold.json",
+  "paths": {
+    "hf_token_file": "/home/{user-id}/.hf_token",
+    "log_dir": "/home/{user-id}/LOG_DIR",
+    "scripts_dir": "/home/{user-id}/SCRIPTS",
+    "data_cache_dir": "/home/{user-id}/cache",
+    "rocm_dir": ""
+  },
+  "train_params": {
```
solaiys: place the train_params{} block after the container{} block.
idea is keep the block at the top which needs user edits (changeme)

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_qwen3-32b_distributed.json
```diff
@@ -0,0 +1,128 @@
...
+    "data_parallel_shard_degree": "16",
+    "global_batch_size": "512",
+    "micro_batch_size": "1",
+    "precision": "BF16",
+    "tokenizer_model": "Qwen/Qwen2.5-32B",
+    "training_iterations": "<changeme>",
+    "torchtitan_root": "/workspace/Primus/third_party/torchtitan",
+    "nic_type": "broadcom"
+  },
+  "container": {
```
solaiys, line 13: move it to before train_params{} after paths{}

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_qwen3-32b_distributed.json
```diff
@@ -0,0 +1,128 @@
...
+        ]
+      }
+    },
+    "env": {
+      "MASTER_ADDR": "<changeme>",
+      "NCCL_SOCKET_IFNAME": "<changeme>",
+      "GLOO_SOCKET_IFNAME": "<changeme>",
+      "NCCL_IB_GID_INDEX": "3",
+      "NCCL_DEBUG": "ERROR",
+      "NCCL_IB_HCA": "<changeme>"
```
solaiys: Keep the example values + Have the changeme tag at the end of the values.
OR
Add some comment line to show the default values which are in use in your testing.

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_qwen3-32b_distributed.json
```diff
@@ -0,0 +1,128 @@
...
+    "tokens_per_sec_total": 0.0,
+    "num_nodes": 1
+  },
+  "_smoke_comment": "test_smoke loads the model and runs a small fixed cell without metric gating. enabled=false skips it.",
+  "_loss_curve_comment": "sample training loss every sample_every steps (plus milestone_steps) and pass when the least-squares slope < max_slope. enforce=false makes it record-only.",
+  "_convergence_comment": "target_metric auto uses validation loss when available, else training loss. target_value <= 0 disables convergence checking (record-only).",
+  "_scaling_baseline_comment": "1-node total tokens/sec baseline for scaling-efficiency %. 0.0 = disabled (record-only). Populate from a prior MI355X single-node run (tok/s/GPU * 8) to enable the metric.",
+  "_checkpoint_comment": "checkpoint_dir is the shared filesystem path where TorchTitan writes and reads checkpoints across nodes. save_interval controls checkpoint frequency (steps); save_iters is the step at which the save-phase run stops; resume_iters is the step at which the resume-phase run stops. Loss at resume_iters must match the save-phase loss within loss_rtol. enforce=false makes checkpoint testing record-only.",
```
solaiys: Place all these comments just before its respective blocks{}.

cvs/lib/training/torchtitan/torchtitan_lib.py
```diff
@@ -167,7 +167,6 @@ def __init__(
         self.nnodes = int(self.config.get('nnodes', 1))
```
solaiys: **Blocking:** after dropping `NNODES` from distributed JSON, this still defaults to `1`. Megatron and JaxMaxText use `len(orch.hosts)`.

`job_config_dict()` only sees JSON `container.env`, which no longer has `NNODES`, so every distributed stem does `torchrun --nnodes 1` and `start_training_job` only launches `orch.hosts[0]`.

Set `self.nnodes = len(self.orch.hosts)` (same as Megatron). Docker `-e NNODES` does not update this field.

cvs/tests/training/torchtitan/conftest.py
```diff
@@ -106,6 +106,13 @@ def orch(cluster_dict, variant_config, lifecycle):
     double-tearing down in the normal case.
     """
     container_block = _deep_merge(cluster_dict.get("container", {}), variant_config.container.model_dump())
+
+    # Inject NNODES based on cluster node count (str(len(node_dict)) at docker launch)
+    # Distributed configs should not hard-code NNODES; calculate from actual cluster size
+    if "env" not in container_block:
+        container_block["env"] = {}
+    node_list = cluster_dict.get("nodes", [])
```
solaiys: **Blocking:** cluster files use `node_dict`, not `nodes`. `cluster_dict.get("nodes", [])` is always `[]`, so this always injects `NNODES=1`.

Even with the right key, this only sets docker `-e`; `TorchTitanTrainingJob` still reads `self.config.get("nnodes", 1)` from the JSON-loaded variant and never sees this value.

Use `len(cluster_dict.get("node_dict") or {})` here if you keep docker injection, and set `self.nnodes = len(orch.hosts)` in the job.

solaiys (changes requested):
## Re-review (HEAD a8a9ce4)

Previous blockers are fixed: distributed `train_params` is flattened, `NCCL_IB_HCA` is mapped, `data_cache_dir` uses `{user-id}`, and how-to docs are MI355X-only.

One new blocker from the NNODES follow-up: distributed jobs still always run as **1 node**. Inline comments below.

Please also get @cijohnson's approval before merge.

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_deepseek-v2-lite_distributed.json
amd-rthummal: resolved

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_qwen3-32b_distributed.json
amd-rthummal: resolved

amd-rthummal: fixed

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_deepseek-v2-lite_distributed.json
```diff
@@ -0,0 +1,128 @@
...
+  "framework": "torchtitan_distributed",
```
solaiys: We decided to drop this as well.

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_qwen3-32b_distributed.json
amd-rthummal: fixed

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_qwen3-32b_distributed.json
amd-rthummal: fixed

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_qwen3-32b_distributed.json
amd-rthummal: fixed

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_deepseek-v2-lite_distributed.json
```diff
@@ -0,0 +1,128 @@
...
+  "_checkpoint_comment": "checkpoint_dir is the shared filesystem path where TorchTitan writes and reads checkpoints across nodes. save_interval controls checkpoint frequency (steps); save_iters is the step at which the save-phase run stops; resume_iters is the step at which the resume-phase run stops. Loss at resume_iters must match the save-phase loss within loss_rtol. enforce=false makes checkpoint testing record-only.",
+  "sweep": {
+    "combinations": {
+      "MBS=2,GBS=512,PRECISION=BF16": {
+        "name": "deepseek_v2_lite_mbs2_gbs512_BF16",
+        "global_batch_size": "512",
+        "micro_batch_size": "2",
+        "precision": "BF16"
+      }
+    },
```
solaiys, line 118: ompare the JAX or Megatron configs.
 
Are you parsing the values from the  KEY name itself. ( This was the ask from Ignatious. )
"MBS=2,GBS=512,PRECISION=BF16"
So body of this block should contains the additional params if anything needs override.
logically we can have the same variable inside as well, which should get overwritten.

cvs/lib/training/torchtitan/torchtitan_lib.py
amd-rthummal: fixed

cvs/tests/training/torchtitan/conftest.py
amd-rthummal: fixed

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_deepseek-v2-lite_distributed.json
amd-rthummal: fixed

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_deepseek-v2-lite_distributed.json
amd-rthummal: fixed

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_llama-3.1-70b_distributed.json
```diff
@@ -0,0 +1,123 @@
...
+        "network": "host",
+        "ipc": "host",
+        "privileged": true,
+        "shm_size": "128G",
+        "volumes": [
+          "/home/{user-id}:/home/{user-id}",
+          "/dev/infiniband:/dev/infiniband",
+          "/usr/local/lib/libbnxt_re-rdmav34.so:/usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so.host",
+          "/lib/libibverbs.d:/lib/libibverbs.d",
+          "<changeme>:<changeme>",
```
solaiys: What is this for?
Do we need this line?  
  "\<changeme\>:\<changeme\>",

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_llama-3.1-8b_single.json
```diff
@@ -70,47 +22,98 @@
...
+  },
   "sweep": {
     "combinations": {
-      "llama3_1_8b-mi355-bs48-mbs6-bf16": {
-        "name": "llama3_1_8b_mbs6_gbs48_BF16",
-        "global_batch_size": "48",
-        "micro_batch_size": "6",
-        "precision": "BF16"
+      "MBS=6,GBS=48,PRECISION=BF16": {
+        "name": "llama3_1_8b_mbs6_gbs48_BF16"
```
solaiys, line 100: **Blocking:** combination bodies now only have `name`, matching Megatron PR 405. TorchTitan does not parse MBS/GBS/PRECISION from the key.

`TorchTitanSweepCombo` still requires `micro_batch_size` and `global_batch_size`, and `validate_combo_keys_match_params` builds the key from those fields. Load will fail on every packaged file.

`pytest_generate_tests` in `torchtitan_single.py` / `torchtitan_distributed.py` still does `combo["micro_batch_size"]` on the raw JSON and will KeyError at collection.

Port Megatron's `parse_sweep_cell_key` / `_assign_params_from_keys` (and parametrize from the key, or from the filled combo). Keep extra body fields as overlays if you need them.

cvs/lib/training/torchtitan/utils/training_config_loader.py
```diff
@@ -46,6 +46,33 @@ class TorchTitanSweepCombo(_Forbid):
     precision: str = ""
 
 
+def sweep_cell_key(combo) -> str:
```
solaiys, line 74: **Blocking:** this still requires combo-body `micro_batch_size` / `global_batch_size` and rebuilds the cell key from those fields. There is no `parse_sweep_cell_key` / `_assign_params_from_keys` like Megatron.

After this PR's JSON change, bodies are `{ "name": "..." }` only, so `TorchTitanVariantConfig` cannot load any packaged template.

Fill MBS/GBS/PRECISION from the combination key before constructing `TorchTitanSweepCombo`, and allow extra overlay fields (`_Allow` + `model_extra`) if the body should override `train_params`.

cvs/lib/training/torchtitan/utils/training_config_loader.py
```diff
@@ -144,30 +186,93 @@ class CheckpointConfig(_Forbid):
...
+    "NCCL_IB_HCA": "nccl_ib_hca",
+    "NCCL_SOCKET_IFNAME": "nccl_socket_ifname",
+    "GLOO_SOCKET_IFNAME": "gloo_socket_ifname",
+    "NCCL_DEBUG": "nccl_debug",
+    "NCCL_IB_GID_INDEX": "nccl_ib_gid_index",
+}
+
+
 class TorchTitanVariantConfig(_Forbid):
     schema_version: Literal[1]
```
solaiys: **Blocking:** packaged JSON no longer has `schema_version` or `framework` (Megatron layout), but this model still requires both. `load_training_variant` will ValidationError on every shipped file.

Drop these fields from `TorchTitanVariantConfig` (Megatron) or put them back on every JSON. Suite choice is which pytest module you run, not a JSON discriminator.

solaiys (changes requested):
## Re-review (HEAD 5050477)

The previous NNODES blocker is fixed: `self.nnodes = len(self.orch.hosts)` and conftest injects from `node_dict`.

**Not approving.** Packaged JSON was switched to Megatron-style sweep keys (body is only `name`) and dropped `schema_version` / `framework`, but the TorchTitan loader and pytest parametrize still require the old fields. Packaged configs will fail to load and collection will KeyError. Inline comments below.

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_llama-3.1-70b_distributed.json
```diff
@@ -0,0 +1,123 @@
...
+      "name": "docker",
+      "args": {
+        "network": "host",
+        "ipc": "host",
+        "privileged": true,
+        "shm_size": "128G",
+        "volumes": [
+          "/home/{user-id}:/home/{user-id}",
+          "/dev/infiniband:/dev/infiniband",
+          "/usr/local/lib/libbnxt_re-rdmav34.so:/usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav34.so.host",
```
solaiys: Are you still using this mount ?
copying this so.host to .so file inside the container?

Refer JAXmaxtext config file.
mount it with ":ro" as read only mount directly to the exact file.

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_llama-3.1-70b_distributed.json
amd-rthummal: removed

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_deepseek-v2-lite_distributed.json
```diff
@@ -0,0 +1,123 @@
...
+  "enforce_thresholds": true,
+  "threshold_json": "mi355x_torchtitan_deepseek-v2-lite_distributed_threshold.json",
+  "paths": {
+    "hf_token_file": "/home/{user-id}/.hf_token",
+    "log_dir": "/home/{user-id}/LOG_DIR",
+    "scripts_dir": "/home/{user-id}/SCRIPTS",
+    "data_cache_dir": "/home/{user-id}/cache",
+    "rocm_dir": ""
+  },
+  "verify_network_errors": "True",
```
solaiys, line 12: This is something you (torchtitan) only have it.
is it needed for the user to choose false ?

solaiys (approved):
@amd-rthummal ,
I am approving it to unblock the merge for you.
After fixing the review comments, try run the full ut and torch titan training once.
Then confirm with Ignatious and then merge it.
thanks.

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_llama-3.1-8b_single.json
amd-rthummal: resolved

cvs/lib/training/torchtitan/utils/training_config_loader.py
amd-rthummal: implemented

cvs/lib/training/torchtitan/utils/training_config_loader.py
amd-rthummal: resolved

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_llama-3.1-70b_distributed.json
amd-rthummal: resolved

cvs/input/config_file/training/torchtitan/mi355x_torchtitan_deepseek-v2-lite_distributed.json
amd-rthummal: fixed

cijohnson (approved):

PR #401, opened 2026-09-10
feat(vllm): flatten metric threshold contract

atnair-amd:
## Summary

Replace vLLM's namespaced metric vocabulary with one bare, versioned 56-metric reporting contract while preserving per-cell threshold subtests.

Ticket: https://amd-hub.atlassian.net/browse/AIMVT-329

## Change

- flatten `client.`, `gpu.`, and `prom.` metric keys across results, thresholds, subtests, HTML, JUnit, and Run Deck
- report all 56 registered metrics while retaining the packaged catalog's established 32-metric assertion subset
- validate partial threshold cells as strict finite numeric `min`/`max` specs
- preserve merged metric-subtest reporting and leave ATOM behavior unchanged
- reject incompatible historical prefixed Run Deck baselines explicitly

## Live evidence

One-node TensorWave validation pending.

## Tests

- `make fmt-check`
- `make lint`
- `make test` — 1,834 unit tests passed with 1 expected skip; 65 CLI checks passed

### DIFF
--- a/cvs/lib/inference/utils/vllm_metrics.py
+++ b/cvs/lib/inference/utils/vllm_metrics.py
@@ -36,6 +36,10 @@ _METADATA_FIELDS = frozenset(
 )
 
 
+class UnknownMetricContractError(ValueError):
+    """A finite artifact field falls outside the canonical vLLM registry."""
+
+
 def _raw(name, unit, category, direction, raw_source=None):
     source = raw_source or name
     return MetricDefinition(name, source, 'identity', (source,), unit, 'client', category, direction)
@@ -337,7 +341,7 @@ def project_vllm_metrics(raw, *, tp, isl, pp='1', artifact_path='results'):
             continue
         if raw_name in _METADATA_FIELDS or not is_finite_number(value):
             continue
-        raise ValueError(f'unknown finite numeric vLLM result field {raw_name!r}: {artifact_path}')
+        raise UnknownMetricContractError(f'unknown finite numeric vLLM result field {raw_name!r}: {artifact_path}')
 
     context = {'tp': tp, 'pp': pp, 'isl': isl}
     for definition in _DERIVED_CLIENT_METRICS:
--- a/cvs/lib/inference/utils/vllm_metrics.py
+++ b/cvs/lib/inference/utils/vllm_metrics.py
@@ -1,0 +1,393 @@
+'''Canonical metric contract for CVS vLLM benchmark results.'''
+
+import math
+from collections import namedtuple
+from types import MappingProxyType
+
+from cvs.lib.utils.gpu import GPU_METRICS as _SHARED_GPU_METRICS
+
+
+MetricDefinition = namedtuple(
+    'MetricDefinition',
+    (
+        'name',
+        'raw_source',
+        'derivation',
+        'required_inputs',
+        'unit',
+        'datasource',
+        'category',
+        'direction',
+    ),
+)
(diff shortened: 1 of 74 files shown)


cvs/lib/inference/utils/vllm_metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    metrics = {}
+    for raw_name, value in raw.items():
+        metric_name = _RAW_CLIENT_METRICS.get(raw_name)
+        if metric_name is not None:
+            if is_finite_number(value):
+                metrics[metric_name] = value
+            continue
+        if raw_name in _METADATA_FIELDS or not is_finite_number(value):
+            continue
+        raise UnknownMetricContractError(f'unknown finite numeric vLLM result field {raw_name!r}: {artifact_path}')
```
amd-droy, line 344: Closed-world projection will fail the whole cell on any new finite numeric field in the vLLM results artifact, while a registered field that is NaN/inf/non-numeric is dropped with no error. That is a brittle upgrade: a vLLM patch that adds one extra number (or emits duration as a string) either aborts inference or silently omits a gated metric. The old to_client_metrics path also derived failed from num_prompts - completed when failed was omitted; that fallback is gone, so failed and success_rate can be missing on otherwise successful runs.

cvs/tests/inference/vllm/_common.py
```diff
@@ -228,26 +296,30 @@ def test_vllm_inference(orch, variant_config, hf_token, vllm_targets, run, inf_r
...
+            "gpu_bandwidth_util_pct": _finite_or_none(aggregate.get("gpu_bandwidth_util_pct")),
+            "gpu_compute_util_pct": _finite_or_none(aggregate.get("gpu_compute_util_pct")),
+        }
+        prom_results = to_prom_metrics(before_prom, after_prom)
+        published_results = {
+            host: merge_metric_sources(actuals, gpu_results, prom_results) for host, actuals in results.items()
+        }
+        inf_res_dict[_cell_result_key(variant_config, run)] = published_results
+    except UnknownMetricContractError:
+        raise
```
amd-droy: UnknownMetricContractError is re-raised without setting lifecycle.failed. Every other exception in test_vllm_inference marks the lifecycle failed so later cells skip. A contract parse error fails only that parametrized cell; remaining cells still launch servers and run the client. If the extra field is in every artifact, that wastes the rest of the sweep; if it is intermittent, later cells keep going with no suite-level abort. Set lifecycle.failed = True before re-raising, same as the generic except Exception path.

cvs/lib/inference/utils/vllm_metrics.py
atnair-amd: No code change: unknown finite numeric fields are intentionally rejected by the documented 56-metric contract. Registered invalid values cannot silently pass an active threshold, and ATOM retains the failed-count fallback in its existing parser.

cvs/tests/inference/vllm/_common.py
atnair-amd: Unknown contract fields now fail the cell and mark lifecycle.failed so the sweep stops; later cells no longer reuse a tainted server. Coverage is in test_load_metric_reuse.py.

hnimra-amd (approved):
LGTM! The two inline threads are already handled: closed-world projection is an intentional contract, and lifecycle.failed on unknown fields is in the latest commit.

PR #405, opened 2026-09-10
feat(megatron): parseable sweep keys, combo overlays, and MI355X Llama 8B/70B single templates

sukesh-amd:
## Summary

Unified Megatron suites (`megatron_single` / `megatron_distributed`) now identify sweep cells with a parseable key, treat combination bodies as optional `train_params` overlays, use a consistent packaged JSON layout, and ship MI355X Llama 8B/70B single-node templates only.

### Commits

- `25f1367e` — parseable sweep keys (`MBS`/`GBS`/`PRECISION`) and per-combo overlays
- `d53c3aec` — format `training_config_loader.py`
- `8dd0e56d` — keep MI355X Llama 8B/70B single templates; order blocks `paths` → `container` → `train_params`

### Sweep keys

Combination keys (and matching `sweep.runs` entries and threshold cells) are:

```text
MBS=<micro_batch_size>,GBS=<global_batch_size>,PRECISION=<precision>
```

The loader parses MBS, GBS, and precision from the key. Repeating those fields in the combination body is optional. If the body sets them, they must match the key or load fails.

`pytest_generate_tests` in `cvs/tests/training/megatron/conftest.py` parametrizes `sweep_name` from `sweep.runs` (must exist in `sweep.combinations`). There is no cartesian product. The pytest ID is the combination key so it matches the threshold cell.

`test_training`, `test_metric`, and `test_loss_curve` in `megatron_single.py` and `megatron_distributed.py` take `sweep_name` instead of looping combinations inside one test.

### Combination body overlay

`MegatronSweepCombo` allows extra fields. Packaged cells only set `"training_iterations": "20"`. Those extras are `combo.model_extra` and merge into `train_params` as `sweep_overrides` in `megatron_lib.py` / `primus_lib.py`. Any other `train_params` key can be overlaid the same way; MBS/GBS/precision still come from the key.

### Implicit default cell

If `sweep` is omitted or `combinations` is empty, the loader inserts one cell named `default` from `train_params` (MBS/GBS/precision fall back to `2` / `128` / `BF16` if those keys are missing). The threshold file needs a `"default"` cell when `enforce_thresholds` is true; that cell is optional when it is false.

`NNODES` is not a JSON field (injected from host count).

### Packaged JSON and docs

Remaining templates use `paths`, then `container`, then `train_params`. On distributed files, `verify_network_errors` sits after `train_params`. `_sweep_comment` documents the key format and overlays.

On `*_distributed.json`, `paths.data_cache_dir` must be a shared filesystem. The packaged `/home/{user-id}/cache` value is node-local once home is bind-mounted.

Llama 3.1 8B, Llama 3.3 70B, and DeepSeek V2 Lite support Megatron-LM and Primus (backend is `container.image`). Llama 3.1 405B remains distributed-only and Primus-only (`mi3xx_megatron_llama-3.1-405b_distributed.json`).

Leftover nested `mi3xx_megatron_llama_{single,distributed}.json` / `mi35x_megatron_llama_single.json` are unchanged and are for `megatron_llama3_1_*` only.

### MI355X catalog

Keep:

- `mi355x_megatron_llama-3.1-8b_single.json` and threshold
- `mi355x_megatron_llama-3.3-70b_single.json` and threshold

Remove (config and matching threshold):

- MI355X DeepSeek V2 Lite single and distributed
- MI355X Llama 3.1 405B distributed
- MI355X Llama 3.1 8B distributed
- MI355X Llama 3.3 70B distributed

How-to, reference schema, pick-config-file, and both Megatron READMEs match that catalog.

## Test plan

- [ ] Load a packaged `mi3xx_*_single.json` and `*_distributed.json` after replacing `<changeme>`; pytest IDs should be `MBS=...,GBS=...,PRECISION=...` and match threshold cells
- [ ] Confirm combo overlay: listed cells run with `training_iterations` 20; other `train_params` unchanged unless set in the combination body
- [ ] Confirm a body that sets `micro_batch_size` / `global_batch_size` / `precision` different from the key fails at load
- [ ] Confirm a config with no `sweep` runs one implicit `default` cell; with `enforce_thresholds: true` the threshold file needs `"default"`
- [ ] Confirm `cvs config list training/megatron` for MI355X only lists Llama 8B/70B single files
- [ ] Confirm distributed `data_cache_dir` is documented as a shared filesystem; do not use node-local home cache across nodes
- [ ] Run `make fmt-check` and `make lint`


solaiys (approved):
LGTM.

Reviewed sweep-key parsing (`MBS=…,GBS=…,PRECISION=…`), combo `model_extra` overlays into `train_params`, implicit `default` cell, pytest parametrization from `sweep.runs`, `len(orch.hosts)` / `node_dict` NNODES injection, and the MI355X Llama 8B/70B single-only catalog. Packaged JSON, threshold cells, and docs match. No blocking issues.

PR #407, opened 2026-09-10
feat(vllm): allow vllm_distributed on more than two hosts

atnair-amd:
Removes the hard two-host cap on `vllm_distributed`.

`resolve_vllm_topology()` previously rejected any cluster with more than two hosts. The job path already derived `--nnodes` and per-rank `--node-rank` from the orchestrator host list, so the cap was the only blocker for 3+ node runs.

**Behavior now**
- One host: unchanged single-node fallback (`PP=1`)
- Two or more hosts: one distributed service across every cluster host
- Unchanged: mp still requires `pipeline_parallel_size > 1`; multi-host still requires `container.env.NCCL_SOCKET_IFNAME` (or the legacy `ib_netdev` fallback)

Packaged MI3xx/MI325X recipes remain two-host with `pipeline_parallel_size: 2`. Operators running a larger cluster need matching PP and calibrated thresholds; this PR does not add 4-node configs.

**Docs** updated in the vLLM how-to, config reference, and `cvs/input/config_file/inference/vllm/README.md`.

**Checks**: `make fmt-check`, `make lint`, and `make ut` (1825 tests) pass.

### DIFF
--- a/cvs/input/config_file/inference/vllm/README.md
+++ b/cvs/input/config_file/inference/vllm/README.md
@@ -24,7 +24,7 @@ is absent, but keeping one pair per directory avoids the trap entirely.
 | Mode | PP | Host behavior |
 |---|---|---|
 | `single` | 1 | first cluster host only |
-| `distributed` | 2 | exactly two cluster hosts form one service |
+| `distributed` | from config | all cluster hosts form one service |
 
 For mapping-style `node_dict`, "first cluster host" means the first JSON key in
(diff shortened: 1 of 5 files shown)


hnimra-amd (approved):
LGTM. The two host abort was the only thing blocking N-node vllm_distributed; --nnodes / --node-rank already come from the orchestrator host list. One-host fallback, mp PP>1, and NCCL_SOCKET_IFNAME are unchanged. Docs correctly say packaged recipes/thresholds are still two-host.

PR #439, opened 2026-09-15
removed the values of few variables that were accidently checked in t…

urtiwari:
removed the values of few variables that were accidently checked in the code, added <changeme>


cvs/input/config_file/preflight/preflight_config.json
```diff
@@ -21,15 +21,15 @@
       "rdma": {
         "_comment": "RDMA connectivity testing parameters using ibv_rc_pingpong (direct IB verbs) for RCCL-compatible testing",
 
-        "connectivity_mode": "skip",
+        "connectivity_mode": "run",
```
speriaswamy-amd: I don't think `run` is a connectivity mode, only options are `basic, full_mesh and skip` also running rdma connectivity tests by default would add some delay

speriaswamy-amd (changes requested):

speriaswamy-amd (approved):

PR #455, opened 2026-09-18
Remove ATOM from the 0.2.0 inference changelog entry.

hnimra-amd:
CVS 0.2.0 does not claim ATOM support, so drop the ATOM clause from the unreleased inference entry. The vLLM, SGLang, and pytorch_xdit entries are unchanged.


amd-droy (approved):
lgtm, thanks

